PR #4, opened 2024-11-19
remove kokkos MDspan and fix vectorAdd bug

psychocoderHPC:
- remove kokkos MDspan with a self-implemented memory view object
- fix missing wait in vectorAdd


SimeonEhrig:
Why did you remove mdspan and implement something by yourself?

PR #32, opened 2024-12-19
allow using scalar values instead of `Vec{value`}

psychocoderHPC:
This change simplifies using interfaces where a 1-dimensional vector is required.
The user can now pass a scalar value instead of a 1-dimension vector.

Maybe I missed updating some interfaces, if so please notify me.


SimeonEhrig:
Looks good. I think for the common use cases it fully fine to set 1D with a scalar and multi dimensional with `alpaka::Vec`. I will check if you missed something after Christmas.

PR #47, opened 2025-02-19
create for each test source file an own executable

SimeonEhrig:
The CMake iterate over all sub directories for a given folder and detects all `.cpp` files in it. For each source file, it creates a test executable. If an include folder exist in the sub folder, it is set as include path for each test executeable.

For example:
```
unit
|->TestA
     |-> testA.cpp
|->TestB
     |-> include
           |-> testB.hpp
     |-> testB.cpp
|->TestC
     |-> include
           |-> testC.hpp
     |-> testC1.cpp
     |-> testC2.cpp
     |-> testC3.cpp
```

- `TestA`: One executable uses only the source file `testA.cpp`.
- `TestB`: One executable uses the `include` folder and the source file `testB.cpp`.
- `TestC`: Three executable created for the source files `testC1.cpp`, `testC2.cpp` and `testC3.cpp`. All uses the `include` folder as include path.


psychocoderHPC:
Thanks for the PR, if you add that an **optional** include folder will be usable it will be cool. A binary per cpp file is fine.
At the moment I do not see a need for having to cpp, Maybe for testing global memory device variables but for this test, we could simply exclude the folder from the cpp scan and have a hard-wired cpp CMake file instead of a complicated generic approach.

SimeonEhrig:
> Thanks for the PR, if you add that an **optional** include folder will be usable it will be cool. A binary per cpp file is fine. At the moment I do not see a need for having to cpp, Maybe for testing global memory device variables but for this test, we could simply exclude the folder from the cpp scan and have a hard-wired cpp CMake file instead of a complicated generic approach.

If I add the `include` support without extra folder for each test, it does mean that all header files for all unit or integration tests are located in a single include folder.

```
unit
|-> include
           |-> testB.hpp
           |-> testC.hpp
|-> testA.cpp
|-> testB.cpp
|-> testC1.cpp
|-> testC2.cpp
|-> testC3.cpp
```

By the way, if I implement sub director support, I would use the folder name as cmake target name.

psychocoderHPC:
all integration tests can be moved to unit test, at the moment I have no real integration tests

psychocoderHPC (approved):

PR #81, opened 2025-03-18
grayscale example simd

mehmetyusufoglu:
Grayscale value calculation from RGB values. R,G and B are 1D arrays, result is grayScaleValues array. Simd feature is used.

PR #90, opened 2025-03-21
Fix DivExZero and Value Type Alias in Vec

mehmetyusufoglu:
1. DivExZero was not returning anything, `return` statement added. 
2.  GetValueType alias is added to Vec.
Because value_type type alias is not defined in Vec.  Standard trait returning value_type did not work.
In Alpaka3:
- `value_type` alias is defined  in  some alpaka structs (like in alpaka mdspan) showing the value type
-  `type` alias is defined showing the  value type in  some alpaka structs (like in onHost::mem::stdContainer)

This PR can be tested by calling DivExZero as in #81

### DIFF
--- a/tests/unit/vec/vec.cpp
+++ b/tests/unit/vec/vec.cpp
@@ -281,13 +281,50 @@ struct CompileTimeKernelCompare2D
     }
 };
 
+/** Compile-time test cases for divCeil and divExZero */
+struct CompileTimeKernelDivCeilAndDivExZero
+{
+    ALPAKA_FN_HOST_ACC void operator()() const
+    {
+        using namespace alpaka;
+
+        // Test divCeil with 1D vectors
+        constexpr auto vec1 = Vec{7};
+        constexpr auto vec2 = Vec{3};
+        // (7 + 3 - 1) / 3 = 9 / 3 = 3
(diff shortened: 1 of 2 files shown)


tests/unit/vec/vec.cpp
```diff
@@ -281,6 +281,54 @@ struct CompileTimeKernelCompare2D
...
+        // Test divCeil with 1D vectors
+        constexpr auto vec1 = Vec{7};
+        constexpr auto vec2 = Vec{3};
+        static_assert(divCeil(vec1, vec2) == Vec{(7 + 3 - 1) / 3}); // (7 + 3 - 1) / 3 = 9 / 3 = 3
+
+        // Test divCeil with 3D vectors
+        constexpr auto vec3 = Vec{3, 7, 5};
+        constexpr auto vec4 = Vec{2, 3, 4};
+        static_assert(divCeil(vec3, vec4) == Vec{(3 + 2 - 1) / 2, (7 + 3 - 1) / 3, (5 + 4 - 1) / 4});
+        // (3 + 2 - 1) / 2 = 4 / 2 = 2
```
psychocoderHPC, line 300: please always comment above the line, the auto formation can make strange things

psychocoderHPC (approved):

PR #92, opened 2025-03-24
remove `std::move()`

psychocoderHPC:
- remove std::move()` for KernelBundle to avoid UB in cases where const members in the kernel functor or arguments are used
- use references to handle `Task` in the interfaces to avoid `std::move()` UB in cases where the task carry const members

### DIFF
--- a/include/alpaka/KernelBundle.hpp
+++ b/include/alpaka/KernelBundle.hpp
@@ -23,28 +23,38 @@ namespace alpaka
     {
     public:
         //! The function object type
-        using KernelFn = TKernelFn;
+        using KernelFn = std::decay_t<TKernelFn>;
         //! Tuple type to encapsulate kernel function argument types and argument values
         using ArgTuple = std::tuple<remove_restrict_t<std::decay_t<TArgs>>...>;
 
         // Constructor
-        constexpr KernelBundle(KernelFn kernelFn, TArgs&&... args)
-            : m_kernelFn{std::move(kernelFn)}
-            , m_args(std::forward<TArgs>(args)...)
+        constexpr KernelBundle(KernelFn const& kernelFn, TArgs const&... args) : m_kernelFn{kernelFn}, m_args(args...)
         {
         }
 
         constexpr KernelBundle(KernelBundle const& b) = default;
+        constexpr KernelBundle& operator=(KernelBundle const&) = default;
 
+        /** allow move assignment and constriction
+         *
+         *  @attention if the functor or the arguments contains non movable types the move operators can be
+         * inaccessible.
+         *
+         *  @{
+         */
         constexpr KernelBundle(KernelBundle&& b) = default;
+        constexpr KernelBundle& operator=(KernelBundle&&) = default;
+
+        /** @} */
 
         constexpr auto operator()(auto const& acc) const
         {
             std::apply([&](auto const&... args) constexpr { m_kernelFn(acc, args...); }, m_args);
         }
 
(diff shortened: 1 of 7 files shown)


include/alpaka/onAcc/Acc.hpp
```diff
@@ -20,7 +20,7 @@ namespace alpaka::onAcc
     template<typename T_Storage>
     struct Acc : T_Storage
     {
-        constexpr Acc(T_Storage const& storage) : T_Storage{storage}
+        constexpr Acc(T_Storage&& storage) : T_Storage{std::forward<T_Storage>(storage)}
```
SimeonEhrig: This does not work, like you expected. `T_Storage` is resolved, before you call the function. Therefore `T_Storage` is no forward reference and this is a move constructor all the time. If this should be a move constructor the following is correct:

```c++
constexpr Acc(T_Storage&& storage) : T_Storage{std::move(storage)}
```

If this should be a constructor with a forward reference as argument, you need to write:

```c++
template<typename T>
requires std::same_as<T, T_Storage>
constexpr Acc(T&& storage) : T_Storage{std::forward<T>(storage)}
```

include/alpaka/KernelBundle.hpp
```diff
@@ -28,23 +28,35 @@ namespace alpaka
...
+        constexpr KernelBundle& operator=(KernelBundle&&) = default;
+
+        /** @} */
 
         constexpr auto operator()(auto const& acc) const
         {
             std::apply([&](auto const&... args) constexpr { m_kernelFn(acc, args...); }, m_args);
         }
 
         KernelFn const m_kernelFn;
```
SimeonEhrig: Should be not const, because with const member variables, the KernelBundle is not moveable anymore.

include/alpaka/KernelBundle.hpp
```diff
@@ -28,23 +28,35 @@ namespace alpaka
         using ArgTuple = std::tuple<remove_restrict_t<std::decay_t<TArgs>>...>;
 
         // Constructor
-        constexpr KernelBundle(KernelFn kernelFn, TArgs&&... args)
-            : m_kernelFn{std::move(kernelFn)}
+        constexpr KernelBundle(KernelFn const& kernelFn, TArgs&&... args)
```
SimeonEhrig:
```suggestion
        constexpr KernelBundle(KernelFn const& kernelFn, auto&&... args)
```

Otherwise `args` are rvalue references.

psychocoderHPC: I will remove `&&` with `const&`

include/alpaka/onAcc/Acc.hpp
psychocoderHPC: will remove the use of `&&` by `const&`

SimeonEhrig:
In `Acc.hpp`, `constexpr Acc(Acc const&&) = delete;` can be deleted. This construtor will be not implicit generated. The move constructor take the xrvalue argument all the time as non-const argument (`constexpr Acc(Acc&&) = delete;`).

include/alpaka/onAcc/Acc.hpp
SimeonEhrig: It is still a move constructor and not a copy constructor.

SimeonEhrig:
@psychocoderHPC Solve the tow issues and the rest looks fine.

SimeonEhrig (approved):

PR #94, opened 2025-03-24
sync alpaka_TESTING to cmake built-in BUILD_TESTING

mehmetyusufoglu:
Problem: BUILD_TESTING and alpaka_TESTING were both visible in ccmake since BUILD_TESTING is cmake build in. But alpaka_TESTING was the correct cmake cache variable to set.
Solution (To be discussed):
`alpaka_TESTING` cache variable is synched to cmake build in `BUILD_TESTING`. It is also converted to a internal variable. 

Only BUILD_TESTING will be visible and settable from command line/cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,12 @@
 cmake_minimum_required(VERSION 3.20)
 project(alpaka CXX)
 
+set(BUILD_TESTING OFF CACHE BOOL "Build the testing tree.")
+include(CTest)
+# Sync alpaka_TESTING internal variable with cache variable BUILD_TESTING. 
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
In principle I refactored the behaviour in https://github.com/alpaka-group/alpaka3/pull/83 as soon as `alpaka_TESTING` is enabled `ctests` worked. CTest currently not works if only `alpaka_EXAMPLES` or `alpaka_BENCHMARKS` is set.

I think only `enable_testing()` must be called in CMake. Let us discuss this in person.

mehmetyusufoglu:
Ol, thanks.

René Widera ***@***.***> schrieb am Di., 25. März 2025,
8:57 AM:

mehmetyusufoglu:
Solved by another PR

PR #105, opened 2025-03-27
add more math functions

mehmetyusufoglu:
While adding random function generators to alpaka3, I noticed some math functions needed for generators were not implemented.

Add math functions (except SYCL backend): `sqrt, cos, log, tan, acos, asin` and `isnan`. There were already 2 functions: sin and exp.

TO DO: 1. Tests. 2. Sycl backend 3. Other   input types of `isnan` be determined.

### DIFF
--- a/include/alpaka/api/unifiedCudaHip/math.hpp
+++ b/include/alpaka/api/unifiedCudaHip/math.hpp
@@ -12,13 +12,13 @@
 #include "alpaka/core/decay.hpp"
 
 #include <cmath>
+#include <concepts>
 
 namespace alpaka::math::internal
 {
 #if ALPAKA_LANG_CUDA || ALPAKA_LANG_HIP
     //! The CUDA sin trait specialization for real types.
-    template<typename T_Arg>
-    requires(std::is_floating_point_v<T_Arg>)
+    template<std::floating_point T_Arg>
     struct Sin::Op<CudaHipMath, T_Arg>
     {
         constexpr auto operator()(CudaHipMath, T_Arg const& arg)
@@ -34,8 +34,7 @@ namespace alpaka::math::internal
         }
     };
 
-    template<typename T_Arg>
-    requires(std::is_floating_point_v<T_Arg>)
+    template<std::floating_point T_Arg>
     struct Exp::Op<CudaHipMath, T_Arg>
     {
         constexpr auto operator()(CudaHipMath, T_Arg const& arg)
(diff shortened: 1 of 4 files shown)


tests/unit/math/mathFunctions.cpp
```diff
@@ -0,0 +1,154 @@
...
+    std::random_device rd{};
+    std::default_random_engine rand{rd()};
+    std::uniform_real_distribution<float> dist{-10.0f, 10.0f};
+
+           // Buffer size
+    constexpr uint32_t size = 1024;
+
+           // Allocate input and output buffers on the host
+    auto in_h = alpaka::onHost::alloc<float>(device, size);
+    auto out_h = alpaka::onHost::alloc<float>(device, size * 8);
```
psychocoderHPC: please move the magic `8` into it's own variable with a proper name. You can also have the variable in the global namespace of the `cpp`-file to use it in the kernel too.

tests/unit/math/mathFunctions.cpp
```diff
@@ -0,0 +1,154 @@
...
+        TAcc const& acc,
+        alpaka::concepts::MdSpan auto const in,
+        alpaka::concepts::MdSpan auto out,
+        uint32_t size) const
+    {
+        for(auto [index] :
+            alpaka::onAcc::makeIdxMap(acc, alpaka::onAcc::worker::threadsInGrid, alpaka::IdxRange{size}))
+        {
+            float x = in[index];
+            out[index * 8 + 0] = alpaka::math::sqrt(x);  // sqrt
```
psychocoderHPC: I suggest to use a `enum` for the enumeration to avoid mixing indces.

```C++
enum MathFunction {
  SQRT,
  COS ,
  LOG,
  ...,
  SIZE
};

Your magic 8 becomes in this case `MathFunction::SIZE`
```

psychocoderHPC (changes requested):

tests/unit/math/mathFunctions.cpp
psychocoderHPC: please do not comment on the end of a line, but with the enum the doc string can be removed anyway

psychocoderHPC (approved):

tests/unit/math/mathFunctions.cpp
mehmetyusufoglu: I have changed. Now every function has a seperate kernel.

tests/unit/math/mathFunctions.cpp
mehmetyusufoglu: I have changed, no need enum anymore. removed bad-formatted comments.

include/alpaka/api/unifiedCudaHip/math.hpp
```diff
@@ -50,6 +50,131 @@ namespace alpaka::math::internal
             ALPAKA_UNREACHABLE(T_Arg{});
         }
     };
+
+    //! The CUDA sqrt trait specialization for real types.
+    template<typename T_Arg>
+    requires(std::is_floating_point_v<T_Arg>)
```
SimeonEhrig:
```suggestion
    template<std::floating_point T_Arg>
```
Please use the concept and apply it to the other functions. You need also to `#include <concepts>`.

mehmetyusufoglu: Done, thanks.

psychocoderHPC (approved):

PR #145, opened 2025-06-12
Improved scan example

AntonReinhard:
This PR adds a second scan example with some more advanced performance improvements over the first one.
I also changed the implementation of inclusive scan to be faster by merging it into the exclusive scan kernel. It's now only a few percent slower than an exclusive scan.

Left to do:
- [x] CLI switch to turn on/off result checks
- [x] CLI switch to test in-place scan vs. inputVec -> outputVec scan
- [x] Add a CUDA library scan call for comparison, similar to the `std::exclusive_scan` currently
- [x] Add a HIP library scan call for comparison


AntonReinhard:
It seems some of the cmake doesn't work entirely correctly yet for finding the cub libraries and reporting if they actually are available.

psychocoderHPC (approved):

PR #144, opened 2025-06-12
Add thread spec to FrameSpec

AntonReinhard:
I also renamed some `...Extent` type parameters to `...Extents` for consistency.

PR #146, opened 2025-06-13
Introduce new concepts

AntonReinhard:
Introduces concepts for
- `FrameSpec`
- `ThreadSpec`
- `ThreadOrFrameSpec`
- `KernelBundle`

Merge after #144

### DIFF
--- a/include/alpaka/KernelBundle.hpp
+++ b/include/alpaka/KernelBundle.hpp
@@ -110,18 +110,21 @@ namespace alpaka
     template<typename TKernelFn, typename... TArgs>
     ALPAKA_FN_HOST KernelBundle(TKernelFn const&, TArgs&&...) -> KernelBundle<TKernelFn, TArgs...>;
 
-    template<typename T>
-    struct IsKernelBundle : std::false_type
+    namespace trait
     {
-    };
+        template<typename T>
+        struct IsKernelBundle : std::false_type
+        {
+        };
 
-    template<typename TKernelFn, typename... TArgs>
-    struct IsKernelBundle<KernelBundle<TKernelFn, TArgs...>> : std::true_type
-    {
-    };
+        template<typename TKernelFn, typename... TArgs>
+        struct IsKernelBundle<KernelBundle<TKernelFn, TArgs...>> : std::true_type
+        {
+        };
+    } // namespace trait
 
     template<typename T>
-    constexpr bool isKernelBundle_v = IsKernelBundle<T>::value;
+    constexpr bool isKernelBundle_v = trait::IsKernelBundle<T>::value;
 
 } // namespace alpaka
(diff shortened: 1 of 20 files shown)


include/alpaka/KernelBundle.hpp
```diff
@@ -104,4 +109,28 @@ namespace alpaka
     //! arguments.
     template<typename TKernelFn, typename... TArgs>
     ALPAKA_FN_HOST KernelBundle(TKernelFn const&, TArgs&&...) -> KernelBundle<TKernelFn, TArgs...>;
+
+    template<typename T>
+    struct IsKernelBundle : std::false_type
```
psychocoderHPC: Please add the namespace `trait` around `IsKernelBundle`

`isKernelBundle_v` should stay in alpaka namespace only

include/alpaka/api/host/Queue.hpp
```diff
@@ -80,10 +80,9 @@ namespace alpaka::onHost
 
             friend struct internal::Enqueue;
 
-            template<alpaka::concepts::Vector T_NumBlocks, alpaka::concepts::Vector T_NumThreads>
             void enqueue(
                 auto const executor,
-                ThreadSpec<T_NumBlocks, T_NumThreads> const& threadBlocking,
+                alpaka::concepts::ThreadSpec auto const& threadBlocking,
```
psychocoderHPC: IMO if we keep the type here this interface can be called like this 

```
enqueue(exec::serial,{Vec{10},Vec {1}},[](){});
```

We should keep in the user interface the type but the changes in `api/*` or  `internal.hpp` are fine.

include/alpaka/api/host/Queue.hpp
```diff
@@ -94,14 +93,10 @@ namespace alpaka::onHost
...
-            template<
-                typename T_Mapping,
-                alpaka::concepts::Vector T_NumFrames,
-                alpaka::concepts::Vector T_FrameExtents,
-                alpaka::concepts::Vector T_ThreadExtents>
+            template<typename T_Mapping>
             void enqueue(
                 T_Mapping const executor,
-                FrameSpec<T_NumFrames, T_FrameExtents, T_ThreadExtents> frameSpec,
+                alpaka::concepts::FrameSpec auto const& frameSpec,
```
psychocoderHPC: same here, we should keep the typed interface.

psychocoderHPC (changes requested):
small changes required

include/alpaka/api/host/Queue.hpp
AntonReinhard: Right, that makes sense, I'll change that

psychocoderHPC (approved):

PR #157, opened 2025-06-23
Single-pass scan algorithm

AntonReinhard:
Work in Progress

The algorithm is correct (at least for CPU and CUDA), but currently uses non-Alpaka memory fences (because they're not yet implemented) and is also slow, because the exclusive block prefix calculation is slow.

If the exclusive prefix calculation is left out, the performance on CUDA is roughly the same as the cuda library implementation, which is promising, since it should be possible to almost entirely hide the latency of this calculation.

Based on https://research.nvidia.com/sites/default/files/pubs/2016-03_Single-pass-Parallel-Prefix/nvr-2016-002.pdf


example/scan/src/scan_single_pass.hpp
```diff
@@ -0,0 +1,525 @@
...
+    constexpr auto chunkExtent = CVec<IdxType, 2048u>{};
+    auto numFrames = divCeil(inputVec.getExtents(), chunkExtent);
+    auto const frameSpec = onHost::FrameSpec{numFrames, chunkExtent, CVec<IdxType, 256u>{}};
+
+    // allocate aggregates, one per block
+    auto flags = onHost::alloc<FlagType<DeviceType>>(devAcc, frameSpec.m_numFrames);
+    auto aggregates = onHost::alloc<Data volatile>(devAcc, frameSpec.m_numFrames);
+    auto prefixes = onHost::alloc<Data volatile>(devAcc, frameSpec.m_numFrames);
+
+    onHost::memset(queue, flags, static_cast<uint8_t>(0));
```
psychocoderHPC: cast to `uint8_t` is not rquired because the interface is enforcing `uint8_t`

include/alpaka/api/unifiedCudaHip/Device.hpp
```diff
@@ -182,7 +182,11 @@ namespace alpaka::onHost
                 auto deviceDependency = onHost::Device{device.getSharedPtr()};
 
                 auto deleter = [ptr, deviceDependency]()
-                { ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(ApiInterface, ApiInterface::free(ptr)); };
+                {
+                    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(
```
psychocoderHPC: can you push this as an seperate PR please

include/alpaka/onHost/Device.hpp
```diff
@@ -26,6 +26,7 @@ namespace alpaka::onHost
         friend struct internal::GetNativeHandle;
 
         using element_type = typename DeviceHandle::element_type;
+        using DeviceKind = T_DeviceKind;
```
psychocoderHPC: this should not be required there is a  `getDeviceKind(any)` you could  use. The type can be queried via
`ALPAKA_TYPEOF(getDeviceKind(any))`

psychocoderHPC (changes requested):

include/alpaka/api/unifiedCudaHip/Device.hpp
AntonReinhard: I opened #202

AntonReinhard:
Rebased on #214 to see if it works

PR #185, opened 2025-07-17
fix lossless convertible

ikbuibui:
Now supports valid unsigned int to signed int conversions and valid integral to floating point type conversions

### DIFF
--- a/include/alpaka/vecConcepts.hpp
+++ b/include/alpaka/vecConcepts.hpp
@@ -12,6 +12,34 @@ namespace alpaka
 {
     namespace concepts
     {
+        namespace detail
+        {
+            // integral to integral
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
you can not convert lossless signed to unsigned or int to float.
For large unsigned numbers the integer range is to small.
For integer to float the same, not all integer numbers can be casted to float without lossing precision.
Do I missed something?

ikbuibui:
This PR only adds unsigned to signed conversion only if it is lossess (for example `unit32` to `int64`) by using information from numeric limits and similarly for valid integral to floating point (it will work for `int` to `double` but not for `int` to `float`)

psychocoderHPC:
I added a unit test to validate the conversions.

PR #195, opened 2025-08-01
icpx shared memory bug search 

psychocoderHPC:


psychocoderHPC:
This test is replaced by the clean implementation #196

PR #202, opened 2025-08-11
Use remove_cv before freeing to allow pointers to volatile memory

AntonReinhard:
Opened from https://github.com/alpaka-group/alpaka3/pull/157#discussion_r2171318420


psychocoderHPC:
I currently implement mapped memroy and will after I mered it apply your fix. I will apply it also to new methods e.g. `AllocMapped`, ...

psychocoderHPC:
I refactored it a little bit. For host allocations I updated `alignedFree()` and for all other usage CUDA/HIP/SYCL I used your way.

PR #214, opened 2025-08-13
Fix alloc of volatile memory

AntonReinhard:
- Add tests for allocating volatile memory
- Also did some spelling fixes

### DIFF
--- a/include/alpaka/api/syclGeneric/Device.hpp
+++ b/include/alpaka/api/syclGeneric/Device.hpp
@@ -104,14 +104,7 @@ namespace alpaka::onHost
 
                 T_Type* ptr = reinterpret_cast<T_Type*>(
                     sycl::aligned_alloc_device(alignment, memSizeInByte, sycl_device, sycl_context));
-                auto deleter = [ctx = sycl_context, ptr]()
-                {
-                    void* ptrToFree = reinterpret_cast<void*>(
-                        const_cast<
-                            std::add_pointer_t<std::remove_volatile_t<std::remove_pointer_t<ALPAKA_TYPEOF(ptr)>>>>(
-                            ptr));
-                    sycl::free(ptrToFree, ctx);
-                };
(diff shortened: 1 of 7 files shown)


AntonReinhard:
The lines have become kind of monstrous now, if someone has a better idea, I'm all ears. But it's currently necessary to remove the pointer, then do `remove_volatile`, and then add the pointer again, because `remove_volatile` alone doesn't work on pointer types:
> Removing const/volatile from const volatile int* does not modify the type, because the pointer itself is neither const nor volatile.

From [here](https://en.cppreference.com/w/cpp/types/remove_cv.html#:~:text=Removing%20const%2Fvolatile%20from%20const%20volatile%20int*%20does%20not%20modify%20the%20type%2C%20because%20the%20pointer%20itself%20is%20neither%20const%20nor%20volatile.)

include/alpaka/api/syclGeneric/Device.hpp
```diff
@@ -106,8 +106,10 @@ namespace alpaka::onHost
                     sycl::aligned_alloc_device(alignment, memSizeInByte, sycl_device, sycl_context));
                 auto deleter = [ctx = sycl_context, ptr]()
                 {
-                    void* ptrToFree
-                        = reinterpret_cast<void*>(const_cast<std::remove_volatile_t<ALPAKA_TYPEOF(ptr)>>(ptr));
+                    void* ptrToFree = reinterpret_cast<void*>(
+                        const_cast<
```
psychocoderHPC: could you introduce in `util.hpp`

```C++
template<typename T>
using RemoveVolatileFromPointee_t = std::add_pointer_t<std::remove_volatile_t<std::remove_pointer_t<T>>>;
```

But the other question is, we cast the pointer anyway to `void*` why do we need to remove the volatile before?

psychocoderHPC (changes requested):

psychocoderHPC (approved):

PR #218, opened 2025-08-18
host task enqueue for all backends

psychocoderHPC:
- add host task support for CUDA, HIP and OneApi.
- add tests from alpaka mainline to check the feature


include/alpaka/api/oneApi/Queue.hpp
```diff
@@ -111,7 +111,7 @@ namespace alpaka::onHost::internal
     struct internal::Fill::Op<syclGeneric::Queue<T_Device>, T_Dest, T_Value, T_Extents>
     {
         void operator()(
-            syclGeneric::Queue<T_Device> queue,
+            syclGeneric::Queue<T_Device>& queue,
```
psychocoderHPC, line 114: small fix for bug which was hidden until now.

Issue #84, opened 2025-03-19
list of missing features compared to mainline alpaka


psychocoderHPC:
# general
  - [x] random numbers #107 #365 #386
  - [x] INTEL TBB support #321
  - [x] SYCL support #58
  - [x]  SYCL global device nemory #340 #345
  - [ ] host only compile (allow alloc,memset,...) without the backend compiler e.g. nvcc
  - [ ] standalone header
    - In principle, this feature should not be required because it is simply possible to include `alpaka/alpaka.hpp` because backends are enabled in C++ and not via CMake options or preprocessor macros.
    - any example can be copied into godbolt: https://alpaka3.readthedocs.io/en/latest/basic/example.html#play-online-with-compiler-explorer
  - [ ] cuda or hip only mode (question is, should we still support it, if yes why?)
  - [ ] GitLab CI
  - [x] version numbers, including CMake #376 

# onHost
  - [x] events #222 
  - [x] wait(device)  #221
  - [x] non owning Views of data
  - [x] SubViews of data/Views
  - [x] enqueue host task on all backends #218
  - [ ] OpenMP scheduling selection
  - [ ] is it possible to write an queue which behaves like https://github.com/alpaka-group/alpaka/blob/25bb7b2e1fa29005e753883257d82e84e0610f09/test/unit/queue/src/CollectiveQueue.cpp#L20
  - [x] `allocAsync()` #192
  - [x] mapped memory allocation #204 and managed memory #203
  - [x] blocking queues #229

# onAcc
  - [x] warp operations #339 #353
  - [x] threadfence #258
  - [X] atomic definition per hierarchy #360
    - we can currently not transform `CpuOmpBlocks` atomics between threads to a non atomic operation
  - [x] intrinsic `popc` #369

mehmetyusufoglu:
random numbers: philox integer uniform generator and unifrom real transporter: #107
Box-Müller transformer from uniform to normal real distribution : #111

SimeonEhrig:
@psychocoderHPC Please add 
- new CMake
- finalizing managed views and buffers

mehmetyusufoglu:
Blocking Queue: https://github.com/alpaka-group/alpaka3/pull/229

mehmetyusufoglu:
[...]

PR #221, opened 2025-08-21
Add alpaka::wait(device) synchronization helper and tests

mehmetyusufoglu:
Issue is selected from[ "Features to be implemented list"](https://github.com/alpaka-group/alpaka3/issues/84). Device-scoped wait. Locally only tested with cpu-cuda.
2 tests are needed. 
 -  multiple queues 
  -  single queue but async operations ( Use  streamNonBlocking, memcpyAsync  kernelLaunch would return immediately?)
  ## Merge requirements
  - [x] Squash commits
  - [x] Test for HIP 

## Runs
```

alpaka3/build-cpu-gpu$  ./tests/unit_test_waitDevice
Randomness seeded to: 2085127153
device spec: Host Cpu
executor   : CpuOmpBlocks
device name: 13th Gen Intel(R) Core(TM) i7-1360P id=0 (idx=0)
✓ wait(device) verified with 3 concurrent queues + 3 ops on single queue
device spec: Host Cpu
executor   : CpuSerial
device name: 13th Gen Intel(R) Core(TM) i7-1360P id=0 (idx=0)
✓ wait(device) verified with 3 concurrent queues + 3 ops on single queue
device spec: Host Cpu
executor   : CpuOmpBlocksAndThreads
device name: 13th Gen Intel(R) Core(TM) i7-1360P id=0 (idx=0)
✓ wait(device) verified with 3 concurrent queues + 3 ops on single queue
device spec: Cuda NvidiaGpu
executor   : GpuCuda
device name: NVIDIA RTX A500 Laptop GPU id=0 (idx=0)
✓ wait(device) verified with 3 concurrent queues + 3 ops on single queue
===============================================================================
All tests passed (28 assertions in 4 test cases)

Local icpx Sycl backend test:

alpaka3/build-sycl$ cd /home/yusufo81/projects/alpaka-dir/other/alpaka3/build-sycl/tests && echo "=== Running waitDevice test with SYCL support ===" && ./unit_test_waitDevice
=== Running waitDevice test with SYCL support ===
Randomness seeded to: 294609139
device spec: Host Cpu
executor   : CpuOmpBlocks
device name: 13th Gen Intel(R) Core(TM) i7-1360P id=0 (idx=0)
✓ wait(device) verified with 3 concurrent queues + 3 ops on single queue
device spec: Host Cpu
executor   : CpuSerial
device name: 13th Gen Intel(R) Core(TM) i7-1360P id=0 (idx=0)
✓ wait(device) verified with 3 concurrent queues + 3 ops on single queue
device spec: Host Cpu
executor   : CpuOmpBlocksAndThreads
device name: 13th Gen Intel(R) Core(TM) i7-1360P id=0 (idx=0)
✓ wait(device) verified with 3 concurrent queues + 3 ops on single queue
device spec: OneApi Cpu
executor   : OneApi
device name: 13th Gen Intel(R) Core(TM) i7-1360P (idx=0)
✓ wait(device) verified with 3 concurrent queues + 3 ops on single queue
No device available for OneApi IntelGpu
===============================================================================
All tests passed (28 assertions in 5 test cases)
```

### DIFF
--- a/include/alpaka/api/host/Device.hpp
+++ b/include/alpaka/api/host/Device.hpp
@@ -53,6 +53,24 @@ namespace alpaka::onHost
                 return m_idx != other.m_idx;
             }
 
+            void wait()
+            {
+                // Host device synchronization - wait on all queues associated with this device.
+                // IMPORTANT: Do not hold queuesGuard across potentially long waits; copy weak refs first.
+                std::vector<std::weak_ptr<cpu::Queue<Device>>> tmpQueues;
+                {
+                    std::lock_guard<std::mutex> lk{queuesGuard};
+                    tmpQueues = queues; // copy weak_ptr list
+                }
+                for(auto& weakQueue : tmpQueues)
+                {
+                    if(auto queue = weakQueue.lock())
+                    {
+                        internal::wait(*queue);
+                    }
+                }
+            }
+
         private:
             void _()
             {
--- a/include/alpaka/api/syclGeneric/Device.hpp
+++ b/include/alpaka/api/syclGeneric/Device.hpp
@@ -64,6 +64,23 @@ namespace alpaka::onHost
                 return {m_sycl_dev, m_platform->getContext()};
             }
 
+            void wait()
+            {
+                // Copy queue weak refs under lock then release to avoid blocking other operations while waiting.
+                std::vector<std::weak_ptr<syclGeneric::Queue<Device>>> tmpQueues;
+                {
+                    std::lock_guard<std::mutex> lk{m_writeGuard};
+                    tmpQueues = queues;
+                }
+                for(auto& weakQueue : tmpQueues)
+                {
+                    if(auto queue = weakQueue.lock())
+                    {
+                        queue->wait();
+                    }
+                }
+            }
+
         private:
             friend struct internal::MakeEvent;
 
@@ -92,9 +109,11 @@ namespace alpaka::onHost
             Handle<T_Platform> m_platform;
             uint32_t m_idx = 0u;
             sycl::device m_sycl_dev;
+
             std::vector<std::weak_ptr<syclGeneric::Queue<Device>>> queues;
             std::vector<std::weak_ptr<syclGeneric::Event<Device>>> events;
             std::mutex m_writeGuard;
+
             DeviceProperties m_properties;
 
             friend struct alpaka::internal::GetApi;
--- a/tests/unit/wait/waitDevice.cpp
+++ b/tests/unit/wait/waitDevice.cpp
@@ -1,0 +1,144 @@
+/* Copyright 2025 Mehmet Yusufoglu, René Widera
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#include <alpaka/alpaka.hpp>
+#include <alpaka/onHost/example/executors.hpp>
+#include <alpaka/onHost/executeForEach.hpp>
+
+#include <catch2/catch_template_test_macros.hpp>
+#include <catch2/catch_test_macros.hpp>
+
+using namespace alpaka;
+
+// any value for testing
+int const testValue = 49;
+
+using TestApis = std::decay_t<decltype(onHost::allBackends(onHost::enabledApis, onHost::example::enabledExecutors))>;
+
+// Simple write kernel used to create actual device work for testing wait(device).
+struct WaitTestWriteKernel
+{
+    template<typename TAcc, typename TBuf>
+    ALPAKA_FN_ACC void operator()(TAcc const&, TBuf out) const
+    {
+        out[0] = testValue;
+    }
+};
+
+template<typename T_DataType>
+void executeWaitTest(auto& device, auto const& exec)
+{
+    using DataType = T_DataType;
+
+    constexpr uint32_t num = 3;
+    // for multiple concurrent queues
+    constexpr uint32_t numQueues = num;
+    // for multiple ops at a single queue test
+    constexpr uint32_t numOpsPerQueue = num;
+
+    // Test 1: Multiple concurrent queues
+
+    std::vector<decltype(device.makeQueue())> queues;
+    // Create host and device buffers
+    std::vector<decltype(onHost::alloc<DataType>(device, Vec<uint32_t, 1>{1u}))> devBufs;
+    std::vector<decltype(onHost::allocHostLike(onHost::alloc<DataType>(device, Vec<uint32_t, 1>{1u})))> hostBufs;
+
+    // allocate memory and create queues
+    constexpr auto extent = Vec<uint32_t, 1>{1u};
+    for(uint32_t i = 0; i < numQueues; ++i)
+    {
+        queues.push_back(device.makeQueue());
+        devBufs.push_back(onHost::alloc<DataType>(device, extent));
+        hostBufs.push_back(onHost::allocHostLike(devBufs.back()));
+    }
+
+    // enqueue work on multiple queues concurrently
+    constexpr auto frameSize = Vec<uint32_t, 1>{1u};
+    for(uint32_t i = 0; i < numQueues; ++i)
+    {
+        auto kernelBundle = KernelBundle{WaitTestWriteKernel{}, devBufs[i]};
+        queues[i].enqueue(exec, onHost::FrameSpec{extent, frameSize}, kernelBundle);
+        // Note: We deliberately don't wait on individual queues here
+    }
+
+    // Test 2: Multiple async operations on SAME queue
+    auto singleQueue = device.makeQueue();
+
+    // Create host and device buffers
+    std::vector<decltype(onHost::alloc<DataType>(device, extent))> singleQueueBufs;
+    std::vector<decltype(onHost::allocHostLike(onHost::alloc<DataType>(device, extent)))> singleQueueHostBufs;
+
+    for(uint32_t i = 0; i < numOpsPerQueue; ++i)
+    {
+        singleQueueBufs.push_back(onHost::alloc<DataType>(device, extent));
+        singleQueueHostBufs.push_back(onHost::allocHostLike(singleQueueBufs.back()));
+
+        // Enqueue multiple operations on same queue - they queue up asynchronously
+        auto kernelBundle = KernelBundle{WaitTestWriteKernel{}, singleQueueBufs[i]};
+        singleQueue.enqueue(exec, onHost::FrameSpec{extent, frameSize}, kernelBundle);
+    }
+
+    // TEST: wait for device to complete ALL work on ALL queues (including single queue ops)
+    REQUIRE_NOTHROW(onHost::wait(device));
+
+    // copy results of device buffers to host buffers
+    for(uint32_t i = 0; i < numQueues; ++i)
+    {
+        onHost::memcpy(queues[numQueues - i - 1], hostBufs[i], devBufs[i]);
+    }
+    for(uint32_t i = 0; i < numOpsPerQueue; ++i)
+    {
+        onHost::memcpy(singleQueue, singleQueueHostBufs[i], singleQueueBufs[i]);
+    }
+
+    // TEST: wait for device to complete ALL copies on ALL queues (including single queue ops)
+    REQUIRE_NOTHROW(onHost::wait(device));
+
+    // verify results of different queues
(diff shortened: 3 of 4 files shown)


mehmetyusufoglu:
There is an[ CI error](https://github.com/alpaka-group/alpaka3/actions/runs/17151389515/job/48658301822?pr=221) at icpx. We had an empty tuple and zero executor at `supportedMappings(device) `(trait.hpp line 145) . This case should be managed somehow.

CI had `-DALPAKA_DISABLE_EXEC_CpuSerial `= ON (and other executors disabled)
possible this made` supportedMappings()` return an empty tuple std::tuple<>
std::get<0>(executor) on an empty tuple ?

`  ci.yml` sets (line 136) sets only one DEP (ON); but sets `ONEAPI_IntelGpu` to OFF ?? 
  
   ${{ matrix.config.compiler == 'icpx' && '-DCMAKE_CXX_COMPILER=icpx -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64_x86_64" -Dalpaka_DEP_ONEAPI=ON -Dalpaka_ONEAPI_IntelGpu=OFF -Dalpaka_EXEC_CpuSerial=OFF -Dalpaka_DEP_OMP=OFF' || '' }}  
  
It seems `supportedMappings(device) (trait.hpp line 145) `filters the executors based on whether they're supported by the specific device. If the device/backend combination doesn't support the oneApi executor, it gets removed or filtered out but i am not sure. At least we can put a static assert to check the tuple size. I am investigating ....

include/alpaka/wait.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace alpaka
+{
+    // Concept: type exposes a deviceSynchronize() member
+    template<typename T>
+    concept HasDeviceSynchronize = requires(T t) { t.deviceSynchronize(); };
+
+    // Concept: a realized device (not a DeviceSpec)
+    template<typename T>
+    concept DeviceLike = requires(T& d) {
```
psychocoderHPC: There is already a concept for the device available https://github.com/alpaka-group/alpaka3/blob/d50605ed3fd366c7bd0933731269b19c5f23df6a/include/alpaka/onHost/concepts.hpp#L19
Never the less concepts always placed into a namepace called concepts.

include/alpaka/wait.hpp
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * Backend behavior:
+     * - Host CPU: No-op (synchronous execution)
+     * - CUDA/HIP: Calls deviceSynchronize() via API
+     * - SYCL: Calls deviceSynchronize() via API
+     *
+     * @param device The device to synchronize
+     */
+    template<DeviceLike Device>
+    inline void wait(Device& device)
```
psychocoderHPC: each user interface function is going over an internal function or directly to an internal operation.
For Wait there is already a trait: https://github.com/alpaka-group/alpaka3/blob/d50605ed3fd366c7bd0933731269b19c5f23df6a/include/alpaka/onHost/internal.hpp#L103-L113

This means you need only specialize the operation for a device or the  internaldevice implementation must have a methid `wait()`
https://github.com/alpaka-group/alpaka3/blob/d50605ed3fd366c7bd0933731269b19c5f23df6a/include/alpaka/api/unifiedCudaHip/Queue.hpp#L112-L115

include/alpaka/wait.hpp
```diff
@@ -0,0 +1,60 @@
...
+    {
+        auto api = getApi(device);
+        if constexpr(HasDeviceSynchronize<decltype(api)>)
+        {
+            api.deviceSynchronize();
+        }
+        else
+        {
+            // Final fallback: no-op for synchronous backends like CPU
```
psychocoderHPC: Each device must implement the wait, all APIs should behave equal. If the CPU is not performing a wait the implementation should be empty. But this would be wrong because the CPU hae a queue executed in a different thread. We need to synchronize all queues to synchronize a host device.
This is equal to the mainline alpaka implemenation.

psychocoderHPC (changes requested):

tests/unit/wait/waitDevice.cpp
```diff
@@ -0,0 +1,110 @@
...
+struct WaitTestWriteKernel
+{
+    template<typename TAcc, typename TBuf>
+    ALPAKA_FN_ACC void operator()(TAcc const&, TBuf out) const
+    {
+        out[0] = testValue;
+    }
+};
+
+TEST_CASE("wait all backends", "[wait]")
```
psychocoderHPC: please check this test case as base https://github.com/alpaka-group/alpaka3/blob/d50605ed3fd366c7bd0933731269b19c5f23df6a/tests/unit/algorithm/concurrent.cpp#L20
and use `TEMPLATE_LIST_TEST_CASE` to start the test for each backend.

psychocoderHPC:
I opened a issue #223 for the bug you mentioned https://github.com/alpaka-group/alpaka3/pull/221#issuecomment-3213899590

include/alpaka/wait.hpp
mehmetyusufoglu: ok, good design. I should have checked. Thanks.

include/alpaka/api/host/Device.hpp
```diff
@@ -52,6 +52,11 @@ namespace alpaka::onHost
                 return m_idx != other.m_idx;
             }
 
+            void wait()
+            {
+                // Host device operations are synchronous, so no-op
```
psychocoderHPC: host operations are not synchonious. You need to itwrate over the list of streams and call wait on each stream. I can explain it next week if you like.

include/alpaka/api/syclGeneric/Device.hpp
```diff
@@ -62,6 +62,20 @@ namespace alpaka::onHost
                 return {m_sycl_dev, m_platform->getContext()};
             }
 
+            void wait()
+            {
+                // SYCL device synchronization following original alpaka approach
+                // Wait on all queues associated with this device
+                std::lock_guard<std::mutex> lk{queuesGuard};
+                for(auto& weakQueue : queues)
```
psychocoderHPC: The same you must do for the host device too.

include/alpaka/wait.hpp
psychocoderHPC: there is no need to define an wait method because there is already one https://github.com/alpaka-group/alpaka3/blob/8064fdb62e640bc8bc1a096e88142492d8f933da/include/alpaka/onHost.hpp#L89

Btw methods which can only be called on the host side are always in the namespace `onHost`

If you remove this file the wait call with a device as argument should work out of the box.

psychocoderHPC (changes requested):
minor changes but you are on a good way. thanks for looking into it

include/alpaka/wait.hpp
mehmetyusufoglu: yes, done.

tests/unit/wait/waitDevice.cpp
mehmetyusufoglu: thanks, done

include/alpaka/api/host/Device.hpp
mehmetyusufoglu: ok i implemented. i will check the code at least. thanks.

mehmetyusufoglu:
> There is an[ CI error](https://github.com/alpaka-group/alpaka3/actions/runs/17151389515/job/48658301822?pr=221) at icpx. We had an empty tuple and zero executor at `supportedMappings(device) `(trait.hpp line 145) . This case should be managed somehow.
> 
> CI had `-DALPAKA_DISABLE_EXEC_CpuSerial `= ON (and other executors disabled) possible this made` supportedMappings()` return an empty tuple std::tuple<> std::get<0>(executor) on an empty tuple ?
> 
> ` ci.yml` sets (line 136) sets only one DEP (ON); but sets `ONEAPI_IntelGpu` to OFF ??
> 
> ${{ matrix.config.compiler == 'icpx' && '-DCMAKE_CXX_COMPILER=icpx -DCMAKE_CXX_FLAGS="-fsycl -fsycl-targets=spir64_x86_64" -Dalpaka_DEP_ONEAPI=ON -Dalpaka_ONEAPI_IntelGpu=OFF -Dalpaka_EXEC_CpuSerial=OFF -Dalpaka_DEP_OMP=OFF' || '' }}
> 
> It seems `supportedMappings(device) (trait.hpp line 145) `filters the executors based on whether they're supported by the specific device. If the device/backend combination doesn't support the oneApi executor, it gets removed or filtered out but i am not sure. At least we can put a static assert to check the tuple size. I am investigating ....

This CI error seems to be not happening after the the main dev branch [commit](https://github.com/alpaka-group/alpaka3/commit/580904768b59e2debb1e91f197a13bb254e48e4d) yesterday ( i rebased) .  But still needs to be investigated.

include/alpaka/api/syclGeneric/Device.hpp
mehmetyusufoglu: ok , done. thanks.

include/alpaka/api/host/Device.hpp
```diff
@@ -53,6 +53,19 @@ namespace alpaka::onHost
                 return m_idx != other.m_idx;
             }
 
+            void wait()
+            {
+                // Host device synchronization - wait on all queues associated with this device
+                std::lock_guard<std::mutex> lk{queuesGuard};
```
psychocoderHPC: we should not put the waits under the lock, it can have side effects because we block in this case all other device actions from other threads too.
We should change it to the following for this and other APIs
```C++
std::vector<std::weak_ptr<cpu::Queue<Device>>> tmpQueues;
{
    std::lock_guard<std::mutex> lk{queuesGuard};
    // make a copy to avoid blocking the full device while we wait later
    tmpQueues = queues;
}
for(auto& weakQueue : tmpQueues)
{
    if(auto queue = weakQueue.lock())
    {
        internal::wait(*queue);
    }
}
```

psychocoderHPC (changes requested):

include/alpaka/api/host/Device.hpp
mehmetyusufoglu: ok, we assume that any other queue is not created after we copied. done thanks.

psychocoderHPC: The definition of wait is that all work submitted before must be finished when the wait function is finished and the handle is given back ro the caller thread. If during the wait another thread creates a new queue an submit work it is fine that this work is jot taken into account.

mehmetyusufoglu: ok done.

psychocoderHPC:
please rebase against the latest dev branch that the documenation CI job is able to finish.

tests/unit/wait/waitDevice.cpp
```diff
@@ -0,0 +1,127 @@
...
+        singleQueue.enqueue(exec, onHost::FrameSpec{extent, frameSize}, kernelBundle);
+    }
+
+    // TEST: wait for device to complete ALL work on ALL queues (including single queue ops)
+    REQUIRE_NOTHROW(onHost::wait(device));
+
+    // verify all kernels executed correctly on all queues
+    for(uint32_t i = 0; i < numQueues; ++i)
+    {
+        onHost::memcpy(queues[i], hostBufs[i], devBufs[i]);
```
psychocoderHPC: this memcpy should be in the for loop where you push the work.
The reason is that `onHost::wait(queues[i]);` is syncronizing each queue and even if you skip wait(device) all will work.

tests/unit/wait/waitDevice.cpp
```diff
@@ -0,0 +1,127 @@
...
+        onHost::wait(queues[i]);
+        REQUIRE(hostBufs[i][0] == testValue);
+    }
+
+    // verify all operations on single queue completed
+    for(uint32_t i = 0; i < numOpsPerQueue; ++i)
+    {
+        onHost::memcpy(singleQueue, singleQueueHostBufs[i], singleQueueBufs[i]);
```
psychocoderHPC: should be moved into a for loop without per queue wait and afterwards a wait(device) shoud be added

psychocoderHPC (changes requested):

psychocoderHPC (approved):

psychocoderHPC (approved):

PR #229, opened 2025-08-25
 Blocking queue implementation

mehmetyusufoglu:
Add alpaka style `blocking-queue` implementation.

Design: 
- API uses tags
```
    auto qBlocking = device.makeQueue(queueKind::blocking);
    auto qNonBlocking = device.makeQueue(queueKind::nonBlocking);
```
- internally boolean is used ` bool m_isBlocking{false};`
- `submit` function is used to create an abstraction to check blocking/non-blocking case

To decide (decided already): 
- Should a blocking queue also block async operations like `allocAsync`? (assumed Yes in PR) 
- Global factory functions is a policy of alpaka3 like `makeBlockingQueue`? (added in PR but not necessary i guess) -> NO, CHANGED


psychocoderHPC:
IMO we should not create the blocking queue via makeBlockingQueue(), I would prefere makeQueue(policy)  the reason is that we would like to support user defined queues too. Thisnis a feature required by DLR. It is also an open point in the list with features missing compared to mainlien alpaka.
By default policy should be non blocking.

psychocoderHPC:
I will provide tomorrow more information you should best take into account.

psychocoderHPC:
> Should a blocking queue also block async operations like allocAsync? (assumed Yes in PR)

No please not the name is clear saying how it behaves therefore this should be async.

psychocoderHPC:
My suggestion:

- do not use a template parameter for the internal queue implementations
  - use an member `bool m_isBlocking`
  - for Host queue implement a member function 
    ```
    void submit(auto const & task)
    {
        if (m_isBlocking)
            task();
        else
            m_workerThread.submit(task);
    }
    ```
    - there are many traits I speezialized where I use somthink like `queue.m_workerThread.submit(...)` this should be changed to `queue.submit(...)` to handle blocking and non blocking tasks.
    - the reason why I would not use a template parameter at the moment is, you do not need to go to the full code and seach for all spezializations.
    - the performance overhead of the if should be near zero
- for cuda/hip
  - use local variable ` m_isBlocking` too but cuda kernel start, memcopy are always synchronous so after each call call `wait()` onto the queue
- for sycl you can call write
  ```
  [[maybe_unused]] sycl::event ev = sycl::memcopy();
  if(m_isBlocking)
     ev.wait();
  ``` 

You should only update the `MakeQueue` trait and pass true for blocking an false for nonblocking to the updated queue constructor. There is no need at the moment to make everything DLR queue ready, how to use such a queue is currently not fully clear.

include/alpaka/onHost/Device.hpp
```diff
@@ -230,10 +251,12 @@ namespace alpaka::onHost
...
      */
-    template<typename T_Type, typename T_Device>
-    inline auto allocMapped(Queue<T_Device> const& queue, alpaka::concepts::VectorOrScalar auto const& extents)
+    template<typename T_Type, typename T_Device, typename T_BlockingPolicy>
+    inline auto allocMapped(
+        Queue<T_Device, T_BlockingPolicy> const& queue,
+        alpaka::concepts::VectorOrScalar auto const& extents)
     {
-        allocMapped<T_Type>(queue.getDevice(), extents);
+        return allocMapped<T_Type>(queue.getDevice(), extents);
```
psychocoderHPC, line 257: I will open a PR for this bug, well spotted

psychocoderHPC:
please rebase your PR as asap against https://github.com/alpaka-group/alpaka3/pull/237 is merged and run the git precommit actions to avoid rebase conflics

mehmetyusufoglu:
@psychocoderHPC  using policy namespace is ok for you?    
    
```
    auto qNonBlocking = device.makeQueue(alpaka::onHost::policy::nonBlocking);
    auto qBlocking = device.makeQueue(alpaka::onHost::policy::blocking);
```

mehmetyusufoglu:
ready to review.

include/alpaka/api/syclGeneric/Queue.hpp
```diff
@@ -164,6 +188,20 @@ namespace alpaka::onHost
     {
         void operator()(syclGeneric::Queue<T_Device>& queue, T_Event& event) const
         {
+            // Blocking queue: ensure all prior work is complete and create an already-complete event.
+            if(queue.isBlocking())
```
mehmetyusufoglu, line 183: @fwyzard  Since sycl had no low level synchronization like cuda;  a trivial task and an event is generated to make sure sychronization is done. Is this ok in your opinion?

fwyzard: If I understood correctly what you are aiming for, oneAPI provides this functionality via the `sycl_ext_oneapi_enqueue_barrier` extension: https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_enqueue_barrier.asciidoc

fwyzard: Otherwise, assuming we use only in-order queues, I think it might be better to cache in the alpaka queue the last sycl event returned by a SYCL operation on the queue.

In fact, there seems to be an experimental oneAPI extension that does it already: see `sycl_ext_oneapi_in_order_queue_events` at https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_in_order_queue_events.asciidoc .

fwyzard: ---
On a related topic: last time we checked, not all SYCL events work correctly on all backends.

Please test that events coming from
  - empty kernels
  - memory set/fill
  - memory copies
  - the barrier extension
  
properly wait and synchronise on Intel CPUs, Intel GPUs, NVIDIA GPUs and AMD GPUs, before relying on them in alpaka.

fwyzard:
>     * To Decide: Device Reset at Cuda/Hip backends unconditionally ?

What do you mean by "device reset" ?

If you mean something like calling `cudaDeviceReset`, that should _never_ be called by a library.

> To decide (decided already):
> 
>     * Should a blocking queue also block async operations like `allocAsync`? (assumed Yes in PR) -> NO! CHANGED

Then it was decided wrong.
_Of course_ a blocking queue has to block async operations like `allocAsync`.

include/alpaka/api/syclGeneric/Queue.hpp
mehmetyusufoglu: What i understood from your comment is we need this:

```
 // Helper method to get/create sync event
    sycl::event getBlockingSyncEvent() const {
#ifdef SYCL_EXT_ONEAPI_ENQUEUE_BARRIER
        // First choice: oneAPI barrier extension, check: all sycl implementations have it? 
        return m_queue.ext_oneapi_submit_barrier();
        
#elif defined(SYCL_EXT_ONEAPI_IN_ORDER_QUEUE_EVENTS)
        // Get last event from in-order queue (Intel extension...)
        sycl::event ev = m_queue.ext_oneapi_get_last_event();
        
        // If queue is empty (no last event), create one via memset BUT CHECK MEMSET!!!
        if(!ev.is_valid()) {
            ev = m_queue.submit([&](sycl::handler& cgh) {
                auto acc = m_syncBuffer.get_access<sycl::access::mode::write>(cgh);
                cgh.fill(acc, uint8_t{0});
            });
            // Cache this new event
            m_cachedEvent = ev;
        }
        return ev;
        
#else
        // Fallback: Manual event caching + memset
        // Check if we need a new event
        if(!m_cachedEvent || 
           m_cachedEvent->get_info<sycl::info::event::command_execution_status>() 
           == sycl::info::event_command_status::complete) {
            
            // Create new sync event via memset, but check MEMSET
            m_cachedEvent = m_queue.submit([&](sycl::handler& cgh) {
                auto acc = m_syncBuffer.get_access<sycl::access::mode::write>(cgh);
                cgh.fill(acc, uint8_t{0});
            });
        }
        return *m_cachedEvent;
#endif
    }
```

mehmetyusufoglu:
> > ```
> > * To Decide: Device Reset at Cuda/Hip backends unconditionally ?
> > ```
> 
> What do you mean by "device reset" ?
> 
> If you mean something like calling `cudaDeviceReset`, that should _never_ be called by a library.
> 
> > To decide (decided already):
> > ```
> > * Should a blocking queue also block async operations like `allocAsync`? (assumed Yes in PR) -> NO! CHANGED
> > ```
> 
> Then it was decided wrong. _Of course_ a blocking queue has to block async operations like `allocAsync`.

@psychocoderHPC  What is your opinion about "Should a blocking queue also block async operations like `allocAsync`?"

mehmetyusufoglu:
> > ```
> > * To Decide: Device Reset at Cuda/Hip backends unconditionally ?
> > ```
> 
> What do you mean by "device reset" ?
> 
> If you mean something like calling `cudaDeviceReset`, that should _never_ be called by a library.
> 
> > To decide (decided already):
> > ```
> > * Should a blocking queue also block async operations like `allocAsync`? (assumed Yes in PR) -> NO! CHANGED
> > ```
> 
> Then it was decided wrong. _Of course_ a blocking queue has to block async operations like `allocAsync`.

@fwyzard  `cudaDeviceReset` is not directly related with this PR, sorry. I just found while changing the file  [`unifiedCudaHip/Device.hpp`](https://github.com/alpaka-group/alpaka3/blob/c7721727411dc0d49db750f6c3d2b89c24e9e519/include/alpaka/api/unifiedCudaHip/Device.hpp#L45)

include/alpaka/api/syclGeneric/Queue.hpp
fwyzard: What SYCL back-ends do you plan to support ?
Rather than implementing all possibilities, you could check if they all support `SYCL_EXT_ONEAPI_ENQUEUE_BARRIER` or `SYCL_EXT_ONEAPI_IN_ORDER_QUEUE_EVENTS` and skip the rest.

include/alpaka/api/syclGeneric/Queue.hpp
```diff
@@ -164,8 +242,30 @@ namespace alpaka::onHost
...
-            sycl::event emulatedEvent = queue.m_queue.submit([](sycl::handler& cgh) { cgh.single_task([]() {}); });
-            event.setEvent(emulatedEvent);
+            // Blocking queue: ensure all prior work is complete and create an already-complete event.
+            if(queue.isBlocking())
+            {
+                // Submit a trivial task to capture ordering, then wait for it -> event is immediately complete.
+                // We need an event whose completion implies “everything before is done”. The empty kernel becomes a
+                // fence point.
+                // Use oneAPI barrier extension (50x faster than empty kernels)
+#    ifdef SYCL_EXT_ONEAPI_ENQUEUE_BARRIER
```
psychocoderHPC: Please keep the original code for the PR. 
As @fwyzard wrote, this is a oneAPI extension; if we introduce it, we need to move the code into the oneAPI folder anyway.
Nevertheless, the current code is tested, and we should avoid refactorings while introducing an independent feature like blocking queues.

include/alpaka/api/syclGeneric/Queue.hpp
```diff
@@ -164,8 +242,30 @@ namespace alpaka::onHost
...
+                sycl::event ev = queue.m_queue.ext_oneapi_submit_barrier();
+#    else
+                // Fallback: Host task synchronization (no empty kernels)
+                // Some SYCL implementations may not have the barrier extension
+                queue.m_queue.wait();
+                // Submit a trivial host task to act as a fence
+                sycl::event ev = queue.m_queue.submit([](sycl::handler& cgh) { cgh.host_task([](){}); });
+#    endif
+                // We must hand back an event whose dependence enforces all prior queue work.
+                ev.wait_and_throw();
```
psychocoderHPC: please call first `event.setEvent(ev);` and than call wait. If another thread is using the same event the current event state will not be seen.

psychocoderHPC (changes requested):
@mehmetyusufoglu If all works please give me a sign. I will touch few parts I would do different after your changes are in. This is simpler than X iterations.

include/alpaka/api/syclGeneric/Queue.hpp
psychocoderHPC: @fwyzard `return m_queue.ext_oneapi_submit_barrier();` will be removed from this PR, as you said it must be tested.
The current approach is tested with the same tests we did on mainline alpaka https://github.com/alpaka-group/alpaka3/blob/dev/tests/unit/event/event.cpp
But in mainline alpaka these tests are not executed for SYCL.
@mehmetyusufoglu Please limit the changes to the required changes only, feature improvements should be kept independent, especially SYCL must be tested very well.

psychocoderHPC:
> If you mean something like calling `cudaDeviceReset`, that should _never_ be called by a library.
> 
> > To decide (decided already):
> > ```
> > * Should a blocking queue also block async operations like `allocAsync`? (assumed Yes in PR) -> NO! CHANGED
> > ```
> 
> Then it was decided wrong. _Of course_ a blocking queue has to block async operations like `allocAsync`.

IMO the name of the function should be changed. In principle, it is a queue-dependent allocation.

Something we should discuss at the next developer meeting, but it is a quick change to change the behaviour.

psychocoderHPC:
> If you mean something like calling `cudaDeviceReset`, that should _never_ be called by a library.

Why should we not call device reset if the last handle is leaving the scope in an alpaka only application?
I understand if you would like to combine native code or different libraries, you should not call it.
Currently, in alpaka3 the last handle is calling device reset if the device is managed by alpaka.
If you use the device managed from outside (I call it ::linkDevice() at the moment/ will be introduced asap in a separate PR) for sure, device reset is not called because alpaka is not allocating the device.

include/alpaka/api/syclGeneric/Queue.hpp
mehmetyusufoglu: Yes i am doing. (But i need to design a test somehow to test it. It is a little challenging.)

include/alpaka/api/syclGeneric/Queue.hpp
psychocoderHPC: > What SYCL back-ends do you plan to support ?

Supported SYCL backends are CPU, IntelGPU, AmdGPU and NvidiaGPU, all via OneApi (at the moment) only.
The code is separated to simply support maybe adaptiveCpp in the future too.

Fpga is not support du to the removed support in OneApi.

The code is from time to time tested by hand on all these GPUs, and in the CI the CPU backend is tested.
The tests are not validating anything because, similar to mainline alpaka we do not have a good example that is using multiple queues besides the event test and the blocking queue test within this PR.

include/alpaka/api/syclGeneric/Queue.hpp
mehmetyusufoglu: Ok, done i guess.

fwyzard:
> > If you mean something like calling `cudaDeviceReset`, that should _never_ be called by a library.
> 
> Why should we not call device reset if the last handle is leaving the scope in an alpaka only application?

The question is the opposite: why _would_ you call device reset ?

> Currently, in alpaka3 the last handle is calling device reset if the device is managed by alpaka.

OK, so in alpaka3 you changed how platforms and devices behave, and they are now fully managed by alpaka ?

fwyzard:
> > If you mean something like calling `cudaDeviceReset`, that should _never_ be called by a library.
> > > To decide (decided already):
> > > ```
> > > * Should a blocking queue also block async operations like `allocAsync`? (assumed Yes in PR) -> NO! CHANGED
> > > ```
> > 
> > 
> > Then it was decided wrong. _Of course_ a blocking queue has to block async operations like `allocAsync`.
> 
> IMO the name of the function should be changed. In principle, it is a queue-dependent allocation.
> 
> Something we should discuss at the next developer meeting, but it is a quick change to change the behaviour.

I disagree: on a CPU backend I want the queue to be blocking (because it is more efficient) and I still want to have allocations that follow queue-ordered semantic (which is what async allocation implements in alpaka2 and in CUDA, for example).

If you make it dependent on the function name instead of the queue type, we have to duplicate all code for cpu vs non-cpu memory operations.

include/alpaka/api/syclGeneric/Queue.hpp
mehmetyusufoglu: Ok, i removed the second tier (`SYCL_EXT_ONEAPI_IN_ORDER_QUEUE_EVENTS`)  because if the first is suported second is supported as well (although second needs ordered queues only). Fallback supports all implementations. I could not find similar solutions like Tier-1 (SYCL_EXT_ONEAPI_ENQUEUE_BARRIER) for AdaptiveCpp and ComputeCpp.

mehmetyusufoglu:
> @mehmetyusufoglu If all works please give me a sign. I will touch few parts I would do different after your changes are in. This is simpler than X iterations.

I finished changes. 

> @mehmetyusufoglu If all works please give me a sign. I will touch few parts I would do different after your changes are in. This is simpler than X iterations.

I have finished my changes. You can change anything you want.

psychocoderHPC:
> OK, so in alpaka3 you changed how platforms and devices behave, and they are now fully managed by alpaka ?

Yes, they are managed by alpaka and will cleanly deleted if the last handle runs out of lifetime. Objects like Events, Streams, ... created by devices or queues always reference to the parent and create a dependency tree of instances.

psychocoderHPC:
@mehmetyusufoglu Thanks, I will take over tomorrow or next week.

fwyzard:
> Yes, they are managed by alpaka and will cleanly deleted if the last handle runs out of lifetime. Objects like Events, Streams, ... created by devices or queues always reference to the parent and create a dependency tree of instances.

I see.

I do agree with this approach, but I'm still not convinced that it's a good idea to call `cudaDeviceReset()`.

What happens if one program is using both the native CUDA back-end the SYCL CUDA back-end ?

How is the interaction with the NVIDIA tools like `nsys` and `compute-sanitizer` ?

psychocoderHPC:
> What happens if one program is using both the native CUDA back-end the SYCL CUDA back-end ?

The user should decide who is handling the device. If the device is externally handled it should be explained to alpaka. In this case via linkDevice to create the device handle.

> How is the interaction with the NVIDIA tools like `nsys` and `compute-sanitizer` ?

Nvidia tools handle this correctly. I think also in the past it was required in PIConGPU to add device reset thatvthe tools work correctly because we stored the device handle in a singleton which was destried after the main function finished.

tests/unit/event/event.cpp
```diff
@@ -666,3 +666,209 @@ TEMPLATE_LIST_TEST_CASE("EventOutOfOrderLifetimeRelease", "", TestApis)
     onHost::wait(q2);
     onHost::wait(q3);
 }
+
+//
+// Test events from specific operations
+//
+
```
mehmetyusufoglu: These tests are added due to comment by @fwyzard  stating 
 ```
On a related topic: last time we checked, not all SYCL events work correctly on all backends.
Please test that events coming from
    empty kernels
    memory set/fill
    memory copies
    the barrier extension
properly wait and synchronise on Intel CPUs, Intel GPUs, NVIDIA GPUs and AMD GPUs, before relying on them in alpaka.
```

psychocoderHPC:
> I disagree: on a CPU backend I want the queue to be blocking (because it is more efficient) and I still want to have allocations that follow queue-ordered semantic (which is what async allocation implements in alpaka2 and in CUDA, for example).

I changed the behaviour: `allocAsync()` is now blocking if it points to a blocking queue.
As I wrote before, I do not like the name`allocAsync` because it follows the queue-order semantics. 
I will rename `allocAsync()` to `allocLazy()` in a follow-up PR.

fwyzard:
> I will rename `allocAsync()` to `allocLazy()` in a follow-up PR.

Please don't.

fwyzard:
First, that's an ugly name. 

Second, it should match the semantic of `hip`/`cudaMallocAsync`, so `allocAsync` is a good name for it.

Third, the _allocation_ is not necessarily lazy: for host-accessible memory the memory should be immediately valid.

psychocoderHPC:
> First, that's an ugly name.
> 
> Second, it should match the semantic of `hip`/`cudaMallocAsync`, so `allocAsync` is a good name for it.

But if it is async, it depends on the queue scheduling schema.  HIP and CUDA have no blocking queues to as soon as a stream is used e.g. for `cudaAllocAsync` it is asynchronously created.

> Third, the _allocation_ is not necessarily lazy: for host-accessible memory the memory should be immediately valid.

Same as in CUDA/HIP the pointer is directly valid but de-referencing is not allowed before the allocation is finished in the queue.

I still do not like `allocAsync()` as name. Let's wait for what others say, maybe someone has a good suggestion.

include/alpaka/api/syclGeneric/Queue.hpp
psychocoderHPC: @mehmetyusufoglu I kept the well-tested current solution for event creation.
Using `m_queue.ext_oneapi_submit_barrier()` makes only sence with the the OneApi sycl queue `sycl::ext::oneapi::property::queue::discard_events` where all operation not return an event enymore. This would require some refactorings and make it more hard to introduce later non OnApi SYCL.
In any cases, this would be an independent feature; therefore, it is removed from this PR.

psychocoderHPC (approved):

PR #237, opened 2025-08-28
CMakeStyle: set line length to 120

psychocoderHPC:
This improves the readability.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,6 +35,6 @@ repos:
   rev: 0.22.1
   hooks:
     - id: gersemi
-      args: [-c, --diff, --no-warn-about-unknown-commands]
+      args: [-c, --diff]
     - id: gersemi
-      args: [-i, --no-warn-about-unknown-commands]
+      args: [-i]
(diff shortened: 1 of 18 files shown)


.pre-commit-config.yaml
```diff
@@ -35,6 +35,6 @@ repos:
   rev: 0.22.1
   hooks:
     - id: gersemi
-      args: [-c, --diff, --no-warn-about-unknown-commands]
+      args: [-c, --diff, --no-warn-about-unknown-commands, --line-length, "120"]
```
SimeonEhrig: Can we please store the config in a configuration file. This approach is more IDE friendly. The default is `.gersemirc` in the root directory. The content would be:

```yaml
# yaml-language-server: $schema=https://raw.githubusercontent.com/BlankSpruce/gersemi/0.22.1/gersemi/configuration.schema.json

definitions: []
disable_formatting: false
extensions: []
indent: 4
line_length: 120
list_expansion: favour-inlining
unsafe: false
warn_about_unknown_commands: false
```

SimeonEhrig (dismissed):

.pre-commit-config.yaml
psychocoderHPC: cool, that's what I searched, I was not finding which file I need

PR #246, opened 2025-09-01
set in all CMakeLists.txt the required version to 3.25

SimeonEhrig:


psychocoderHPC:
note: most files required already 3.25, the files in this PR were overseen in the past.

PR #247, opened 2025-09-01
Boundary Iteration

AntonReinhard:
This adds boundary iteration.

- New functionality for iterating over volume (n-dimensional) boundaries using `BoundaryDirectionsContainer` and others.
- Example for the usage of the boundary iteration
- ~~Adaptation of the heatEquation2D to use boundary iteration~~ [update by @psychocoderHPC ]

There are still some improvements to be made. 
- make an n-dimensional stencil example
- use larger halos/stencils in an example
- make it easier to copy chunks with halos to shared memory (see heat equation stencil kernel)
- make it easier to start parallel kernels for boundary iteration

### DIFF
--- a/include/alpaka/mem/BoundaryIter.hpp
+++ b/include/alpaka/mem/BoundaryIter.hpp
@@ -1,0 +1,398 @@
+/* Copyright 2025 Anton Reinhard
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include "alpaka/CVec.hpp"
+#include "alpaka/Vec.hpp"
+#include "alpaka/api/api.hpp"
+#include "alpaka/api/host/Api.hpp"
+#include "alpaka/concepts.hpp"
+#include "alpaka/core/Assert.hpp"
+#include "alpaka/utility.hpp"
+
+#include <ostream>
+
+namespace alpaka
+{
+
+    /**
+     * @brief An enum representing the different types of boundary, with LOWER, MIDDLE, and UPPER being valid states,
+     * and OOB being invalid (out-of-bounds).
+     */
+    enum class BoundaryType : uint32_t
+    {
+        LOWER,
+        MIDDLE,
+        UPPER,
+        OOB
+    };
+
+    /**
+     * @brief An n-dimensional boundary direction. Encodes a single unique boundary of an nD volume, e.g., a specific
+     * corner of a 2D plane or a side of a 3D cube.
+     *
+     * @tparam Dim The dimensionality of the volume that this is a boundary direction for.
+     * @tparam LowHaloVecType The vector type used for the lower halo sizes.
+     * @tparam UpHaloVecType The vector type used for the upper halo sizes.
+     */
+    template<uint32_t Dim, alpaka::concepts::Vector LowHaloVecType, alpaka::concepts::Vector UpHaloVecType>
+    struct BoundaryDirection
+    {
+        using BoundaryVecType = alpaka::Vec<BoundaryType, Dim>;
+
+        BoundaryVecType data;
+        LowHaloVecType lowerHaloSize;
+        UpHaloVecType upperHaloSize;
+
+        constexpr BoundaryDirection(
+            alpaka::concepts::Vector auto const& boundaries,
+            LowHaloVecType const& lower_halo_sizes,
+            UpHaloVecType const& upper_halo_sizes)
+            : data(boundaries)
+            , lowerHaloSize(lower_halo_sizes)
+            , upperHaloSize(upper_halo_sizes)
+        {
+        }
+
+        [[nodiscard]] static constexpr uint32_t dim()
+        {
+            return Dim;
+        }
+
+        [[nodiscard]] constexpr uint32_t boundaryDimensionality() const
+        {
+            uint32_t c = 0;
+            for(auto i = 0; i < Dim; ++i)
+            {
+                if(data[i] == BoundaryType::MIDDLE)
+                    ++c;
+            }
+            return c;
+        }
+
+        [[nodiscard]] constexpr bool isPoint() const
+        {
+            return boundaryDimensionality() == 0;
+        }
+
+        [[nodiscard]] constexpr bool isLine() const
+        {
+            return boundaryDimensionality() == 1;
+        }
+
+        [[nodiscard]] constexpr bool isPlane() const
+        {
+            return boundaryDimensionality() == 2;
+        }
+
+        [[nodiscard]] constexpr bool isVolume() const
+        {
+            return boundaryDimensionality() == 3;
+        }
+
+        [[nodiscard]] constexpr auto operator<=>(BoundaryDirection const&) const = default;
+    };
+
+    /**
+     * @brief The iterator type for [`BoundaryDirectionsContainer`](@ref)
+     *
+     * @tparam Dim The dimensionality of the volume that this is a boundary direction iterator for.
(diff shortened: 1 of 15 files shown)


include/alpaka/internal.hpp
```diff
@@ -108,5 +108,20 @@ namespace alpaka
         {
             return GetAlignment::Op<std::decay_t<decltype(any)>>{}(any);
         }
+
+        // https://codereview.stackexchange.com/questions/250555/integer-power-c-function-with-base-and-exponent-parameters
+        constexpr int ipow(int const base, int const exponent)
```
psychocoderHPC: should be 

```C++
constexpr int ipow(std::integral auto const base, std::integral  auto const exponent) requires std::same_as<ALPAKA_TYPEOF(base),ALPAKA_TYPEOF(exponent)> 
{
 ///...
 ALPAKA_TYPEOF(base) result = ipow(base, exponent / 2);
```

include/alpaka/mem/BoundaryIter.hpp
```diff
@@ -0,0 +1,395 @@
...
+                    oob = false;
+                    break;
+                case BoundaryType::UPPER:
+                    boundaries.data[i] = BoundaryType::LOWER;
+                    --i;
+                    break;
+                case BoundaryType::OOB:
+                    [[fallthrough]];
+                default:
+                    assert(false);
```
psychocoderHPC: On the device you should use `ALPAKA_ASSERT_ACC()` the reason is that mostly you would like to disable assertions because they introduce stack frames and influence the kernel performance.

The problem is to figure out if we are within a device or on host. The only idea I have is

```
constexpr bool onHost = bstd::is_same_v<api::Host, ALPAKA_TYPEOF(thisApi())>;
```

// true if we are on the host or within a kernel running on host so we can use `assert)

include/alpaka/mem/BoundaryIter.hpp
```diff
@@ -0,0 +1,395 @@
...
+
+    namespace concepts
+    {
+        /** @brief Concept checking whether T is a boundary direction.
+         */
+        template<typename T>
+        concept BoundaryDirection = isBoundaryDirection_v<T>;
+    } // namespace concepts
+
+    ALPAKA_FN_HOST std::ostream& operator<<(std::ostream& os, concepts::BoundaryDirection auto const& bd)
```
psychocoderHPC: `ALPAKA_FN_HOST` is not required.

include/alpaka/mem/BoundaryIter.hpp
```diff
@@ -0,0 +1,395 @@
...
+    }
+
+    /** @brief Construct and return a boundary direction container. This container can be iterated over. See
+     * BoundaryDirectionsContainer.
+     * This constructor uses a default halo size of 1 everywhere.
+     *
+     * @tparam Dims The dimensionality of the container.
+     */
+    template<uint32_t Dims>
+    [[nodiscard]] ALPAKA_FN_HOST_ACC constexpr auto makeBoundaryDirIterator()
```
psychocoderHPC: Please do not use `ALPAKA_FN_HOST_ACC` if the compiler is not requiring it. 
The keyword `constexpr` is creating `__host__ __device__` functions too.

psychocoderHPC (changes requested):

include/alpaka/mem/BoundaryIter.hpp
AntonReinhard: Could this be incorporated in a `ALPAKA_ASSERT_HOST_ACC` then? For situations like this?

psychocoderHPC: good question

example/heatEquation2D/src/heatEquation2D.cpp
```diff
@@ -69,11 +73,11 @@ int example(auto const deviceSpec, auto const computeExec)
...
     // {Y, X}
     constexpr IdxVec numNodes{64, 64};
-    constexpr IdxVec haloSize{2, 2};
-    constexpr IdxVec extent = numNodes + haloSize;
+    constexpr IdxVec haloSize{1, 1};
+    constexpr IdxVec extent = numNodes + haloSize + haloSize;
 
     constexpr uint32_t numTimeSteps = 4000;
-    constexpr double tMax = 0.1;
+    constexpr double tMax = 0.01;
```
psychocoderHPC: not required to be changed

include/alpaka/mem/IdxRange.hpp
```diff
@@ -101,6 +102,38 @@ namespace alpaka
...
+                m_end[i] = range.m_end[i];
+                break;
+            case BoundaryType::MIDDLE:
+                m_begin[i] = range.m_begin[i] + boundaryDir.lowerHaloSize[i];
+                m_end[i] = range.m_end[i] - boundaryDir.upperHaloSize[i];
+                break;
+            case BoundaryType::OOB:
+                [[fallthrough]];
+            default:
+                assert(false);
```
psychocoderHPC: please check if you need to handle on device assert usage because this type can be used onHost and onAcc

include/alpaka/mem/View.hpp
```diff
@@ -142,7 +138,38 @@ namespace alpaka
...
+                case BoundaryType::UPPER:
+                    offset[i] = this->getExtents()[i] - boundaryDir.upperHaloSize[i];
+                    extents[i] = boundaryDir.upperHaloSize[i];
+                    break;
+                case BoundaryType::MIDDLE:
+                    offset[i] = boundaryDir.lowerHaloSize[i];
+                    extents[i] = this->getExtents()[i] - boundaryDir.lowerHaloSize[i] - boundaryDir.upperHaloSize[i];
+                    break;
+                default:
+                    throw std::invalid_argument("invalid direction");
```
psychocoderHPC, line 207: This is not working on GPU, please use device assert

psychocoderHPC (changes requested):

include/alpaka/mem/BoundaryIter.hpp
AntonReinhard: It turns out we can just use `ALPAKA_ASSERT_ACC` on the host too already.

include/alpaka/mem/BoundaryIter.hpp
```diff
@@ -0,0 +1,398 @@
...
+        }
+
+        [[nodiscard]] constexpr bool isPlane() const
+        {
+            return boundaryDimensionality() == 2;
+        }
+
+        [[nodiscard]] constexpr bool isVolume() const
+        {
+            return boundaryDimensionality() == 3;
```
psychocoderHPC, line 107: What is of `boundaryDimensionality()` returns a value >= 4?

Maybe I do not get the definition of point/line/plane/volume.

psychocoderHPC (changes requested):
only a last single question

include/alpaka/mem/BoundaryIter.hpp
AntonReinhard: The `isPoint`, `isLine` and `isVolume` are just syntactic sugar for asking if it's 1/2/3D. If the dimensionality is higher then it's simply none of those and the user should use `boundaryDimensionality()` directly.

AntonReinhard: I added some docstrings with explanation and renamed the functions

psychocoderHPC (approved):

psychocoderHPC:
I rebased the PR and fixed few warnings

PR #250, opened 2025-09-02
View/MdSpan preserve constness

psychocoderHPC:
This PR changes the behaviour of MdPSpan and View. The constness of the view/span is forwarded to the data too.

- [x] add tests to ensure correct behaviour

```C++
auto foo = onHost::allocHost<int>(10);
// mutable view
auto fooSub = foo.getSubView(2); 
static_assert(!std::is_constant_v<std::remove_pointer_t<ALPAKA_TYPEOF(fooSub.data())>>);

auto const fooOuterConst = foo;
// outer const of fooOuterConst is propagated into inner const of the subview
auto fooSubInnerConst = fooOuterConst.getSubView(2); 
static_assert(std::is_constant_v<std::remove_pointer_t<ALPAKA_TYPEOF(fooOuterConst.data())>>);
```

### DIFF
--- a/include/alpaka/mem/View.hpp
+++ b/include/alpaka/mem/View.hpp
@@ -23,7 +23,7 @@ namespace alpaka
     /** Non owning view to data
      *
      * This view is only holding a points to real data, copying the view is cheap.
-     * Const-ness of the view is NOT propagated to the data region.
+     * Const-ness of the view instance is propagated to the data region.
      */
     template<
         typename T_Api,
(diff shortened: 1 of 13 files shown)


include/alpaka/mem/View.hpp
```diff
@@ -142,7 +165,16 @@ namespace alpaka
...
-            return View{T_Api{}, shiftedPtr, extents, this->getPitches(), Alignment<>{}};
+            return makeView(T_Api{}, shiftedPtr, extents, this->getPitches(), Alignment<>{});
+        }
+
+        constexpr auto getSubView(
+            alpaka::concepts::Vector auto const& offset,
+            alpaka::concepts::Vector auto const& extents)
+        {
+            assert((extents <= this->getExtents()).reduce(std::logical_and{}));
+            auto shiftedPtr = &getMdSpan()[offset];
```
psychocoderHPC: should be `auto shiftedPtr = &(*this)[offset];` to avoid creation of an temporary type.

tests/unit/mem/constCorrectness.cpp
```diff
@@ -0,0 +1,208 @@
...
+        static_assert(!std::is_const_v<std::remove_pointer_t<decltype(view.data())>>);
+        static_assert(!std::is_const_v<std::remove_reference_t<decltype(view[Vec{0, 0}])>>);
+
+        [[maybe_unused]] MdSpan mdSpan = buffer0.getMdSpan();
+        static_assert(!std::is_const_v<std::remove_pointer_t<decltype(mdSpan.data())>>);
+        static_assert(!std::is_const_v<std::remove_reference_t<decltype(mdSpan[Vec{0, 0}])>>);
+    }
+    // non-mutable views
+    {
+        onHost::ManagedView innerConstBuffer0 = buffer0.getConstManagedView();
```
SimeonEhrig, line 153: Just a question. Because we cannot initialize the values at construction time, it is not possible to create a `onHost::ManageView` with const data type, e.g. `onHost::ManageView<int const>`?

SimeonEhrig (approved):

tests/unit/mem/constCorrectness.cpp
psychocoderHPC: It is possible but it would not make sense

```C++
auto foo = ManagedView<api::Host, int const, Vec<int,1u>,Alignment<>>(
  api::host,
  ptr,
  Vec<int,1u>{10},
  Vec<int,1u>{sizeof(int)}, 
  [](){}, 
  Alignment<>{}
);
```

should create a inner const managed view. (not tested if I have an syntax error in my example).
But all in all it makes no sence because you can not fill it with data.

psychocoderHPC: Where it would make sense is if the pointer you pass to the constructor is already initialized with data

PR #251, opened 2025-09-03
rename `ManagedView` to `SharedBuffer`

psychocoderHPC:
We decided in the developer meeting to rename the managed view to `SharedBuffer`

### DIFF
--- a/docs/snippets/10_memory.cpp
+++ b/docs/snippets/10_memory.cpp
@@ -35,25 +35,25 @@ TEST_CASE("memory", "[docs]")
 
     // Allocate a memory view on the compute device.
     // The memory will be freed automatically when the view goes out of scope.
-    onHost::SharedBuffer computeView = onHost::alloc<int>(computeDev, 10);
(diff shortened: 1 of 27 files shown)


docs/snippets/10_memory.cpp
```diff
@@ -35,9 +35,9 @@ TEST_CASE("memory", "[docs]")
 
     // Allocate a memory view on the compute device.
     // The memory will be freed automatically when the view goes out of scope.
-    onHost::ManagedView computeView = onHost::alloc<int>(computeDev, 10);
+    onHost::SharedBuffer computeView = onHost::alloc<int>(computeDev, 10);
```
SimeonEhrig: Please rename also the variables. If we introduce `alpaka::onHost::View` this can confuse.
This happens at some points in the code.

psychocoderHPC: good point these I missed. Applied a new commit with renaming of instances.

PR #255, opened 2025-09-05
Add cmake options to enable sanitizers for alpaka targets

AntonReinhard:
I added a selection of sanitizers that should work for both gcc and clang:
- ASAN -> address sanitizer for problems like double free, invalid reads, etc.
- TSAN -> thread sanitizer to detect data races
- LSAN -> leak sanitizer to detect memory leaks
- UBSAN -> undefined behavior sanitizer to detect undefined behavior

Maybe it would make sense to activate at least some of these for tests (especially ubsan). Unfortunately, only one can be enabled at a time.


cmake/finalize.cmake
```diff
@@ -207,4 +207,43 @@ function(alpaka_finalize target)
...
+                message(
+                    WARNING
+                    "Multiple sanitizers are enabled, but only one can be linked at a time. Using TSAN only."
+                )
+            endif()
+        elseif(alpaka_LSAN)
+            message(STATUS "Linking LSAN to ${target}")
+            target_compile_options(${target} PRIVATE -fsanitize=leak)
+            target_link_options(${target} PRIVATE -fsanitize=leak)
+            if(alpaka_UBSAN)
```
mehmetyusufoglu, line 237: [maybe] can these warning be at the beginning of elseif block?  First says "Linking LSAN" then says "Multiple ... Using LSAN only."

AntonReinhard: I think it makes sense the way it is, because it first tells you it's linking LSAN, which it is, and then warns you that others are also enabled but will not be linked.

.github/workflows/ci.yml
```diff
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  alpaka_cmake_flags: "-Dalpaka_DOCS=ON -Dalpaka_TESTING=ON -Dalpaka_BENCHMARKS=ON -Dalpaka_EXAMPLES=ON -DBUILD_TESTING=ON -Dalpaka_HEADERCHECKS=ON"
+  alpaka_cmake_flags: "-Dalpaka_DOCS=ON -Dalpaka_TESTING=ON -Dalpaka_BENCHMARKS=ON -Dalpaka_EXAMPLES=ON -DBUILD_TESTING=ON -Dalpaka_HEADERCHECKS=ON -Dalpaka_UBSAN=ON"
```
SimeonEhrig: I think the sanitizer jobs should be extra jobs, because normally the take much more time to compile and run the tests.

AntonReinhard: What would you suggest to do? Trigger them manually somehow? Or just run both with and without them?

psychocoderHPC: I agree with @SimeonEhrig we should have extra jobs for the sanitizer, best explicit for the latest compiler only.
Feel free to add a new column to 

https://github.com/alpaka-group/alpaka3/blob/c7721727411dc0d49db750f6c3d2b89c24e9e519/.github/workflows/ci.yml#L34-L52

And repeat the line for the newest compiler and activate conditionally the sanitizer.

SimeonEhrig: Please create extra jobs for this and then we see, what is the compile time overhead and how the output looks like. In alpaka mainline, I want to create an extra pipeline for jobs like this. Also to collect more experience with sanitizers in the CI. At the moment, we run asan in the normal CI pipeline.

example/tutorial/src/03_memory.cpp
AntonReinhard, line 1: @psychocoderHPC There was a use-after-free here and in tutorial 04, I'm not sure if this is what was intended

AntonReinhard:
It looks like the event test is using some active waiting/spin locks, that tsan of course doesn't recognize and therefore prints warnings about. I'm not sure how to fix this, I guess we'd have to disable TSAN for those specific tests?

AntonReinhard:
I don't really understand what's going on with the paths inside the container. Somehow it keeps not finding the suppressions files, regardless of which documented github variable I use...

psychocoderHPC:
> I don't really understand what's going on with the paths inside the container. Somehow it keeps not finding the suppressions files, regardless of which documented github variable I use...

You can use pop and push within the code to disable option for specific code secions. could you please check alpaka mainline for push pop to see how it works

psychocoderHPC:
> You can use pop and push within the code to disable option for specific code secions. could you please check alpaka mainline for push pop to see how it works

This is only something for code within alpaka directly. For the test single test we should disable the sanitizer.

psychocoderHPC (approved):

PR #258, opened 2025-09-09
Add memoryFence, the "memory ordering API", with support for different scopes

mehmetyusufoglu:
- [x] rebase against #271

Full barrier thread fence implementation of atomic operations. (like in alpaka,  provides full sequential consistency - it acts like  `std::memory_order_seq_cst` . Seemingly Cuda **only supports full barriers**)

`Full barrier:` It's both acquire AND release ( **Acquire:** no atomic reads/writes in the **same thread** can be reordered before this fence point in the code. "Before" here means the operations before the thread-fence line. They cannot jump the fence while being re-ordered by compiler. **Release:** no reads/writes in the **same thread** can be reordered after this fence point (after in the code) before the fence. The code after the thread-fence code line, can not jump back when reordered by compiler.)

#PR Details

• Added memory ordering primitives to alpaka3 via threadFence operations ensuring proper memory visibility.
• Implemented three memory scopes: `Block` (thread block), `Device` (entire GPU), and `System` (cross-device/host)
• Backend-specific implementations: Host uses `std::atomic_thread_fence`, CUDA/HIP use `__threadfence `intrinsics, SYCL uses `sycl::atomic_fence`
• Non-blocking synchronization: Provides memory ordering without thread barriers, enabling efficient producer-consumer patterns in parallel kernels

#Test Files
threadFenceBasic.cpp: Validates compilation and execution of all threadFence API variants across enabled backends.

threadFenceProducerConsumer.cpp: Tests memory ordering correctness with producer-consumer synchronization and shared memory patterns.

- [x] Squash commits


include/alpaka/api/host/threadFence.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+namespace alpaka::onAcc::internalCompute
+{
+    // All scopes collapse to a single host atomic fence
+    template<typename T_Scope>
+    struct ThreadFence::Op<T_Scope, internal::HostFence>
+    {
+        ALPAKA_FN_ACC static void fence(T_Scope const&, internal::HostFence const&) noexcept
+        {
+#if !defined(__CUDA_ARCH__) && !defined(__HIP_DEVICE_COMPILE__)
```
psychocoderHPC: #if !ALPAKA_ARCH_PTX && !ALPAKA_ARCH_AMD

include/alpaka/api/host/threadFence.hpp
```diff
@@ -0,0 +1,49 @@
...
+    };
+} // namespace alpaka::onAcc::trait
+
+namespace alpaka::onAcc::internalCompute
+{
+    // All scopes collapse to a single host atomic fence
+    template<typename T_Scope>
+    struct ThreadFence::Op<T_Scope, internal::HostFence>
+    {
+        ALPAKA_FN_ACC static void fence(T_Scope const&, internal::HostFence const&) noexcept
```
psychocoderHPC: you can use `constexpr` instead of `ALPAKA_FN_ACC`

include/alpaka/api/host/threadFence.hpp
psychocoderHPC: in general I am not sure if this macro guard is required at all because this code should not be called from non-host accelerated code anyway.
Please try if you can remove the macro guard at all.

include/alpaka/api/host/threadFence.hpp
psychocoderHPC: plaese `T_Scope const&` to `T_Scope const` because the scope is e.g. 

```
constexpr auto scopeFoo = internal::HostFence{};
```

There will be issues with nvcc. The scope has no data inside; accepting it by copy does not affect the performance.

include/alpaka/api/host/threadFence.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include "alpaka/core/common.hpp"
+#include "alpaka/onAcc/threadFence.hpp"
+
+#include <atomic>
+
+namespace alpaka::onAcc::trait
+{
+    // Map host executors to HostFence impl tag (mirrors atomic pattern)
+    template<>
+    struct GetThreadFenceImpl::Op<alpaka::exec::CpuSerial>
```
psychocoderHPC: Please have a look here and use similar style to implement methods.

https://github.com/alpaka-group/alpaka3/blob/c7721727411dc0d49db750f6c3d2b89c24e9e519/include/alpaka/onAcc/internal/interface.hpp#L17-L29

Please add the internal interface to `alpaka::onAcc::internalCompute` and not trait like in mainline alpaka.
Best add the method to the [alpaka::onAcc](https://github.com/alpaka-group/alpaka3/blob/dev/include/alpaka/onAcc/Acc.hpp) as free function similar to the others.

include/alpaka/api/unifiedCudaHip/threadFence.hpp
```diff
@@ -0,0 +1,108 @@
...
+} // namespace alpaka::onAcc::trait
+
+namespace alpaka::onAcc::internalCompute
+{
+    // CUDA implementations (device side only). Provide host-side no-ops to satisfy host compilation of __host__
+    // __device__ kernels.
+    template<>
+    struct ThreadFence::Op<memoryScope::Block, internal::CudaFence>
+    {
+        ALPAKA_FN_ACC static void fence(memoryScope::Block const&, internal::CudaFence const&) noexcept
```
psychocoderHPC: please use `constexpr` instead of `ALPAKA_FN_ACC`

include/alpaka/api/unifiedCudaHip/threadFence.hpp
```diff
@@ -0,0 +1,108 @@
...
+        {
+#ifdef __CUDA_ARCH__
```
psychocoderHPC: This should not be required if the full file is guared by https://github.com/alpaka-group/alpaka3/blob/c7721727411dc0d49db750f6c3d2b89c24e9e519/include/alpaka/api/unifiedCudaHip/Event.hpp#L12

include/alpaka/onAcc/threadFence.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include "alpaka/tag.hpp"
+
+#include <type_traits>
+
+namespace alpaka::onAcc
+{
+    namespace memoryScope
+    {
+        //! Thread fences are observed by all threads in the same block.
+        struct Block
```
psychocoderHPC: Similar to https://github.com/alpaka-group/alpaka3/blob/c7721727411dc0d49db750f6c3d2b89c24e9e519/include/alpaka/api/host/executor.hpp#L16-L27 and othe rplaces please always create a constexpr variable with the same name.

e.g. `constexpr block = Block{}`

in the user interface the instance is always used. alpaka3 tries to avoid that the user is required to use `foo<Bar>()` instead the interfaces for the user are `foo(bar)`

tests/unit/mem/threadFenceBasic.cpp
```diff
@@ -0,0 +1,80 @@
...
+    ALPAKA_FN_ACC void operator()(auto const& acc, auto out) const
+    {
+        using namespace alpaka::onAcc;
+        // Iterate over all logical thread indices in the launched frame.
+        for(auto idx : onAcc::makeIdxMap(acc, onAcc::worker::threadsInGrid, onAcc::range::totalFrameSpecExtent))
+        {
+            // Initial write; any subsequent fence must not block, just enforce visibility / ordering.
+            out[idx.x()] = static_cast<uint32_t>(idx.x());
+            // Explicit scope form.
+            threadFence(acc, memoryScope::Block{});
```
psychocoderHPC: as above descriped this will be 

```
threadFence(acc, memoryScope::block);
```

psychocoderHPC (changes requested):

include/alpaka/onAcc/threadFence.hpp
mehmetyusufoglu: But exec::cpuserial is host  variable. How can i make it acc variable at the same time?

include/alpaka/api/unifiedCudaHip/threadFence.hpp
mehmetyusufoglu: But event.hpp functions are host functions? we  have call acc from host etc errors?

include/alpaka/api/host/threadFence.hpp
mehmetyusufoglu: removed

include/alpaka/api/host/threadFence.hpp
mehmetyusufoglu: done

mehmetyusufoglu: in all 4 `threadFence.hpp` files it is changed

include/alpaka/api/host/threadFence.hpp
mehmetyusufoglu: checking...

include/alpaka/api/host/threadFence.hpp
mehmetyusufoglu: https://github.com/alpaka-group/alpaka3/actions/runs/17715339932/job/50339472242?pr=258

Clang compilation failed, i am investigating:

```

In file included from /__w/alpaka3/alpaka3/build/tests/unit/block/hip/block.cpp:6:
In file included from /__w/alpaka3/alpaka3/include/alpaka/alpaka.hpp:53:
/__w/alpaka3/alpaka3/include/alpaka/api/unifiedCudaHip/threadFence.hpp:77:31: error: constexpr function never produces a constant expression [-Winvalid-constexpr]
   77 |         static constexpr void fence(memoryScope::Block const&, internal::HipFence const&) noexcept
      |                               ^~~~~
/__w/alpaka3/alpaka3/include/alpaka/api/unifiedCudaHip/threadFence.hpp:80:13: note: non-constexpr function '__threadfence_block' cannot be used in a constant expression
   80 |             __threadfence_block();
      |             ^
/opt/rocm-6.2.4/lib/llvm/bin/../../../include/hip/amd_detail/amd_device_functions.h:750:13: note: declared here
  750 | static void __threadfence_block()
      |             ^
In file included from /__w/alpaka3/alpaka3/build/tests/unit/block/hip/block.cpp:6:
In file included from /__w/alpaka3/alpaka3/include/alpaka/alpaka.hpp:53:
/__w/alpaka3/alpaka3/include/alpaka/api/unifiedCudaHip/threadFence.hpp:88:31: error: constexpr function never produces a constant expression [-Winvalid-constexpr]
   88 |         static constexpr void fence(memoryScope::Device const&, internal::HipFence const&) noexcept
      |                               ^~~~~
/__w/alpaka3/alpaka3/include/alpaka/api/unifiedCudaHip/threadFence.hpp:91:13: note: non-constexpr function '__threadfence' cannot be used in a constant expression
   91 |             __threadfence();
```

include/alpaka/api/host/threadFence.hpp
mehmetyusufoglu: Ok i understood.
- Entry point
`alpaka::onAcc  threadFence(acc, scope) is forwarder  `
- dispatcher to be specialised 
`internalCompute::ThreadFence::Op<T_Acc, T_Scope> `
- Backend specializations:
```
alpaka::api::host (requires HostApiAcc<T_Acc>)
alpaka::api::unifiedCudaHip (requires CudaAcc<T_Acc> || HipAcc<T_Acc>)
alpaka::api::syclGeneric (requires OneApiAcc<T_Acc>)
```

include/alpaka/api/host/threadFence.hpp
mehmetyusufoglu: ok solved.

include/alpaka/api/host/threadFence.hpp
mehmetyusufoglu: Done.

tests/unit/mem/threadFenceBasic.cpp
mehmetyusufoglu: done , thanks

include/alpaka/onAcc/threadFence.hpp
mehmetyusufoglu: done, thanks.

include/alpaka/api/unifiedCudaHip/threadFence.hpp
mehmetyusufoglu: intrinsics guard is needed just before `__threadfence()`. Removing intrinsics's guards `(#    if defined(__CUDA_ARCH__) || defined(__HIP_DEVICE_COMPILE__))` despite `#if ALPAKA_LANG_CUDA || ALPAKA_LANG_HIP `  were at the  top created compile errors. 

Therefore  intrinsics' guards are needed. Removing the top guard  `#if ALPAKA_LANG_CUDA || ALPAKA_LANG_HIP ` did not create any problem, but i kept to prevent specialisation unnecessarily just in case.

psychocoderHPC (approved):

tests/unit/mem/memoryFenceProducerConsumer.cpp
```diff
@@ -0,0 +1,329 @@
...
+        CHECK(flag[0u] == 1u);
+    }
+
+    {
+        // Run a single kernel, testing a memory fence in global memory across threads in different blocks
+        onHost::memcpy(queue, vars_dev, vars_host);
+        onHost::wait(queue);
+        flag[0u] = 1u;
+        // Device-scope variant
+        queue.enqueue(exec, onHost::FrameSpec{2, 1}, KernelBundle{DeviceFenceTestKernel{}, flag, vars_dev.data()});
```
psychocoderHPC, line 307: This must be `ThreadSpec` because the kernel is not giving the guarantee that it runs with any number of threads

tests/unit/mem/memoryFenceProducerConsumer.cpp
```diff
@@ -0,0 +1,329 @@
...
+
+    int const numElements = 2;
+    auto vars_host = onHost::allocMapped<int>(device, numElements);
+    auto vars_dev = onHost::alloc<int>(device, numElements);
+    vars_host[0] = 1;
+    vars_host[1] = 2;
+
+    {
+        flag[0u] = 1u;
+        queue.enqueue(exec, onHost::FrameSpec{1, 2}, KernelBundle{BlockSharedMemOrderKernel{}, flag});
```
psychocoderHPC, line 296: This must be ThreadSpec because the kernel is not guaranteeing that it runs with any number of threads.
This is wrong only for serial executor because we start always only a single threadblock with one thread.

tests/unit/mem/memoryFenceProducerConsumer.cpp
psychocoderHPC: fixed in #276

tests/unit/mem/memoryFenceProducerConsumer.cpp
psychocoderHPC: fixed in #276

PR #262, opened 2025-09-12
Boundary iterator test fix

AntonReinhard:
I'm not sure why the problem wasn't caught in CI runs, but it should be fixed now, I tested it on a setup that previously reproduced the segfault.

- Fix host usage of device buffer
- Add small example kernel to run on devices

### DIFF
--- a/example/boundaryIter/src/boundaryIter.cpp
+++ b/example/boundaryIter/src/boundaryIter.cpp
@@ -2,6 +2,8 @@
  * SPDX-License-Identifier: MPL-2.0
  */
 
+#include "alpaka/onHost/example/executors.hpp"
+
 #include <alpaka/alpaka.hpp>
 #include <alpaka/onHost/executeForEach.hpp>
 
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
The ci can only run cpu runtime tests and gpu compile tests because we use fithub actions. 
As soon we switch to the HZDR CI we can cover gpu runs too.

AntonReinhard:
> The ci can only run cpu runtime tests and gpu compile tests because we use fithub actions. As soon we switch to the HZDR CI we can cover gpu runs too.

Oh okay, that makes sense, I wasn't aware of that

example/boundaryIter/src/boundaryIter.cpp
```diff
@@ -48,47 +67,86 @@ int example(auto const devSpec)
...
-    auto hostDev = onHost::makeHostDevice();
-    auto deviceQueue = hostDev.makeQueue();
-    onHost::iota(deviceQueue, 1.f, buffer);
-    onHost::wait(deviceQueue);
+    auto bufferTarget = onHost::allocLike(device, buffer);
+    auto viewTarget = buffer.getView();
 
+    auto blockingQueue = device.makeQueue(queueKind::blocking);
+    onHost::iota(blockingQueue, 1.f, buffer);
+    onHost::wait(blockingQueue);
```
psychocoderHPC: wait is not required because the queue is blocking and therefore iota will be finished after the controll flow is coming back from the call.

AntonReinhard: Right, that was leftover. Removed in ef6de2c757e3d409f007cbc3902cf7b92184ec6f

psychocoderHPC (approved):

PR #267, opened 2025-09-17
add parent directory name of unit test source files to the cmake target name 

SimeonEhrig:
Allows to use the same source code file name in different folders. Before if the name was not unique, unit tests could hide each other and was not compiled and executed.


SimeonEhrig:
I write a doc section how to handle code changes, caused by a tool, e.g. running a formatter. Afterwards I change back to normal PR.

psychocoderHPC (approved):

PR #268, opened 2025-09-17
Add a meaningful error message for invalid kernel operators

AntonReinhard:
This is now catching two types of invalid operators:
- The operator is defined but not marked const, giving an error message like `alpaka3/include/alpaka/KernelBundle.hpp:111:22: error: static assertion failed: the operator() function of a kernel must be marked const`
- The operator is defined but returns something other than void, giving an error message like `alpaka3/include/alpaka/KernelBundle.hpp:117:22: error: static assertion failed: the return type of the operator() function of a kernel must be void`

It does not catch errors when the function is not callable because the arguments are wrong or similar. Maybe this case could also be caught with a nicer error message.

I also made the demangledName functions constexpr, they could be used to make the static_assert string print out the invalid kernel's name, but currently that's a c++26 feature or compiler extension, which throws a warning, so I didn't do that yet.

### DIFF
--- a/include/alpaka/KernelBundle.hpp
+++ b/include/alpaka/KernelBundle.hpp
@@ -11,6 +11,7 @@
 #include "alpaka/core/config.hpp"
 #include "alpaka/trait.hpp"
 #include "alpaka/utility.hpp"
+#include "onHost/demangledName.hpp"
 
 #include <alpaka/mem/concepts.hpp>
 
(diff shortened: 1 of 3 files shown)


SimeonEhrig (approved):

include/alpaka/KernelBundle.hpp
```diff
@@ -100,14 +101,26 @@ namespace alpaka
...
-                    remove_restrict_t<ALPAKA_TYPEOF(onHost::makeAccessibleOnAcc(std::declval<TArgs>()))>...>>)
+            && std::is_invocable_v<
+                std::remove_const_t<KernelFn>,
+                TAcc,
+                remove_restrict_t<ALPAKA_TYPEOF(onHost::makeAccessibleOnAcc(std::declval<TArgs>()))>...>)
         constexpr auto operator()(TAcc const& acc) const
         {
+            static_assert(
+                std::is_invocable_v<
+                    std::add_const_t<KernelFn>,
```
SimeonEhrig, line 112: I saw this, after I approved :-( How is this working? It looks like you make the function const.

AntonReinhard: The idea is to add const to the invocable check to check whether the operator is callable when the object is const, since it must be. It's not actually muting anything.

SimeonEhrig: That's assume that you cannot mark a mutable function as const. I don't know if this is the case but I tested it and it is working. Therefore it is fine :-)

AntonReinhard: I'm not sure what you mean, `KernelFn` is an object implementing `operator()`, or at least it should be. If that `operator()` is not marked const, it can't be called later on, and to give a better error message, this is now checked here. If it is marked const, the operator is callable on the constant `KernelFn` object and everything is fine.

PR #270, opened 2025-09-18
Implement interface concepts for the host- and device-side memory

SimeonEhrig:
The interface concepts describes which member function, types and variables are available for the given memory object.

### DIFF
--- a/include/alpaka/mem/concepts/ExpectedValueType.hpp
+++ b/include/alpaka/mem/concepts/ExpectedValueType.hpp
@@ -1,0 +1,18 @@
+/* Copyright 2025 Simeon Ehrig
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include "alpaka/trait.hpp"
+
+#include <concepts>
+
+namespace alpaka::concepts
+{
+    /** Check whether the specified data type matches the expected type, or if the expected type is
+     *`alpaka::NotRequired`, then all data types passes the concept.
+     **/
+    template<typename TData, typename TExpected>
+    concept ExpectedValueType = std::same_as<TExpected, TData> || std::same_as<TExpected, alpaka::NotRequired>;
+} // namespace alpaka::concepts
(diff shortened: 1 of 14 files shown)


include/alpaka/mem/concepts/IBuffer.hpp
```diff
@@ -0,0 +1,69 @@
...
+         *
+         * - <b>t.addDestructorAction</b>: Adds a destructor action to the shared buffer.
+         * @code{.unparsed}
+         *    The action will be executed when the buffer is destroyed.
+         *    This can be used to add additional cleanup actions e.g. waiting on a specific queue.
+         *    Actions are executed in FIFO order.
+         * @endcode
+         * - <b>t.destructorWaitFor</b>: Add an action to be executed when the shared_ptr is destroyed.
+         **/
+        template<typename T, typename MutT, typename ConstT>
```
psychocoderHPC, line 44: This should be `T_Mut` and `T_Const`

include/alpaka/mem/concepts/ExpectedValueType.hpp
```diff
@@ -0,0 +1,18 @@
...
+#include "alpaka/trait.hpp"
+
+#include <concepts>
+
+namespace alpaka::concepts
+{
+    /** Check whether the specified data type matches the expected type, or if the expected type is
+     *`alpaka::NotRequired`, then all data types passes the concept.
+     **/
+    template<typename TData, typename TExpected>
```
psychocoderHPC, line 16: should be `T_Data` and `T_Expeced`

psychocoderHPC (approved):
alpaka3 switched to the naming schema for template arguments we discussed in the past in https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#16-template-declarations-and-specializations

It is not a big issue, we can change it later too. There are other parts which are copied from the mainline alpaka and breaking the rules too.

I found some `\DOXYCOMMAND` instead of `@DOXYCOMMAND` but I will add at some point a style checker to find all of them.

I will merge it as it is but would like to point to it.

PR #282, opened 2025-09-19
[No-merge] hip host2host memset memcopy

mehmetyusufoglu:


psychocoderHPC:
thanks for providing the example, I decided to use shorter code which scans over multiple dimensions

PR #295, opened 2025-09-29
fix some format issues in Sphinx Doc documentation

SimeonEhrig:


docs/source/dev/logging.rst
```diff
@@ -27,17 +27,20 @@ CMake Logging Options
 ---------------------
 
 * ``alpaka_LOG=<X>`` where ``X`` can be
-   - ``OFF``,``off`` - to disable logging, other CMake option ``alpaka_LOG_*`` will not be available.
+
+   - ``OFF`` - to disable logging, other CMake option ``alpaka_LOG_*`` will not be available.
```
psychocoderHPC, line 31: `off` was correct too, I thought small letters is supported too.
I will remove the lowercase version of this option in a separate PR.

psychocoderHPC (approved):

docs/source/dev/logging.rst
psychocoderHPC: #296 remove support for `off`

SimeonEhrig: I was only checking the CMake `option()` which describes the `alpaka_LOG` option. Therefore  I was not aware of this possibility.

PR #306, opened 2025-10-08
make folders in docs snippets for organization

ikbuibui:
split examples and cheat sheet and added a top level `CMakeList`


SimeonEhrig (approved):
The cheat sheet is still working in Sphinx Doc and I checked a CI job. It compiles the cheat sheet code.

PR #307, opened 2025-10-08
add support for using invalid api and device kind combination

psychocoderHPC:
Provide the possibility to handle at runtime invalid api and device kind usage.


psychocoderHPC:
Ohh this is not working as soon as we create a queue it will fail, I do not like to create a dummy queue too.

psychocoderHPC:
This is not working as intended, we would need a queue dummy and many more.
Therefore tis PR is closed.

PR #313, opened 2025-10-09
document terms host and accelerator

SimeonEhrig:


docs/source/basic/terms.rst
```diff
@@ -0,0 +1,47 @@
+Terms & Structure
+=================
+
+.. sectionauthor:: Simeon Ehrig
+
+
+Host and Accelerator
+--------------------
+
+In Alpaka, we distinguish between the between ``host``-side code and ``accelerator``-side code. This separation follows the GPU offload model, which is used by all important GPU vendors. This model means, that some parts of the source code are executed on the processor, where the operation system is executed (``host``, CPU) and some code is offloaded and executed on an extra processor, which dedicated to the algorithmic work (``accelerator``, GPU).
```
psychocoderHPC: ... and some code is ...

should be 

... computations are offloaded and executed as kernel  ...

psychocoderHPC: ` terms onHost for` to `terms ``onHost`` for` ot `terms *onHost* for`

psychocoderHPC: `and onAcc for` to `and ``onAcc`` for` or `and *onAcc* for`

psychocoderHPC: ` ``accelerator``-side. ` to ` ``accelerator``-side (within a kernel).`

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,47 @@
...
+
+In the following of documentation, we will use the terms onHost for ``host``-side code and onAcc for ``accelerator``-side code. This reflects also in the alpaka namespaces. ``alpaka::onHost`` means that the functions and objects are available on the ``host``-side. Functions in the namespace ``alpaka::onAcc`` are available on the ``accelerator``-side. If a function or object is in namespace ``alpaka``, it is available on ``host``- and ``accelerator``-side, e.g. ``alpaka::MdSpan``.
+
+Host
+````
+
+The ``host``-side is mainly controlling the application workflow, like selecting the ``accelerator``-:ref:`Device`, allocate memory, start algorithm on the ``accelerator``-side and more.
```
psychocoderHPC: `start algorithm` to `enqueue algorithms`

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,47 @@
...
+
+[1] In theory all C++ functionality can be used but in practice there are limitation by the SDKs used by the Backends. For example the CUDA SDK does not support C++20 modules (valid for CUDA 13.x and before).
+
+Accelerator
+```````````
+
+The ``accelerator``-side code contains the actual algorithm which are executed on the accelerator device, like a CPU and GPU.
+
+Properties:
+    - The ``accelerator``-side can be only entered from the ``host``-side by special alpaka launch functions.
```
psychocoderHPC: `entered from the ``host``-side by special alpaka launch functions.`

to

`entered from the ``host``-side by enqueuing a compute kernel.`

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,47 @@
...
+    - From ``accelerator``-side, it is not possible to access operation system resources such the filesystem.
+    - Not all C++ features are support on ``accelerator``-side, like recursive function calls.
+    - No C++ references can be passed to the ``accelerator```-side, only trivial copyable values.
+    - From the ``accelerator``-side, not every memory is accessible, which is accessible from the ``host``-side.
+
+.. hint::
+
+    Later we will introduce the keywords :ref:`API` and :ref:`Device`. The device CPU which is created with the API host, can be used on ``host`` and ``accelerator`` side. If the CPU is used on the ``accelerator``, it needs to follow all limitations of the ``accelerator``-side to be portable. For example, it is technically possible to access the files system from ``accelerator``, if the device is a CPU. If the same code is execute on a GPU, the application will crash and is not portable anymore.
```
psychocoderHPC: `The device CPU which is created with the API host, can be used on ``host`` and ``accelerator`` side....`

to 

`If a kernel is enqueued to the host CPU, it is required to follow the limitations of ``accelerators`` not to call operating system methods even if it is technically possible, but would break portability. `

psychocoderHPC (changes requested):

psychocoderHPC (approved):

PR #321, opened 2025-10-16
TBB backend implementation

mehmetyusufoglu:
- add executor `cpuTbbBlocks`    
    - implement the CPU executor `cpuTbbBlocks` similar to `cpuOmpBlocks`
    - CI: add TBB tests
    - **sneak in into this PR:**
      - a very basic atomic test case
      - fix the name of the OpenMP executor in the documenation
        
 ~~**Main difference from mainline alpaka**~~
     ~~Here; each alpaka queue holds its own arena. In legacy alpaka(alpaka mainline),~~ 
     ~~tbb::this_task_arena::isolate is used, that means every queue shares the same global pool of threads.~~ 
     ~~The reason is flexibility of separated arenas: https://www.intel.com/content/www/us/en/docs/onetbb/developer-guide-api-reference/2022-0/guiding-task-scheduler-execution.html~~

  ~~Bablstream output including TBB backend:~~
~~[babelstream-out.txt](https://github.com/user-attachments/files/23053423/babelstream-out.txt)~~
~~Babelstream output including TBB from old or legacy alpaka:
[babelstream-out-oldAlpaka.txt](https://github.com/user-attachments/files/23055926/babelstream-out-oldAlpaka.txt)~~

~~**Important: I** am trying to understand the performance difference between mainline-alpaka and alpaka3, and between Omp and TBB backends.~~
[psychocoderHPC] This points are not valid anymore, with the last commit we switched to the implementation used in mainline alpaka which solves all performance issues.

**TODO before merge:**
- [x]  Squash commits


SimeonEhrig:
You need also add TBB to the CI. Otherwise you test only, if your changes does not break the other backends.

include/alpaka/core/config.hpp
```diff
@@ -273,3 +273,8 @@
 #        define ALPAKA_OMP ALPAKA_VERSION_NUMBER_NOT_AVAILABLE
 #    endif
 #endif
+
+// oneTBB (set to zero unless provided via build system)
+#if !defined(ALPAKA_TBB)
+#    define ALPAKA_TBB 0
```
psychocoderHPC: please use the native tbb macros provided by tbb for the detection. In case no tbb is available use the alpaka_verion_notavailable macro. please use the OMP detection as template

psychocoderHPC (changes requested):
first very short review

include/alpaka/core/config.hpp
mehmetyusufoglu: Ok, I understood what is going on here. Done.

mehmetyusufoglu:
> You need also add TBB to the CI. Otherwise you test only, if your changes does not break the other backends.

`libtbb-dev `is the [Debian/Ubuntu development package](https://launchpad.net/ubuntu/+source/onetbb) for Intel oneTBB, I added it and [test passes](https://github.com/alpaka-group/alpaka3/actions/runs/18621287449/job/53092771467). 
- I have added `icpx+TBB`  and `clang + TBB `cases as well.
- Should we choose different packages of `libtbb-dev` and create different CI combinations or current tests are enough, i dont know.

psychocoderHPC (approved):

PR #322, opened 2025-10-16
make MdSpanArray compatible to the IMdSpan interface

SimeonEhrig:
I moved the `MdSpanArray` class in an extra file to make it more readable. If you want to know, what I changed in the class `MdSpanArray` please have a look on this commit: d57d4498bcb51390243bf5c21d21a77fa847e8a1


include/alpaka/mem/MdSpanArray.hpp
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        constexpr auto end() const
+        {
+            return MdForwardIterEnd{*this};
+        }
+
+        constexpr auto getConstMdSpan() const
+        {
+            return MdSpanArray<ConstArrayType>(*m_ptr);
```
psychocoderHPC, line 122: here is a bug I introduced already before this PR.
The Alignment is missing. I will fix this after merging this PR.

psychocoderHPC (approved):

PR #326, opened 2025-10-22
SIMD generator support

psychocoderHPC:
- introduce `internal::LoadAsSimd` load trait to avoid assuming SIMD load from a pointer only
- add generator `alpaka::LinearizedIdxGenerator`
- test using generators in all onHost algorithms
- add concepts `alpaka::concepts::IGenerator`
  - @SimeonEhrig It is currently an independent concept. We need to discuss a few points about it, but changes can be made in follow-up PRs.

This PR is required for the upcoming webinar.

note: Test cases partly copied and pasted from the same test without containers but IMO this is better to read compared to highly abstracted helpers. Never the there is room for improvements which should go into a separate PR.

### DIFF
--- a/include/alpaka/SimdPtr.hpp
+++ b/include/alpaka/SimdPtr.hpp
@@ -6,10 +6,10 @@
 
 #include "alpaka/Simd.hpp"
 #include "alpaka/Vec.hpp"
-#include "alpaka/cast.hpp"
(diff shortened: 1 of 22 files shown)


SimeonEhrig (approved):

SimeonEhrig:
@psychocoderHPC I created two issues with upcoming work.

PR #335, opened 2025-10-24
fix single header and provide single header via read the docs

psychocoderHPC:
- fix includes to solve single hadre issues
- add example which points to godbolt
- add resolved single header file to the sphinx documenation

### DIFF
--- a/docs/source/basic/example.rst
+++ b/docs/source/basic/example.rst
@@ -1,14 +1,32 @@
 Code Example
 ============
 
-The following example shows a small hello word example written with alpaka that can be run on different processors.
+Play online with Compiler Explorer
+++++++++++++++++++++++++++++++++++
+
+The followed example can be executed online using `Godbolt Compiler Explorer <https://godbolt.org/z/bE7dbrv5f>`__. [#f1]_
+
+.. literalinclude:: ../../snippets/example/30_elementWiseMultiplication.cpp
+   :language: cpp
+   :start-after: BEGIN-EXAMPLE-elementWiseMultiplication
+   :end-before: END-EXAMPLE-elementWiseMultiplication
+   :dedent:
+
+.. note::
+
+   Do not forget to set the compiler flags to compile with C++20 and optimization if you would like to inspect the assembler, e.g. ``-std=c++20 -O3``.
+   If you would like to see vectorization you should set the corresponding flags, e.g. for **avx2** ``-march=haswell``.
+
+.. [#f1] In pull requests, to test the modified code in Godbolt Compiler Explorer, replace the include with `the link to this file <../alpaka.hpp>`_.
+
+Use alpaka in your project
+++++++++++++++++++++++++++
+
+The following example shows a small vector add example written with alpaka that can be run on different processors.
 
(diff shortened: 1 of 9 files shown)


docs/source/basic/example.rst
```diff
@@ -1,15 +1,34 @@
...
 ============
 
-The following example shows a small hello word example written with alpaka that can be run on different processors.
+Play online with Compiler Explorer
+++++++++++++++++++++++++++++++++++
 
-.. literalinclude:: ../../../example/vectorAdd/src/vectorAdd.cpp
-   :language: C++
-   :caption: vectorAdd.cpp
+To run the example that multiplies two vectors element wise in `Godbolt Compiler Explorer <https://godbolt.org/>`__, replace ``#include <alpaka/alpaka.hpp>`` with the path to `this file <../alpaka.hpp>`_.
```
SimeonEhrig: Why all this extra information? Use only this link: https://godbolt.org/z/551Tv5bPM
It provides everything, what the use needs.

psychocoderHPC: The example below is using `#include <alpaka/alpaka.hpp>`, the text is pointing to the necessary changes required to run this example.
The link you posted points to the **latest dev** branch that is not required to be the same version of the documentation. The link behind `this file` points to the file coresponts to the single-header file of the code version you are currently reading the documentation, it can be a pull request or later if we have releases a special release version.

SimeonEhrig: You lost me.
Than simply remove the example code below. The code is already on Godbolt, therefore we don't need a copy in the readthedocs.

psychocoderHPC: If we do not have it in the documentation, it is not tested in the CI and we will not realize when we maybe break it.
The intention is that the user can see all in our documentation, then have the possibility to use the already prepared godbolt link or copy past the example to any other compiler explorer, e.g. a local instance and start from there.

psychocoderHPC:
@SimeonEhrig I simplified the godbolt section and added the suggested footnote.

docs/source/basic/example.rst
```diff
@@ -1,15 +1,33 @@
...
+   :start-after: BEGIN-EXAMPLE-elementWiseMultiplication
+   :end-before: END-EXAMPLE-elementWiseMultiplication
+   :dedent:
+
+.. note::
+
+   Do not forget to set the compiler flags to compile with C++20 and optimization if you would like to inspect the assembler, e.g. ``-std=c++20 -O3``.
+   If you would like to see vectorization you should set the corresponding flags, e.g. for **avx2** ``-march=haswell``.
+
+.. [#f1] To use the single header of this documentation replace the include with `the link to this file <../alpaka.hpp>`_.
```
SimeonEhrig:
```suggestion
.. [#f1] In pull requests, to test the modified code in Godbolt Compiler Explorer, replace the include with `the link to this file <../alpaka.hpp>`_.
```

SimeonEhrig:
FYI: local build still works.

SimeonEhrig (approved):

PR #336, opened 2025-10-27
fix CMake file creation

psychocoderHPC:
On some systems creating the relative path to `CMAKE_CURRENT_LIST_DIR` in CMake is using `../`. Since we use this path to create a unique destination for CUDA/HIP files in some cases we ended up to create files somewhere on the file system. By replacing `../` with `./` the issue is solved and we can not jump out of the the destination folder `CMAKE_BINARY_DIR`.

### DIFF
--- a/cmake/finalize.cmake
+++ b/cmake/finalize.cmake
@@ -57,16 +57,33 @@ endfunction()
 
 ### Copy a source file to the build tree
 ##
-## The copied file will be placed in a subdirectory named api_name inside the original directory structure.
+## The copied file will be placed in a subdirectory:
+## ${CMAKE_BINARY_DIR}/alpaka_build_files/${api_name}/${PROJECT_NAME}/<relative_to_project_root>/${SRC_FILE_NAME}
 ## All source file properties set on the original file will be copied to the new file.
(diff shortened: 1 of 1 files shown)


cmake/finalize.cmake
SimeonEhrig:
```suggestion
### Copy a source file to the build tree
##
## The copied file will be placed in a subdirectory:
## ${CMAKE_BINARY_DIR}/alpaka_build_files/${api_name}/${PROJECT_NAME}/<relative_to_project_root>/${SRC_FILE_NAME}
## All source file properties set on the original file will be copied to the new file.
## The path of the copied file is returned via the OUT_VAR variable.
##
## example:
##   CMAKE_BINARY_DIR=build (defined by CMake configure)
##   PROJECT_NAME=vectorAdd (defined by CMake configure)
##
##   SRC_FILE=alpaka3/example/src/vector_add.cpp
##   api_name=cuda
##   OUT_VAR=alpaka3/build/alpaka_build_files/cuda/vectorAdd/example/src/vector_add.cpp
```

psychocoderHPC: updated

PR #337, opened 2025-10-29
add IDataSource interface

SimeonEhrig:
- IDataSource is the minimum interface of a data object.
- A IDataSource object can be only used to read data.
- Replaces the `IGenerator` and `IGeneratorOrMdSpan` interface

Each memory interface extents the next one

IDataSource -> IMdSpan* -> IView -> IBuffer

* IMdSpan extents also the memory access method. Instead returning a value (read only), it returns a reference (read/write).

# TODO

- [x] move `{ mut_t[vec] } -> std::same_as<T::reference>` to `IDataSource` for the case, that `T` has the type `reference`
- [x] resolve new todos
- [x] remove `IGenerator` and `IGeneratorOrMdSpan`
- [x] write and update documentation


include/alpaka/mem/concepts/IDataSource.hpp
```diff
@@ -0,0 +1,107 @@
...
+#include "alpaka/mem/concepts/ExpectedValueType.hpp"
+#include "alpaka/trait.hpp"
+
+#include <concepts>
+
+namespace alpaka::concepts
+{
+    namespace impl
+    {
+        /** @brief Interface concept for objects describing a multidimensional data source.
```
SimeonEhrig, line 19: @AntonReinhard Can you please check the doc string. @psychocoderHPC does the full review.

include/alpaka/mem/concepts/IDataSource.hpp
```diff
@@ -0,0 +1,107 @@
...
+         *
+         * @details
+         *
+         * An object that implements the interface returns a value for a multidimensional index. Therefore, it behaves
+         * like multidimensional memory that can only be read. It is not permitted to write a new value to an index
+         * position. An `IDataSource` object has an immutable, fixed multidimensional size. The `IDataSource` object is
+         * not required to reference the storage. It can create data instant.
```
AntonReinhard:
```suggestion
         * not required to reference the storage. It may create or calculate the data instead of reading it from memory.
```

AntonReinhard (commented):
I'm slightly confused. The docs say (and the name implies) that the data source doesn't have to just be memory that we read from. But if it's something like a function that can just be calculated from the given indices, then it could never be mutable, but the concept seems to require a mutable version to exist?

SimeonEhrig:
> I'm slightly confused. The docs say (and the name implies) that the data source doesn't have to just be memory that we read from. But if it's something like a function that can just be calculated from the given indices, then it could never be mutable, but the concept seems to require a mutable version to exist?

I you mean this requirement?

```c++
                  requires
                      (requires { typename T::reference; } &&
                          requires {{ mut_t[vec] } -> std::same_as<typename T::reference>; }) ||
                      (!requires { typename T::reference; } &&
                          requires {{ mut_t[vec] } -> std::same_as<typename T::value_type>; });
```

SimeonEhrig:
> > I'm slightly confused. The docs say (and the name implies) that the data source doesn't have to just be memory that we read from. But if it's something like a function that can just be calculated from the given indices, then it could never be mutable, but the concept seems to require a mutable version to exist?
> 
> I you mean this requirement?
> 
> ```c++
>                   requires
>                       (requires { typename T::reference; } &&
>                           requires {{ mut_t[vec] } -> std::same_as<typename T::reference>; }) ||
>                       (!requires { typename T::reference; } &&
>                           requires {{ mut_t[vec] } -> std::same_as<typename T::value_type>; });
> ```

In a VC, we clarified that this is actual not correct. I will try to move this term `(requires { typename T::reference; } && requires {{ mut_t[vec] } -> std::same_as<typename T::reference>; }) ` to `IMdSpan`. If this is working, I can also simplify the concept interface to:

```c++
        template<typename T>
        concept IDataSource
            = requires(T t, alpaka::Vec<typename T::index_type, T::dim()> vec){}
``

File: include/alpaka/mem/concepts/IMdSpan.hpp
Comment (psychocoderHPC) [inline include/alpaka/mem/concepts/IMdSpan.hpp]: Could you change this to 

```
(T::dim() != 1u) ||
                      (T::dim() == 1u && ....
```

**Review by psychocoderHPC — CHANGES_REQUESTED** [review]

File: include/alpaka/mem/concepts/IDataSource.hpp [hunk 3]
```diff
@@ -0,0 +1,84 @@
...
+            // if T has a member type `reference`, it must at least satisfy the IMdSpan interface
+            // in this case, this definition of the access operator is no longer valid
+            requires
+                      (!requires { typename T::reference; } &&
+                          requires {{ t[vec] } -> std::same_as<typename T::value_type>; })
+                          || requires { typename T::reference; };
+
+            // check 1-dimensional access operator
+            requires
+                      (T::dim() != 1u) ||
```
psychocoderHPC: check if we can only indent once or if the formater kicked in and aligned the code.

include/alpaka/mem/concepts/IDataSource.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+            /* check multi-dimensional access operator
+            if T has a member type `reference`, it must at least satisfy the IMdSpan interface
+            in this case, this definition of the access operator is no longer valid */
+            requires (
+                // guard condition: if one of the requirements is satisfied, ignore the actual requirement
+                requires { typename T::reference; } ||
+                // actual requirement:
+                !requires { typename T::reference; } &&
+                    requires {{ t[vec] } -> std::same_as<typename T::value_type>; });
```
AntonReinhard: Doesn't this now mean that if the type has a `::reference` defined, that it doesn't have to implement an access operator at all?

I think something like this would be more explicit and less error-prone than boolean algebra:
```suggestion
            requires (requires {{ t[vec] } -> std::convertible_to<typename T::value_type>; });
```

AntonReinhard: Or maybe there's a better concept than convertible_to that checks  that it is the same base type, regardless of const/ref/volatile specifiers

SimeonEhrig: @psychocoderHPC @ikbuibui Anton solved my problem. I can check with `std::convertible_to`, if I can assign an value to a value or a reference to an value or a reference to a reference.

```c++
template<typename TFrom, typename TTo>
concept assignable_to = std::same_as<std::decay_t<TFrom>, std::decay_t<TTo>> && std::convertible_to<TFrom, TTo>;
 ```

psychocoderHPC (approved):

PR #339, opened 2025-11-03
 Add Warp ops

mehmetyusufoglu:
Warp operations,.

1. Introduce an `alpaka::warp` trait  like `ConceptWarp` for size, masks, votes, and shuffles with tag-based detection.
2. Add  warp implementations: single-thread CPU emulation, CUDA/HIP intrinsics, and SYCL subgroup support? .
3.  Front-end helpers (`onAcc::warp::all/shflDown` etc.) that route through traits while allowing optional explicit warp control. 

Mainline alpaka used CUDA and HIP warp collectives through the concrete class `warp::WarpUniformCudaHipBuiltIn `(alpaka/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp). Trait specialisations were written directly against that class, and every call funnelled through `interface::ImplementationBase<ConceptWarp, WarpUniformCudaHipBuiltIn>` before invoking the backend intrinsic. Alpaka3 removes the class layer and redirects the same trait names to specialisations keyed on the ApiTag and DeviceTag.

**Guidance**
-  For kernels, always call `alpaka::onAcc::warp::getSize(acc) `to respect the actual runtime warp/sub-group width.
-  Use DeviceProperties::getPreferredWarpSize() or DeviceProperties::getWarpSizes() when host code needs the actual warp size that the backend reports at runtime. 
-  Use  alpaka::getWarpSize(apiTag, deviceKind) or alpaka::onAcc::warp::getSize(apiTag, deviceKind) when you’re writing compile-time logic (static asserts, template specialisations) and only have the API/device tags available

**Tests compared to mainline(old) alpaka**
Screenshot from 2025-11-05 13-02-49

**Check**
- [x] Squash
- [ ] Go over tests
- [ ] User documentation


psychocoderHPC:
I deleted my last comment because I misread your code, sry.

include/alpaka/onAcc/warp.hpp
```diff
@@ -0,0 +1,273 @@
...
+
+#include "alpaka/Vec.hpp"
+#include "alpaka/interface.hpp"
+#include "alpaka/onAcc/Acc.hpp"
+#include "alpaka/tag.hpp"
+#include "alpaka/warp/Traits.hpp"
+
+#include <cstdint>
+
+namespace alpaka::onAcc::warp
```
mehmetyusufoglu, line 25: @psychocoderHPC  since we have "scope" namespace i preferred a new namespace but i am not sure.

psychocoderHPC: it looks ok, for now I also do not have a fixed opinion for now.

psychocoderHPC:
@mehmetyusufoglu all files in the folder warp which are API spezific should be moved into the folder `api/*`.
Api spezific implementation are in the namespace `alpaka::internal` and are not spread around in the code base, they are in the coresponding folder e.g. `alpaka/api/onApi` for oneApi spezific code if it is used oneApi extensions else if it is pure sycl it is in the sycl folder.

mehmetyusufoglu:
> @mehmetyusufoglu all files in the folder warp which are API spezific should be moved into the folder `api/*`. Api spezific implementation are in the namespace `alpaka::internal` and are not spread around in the code base, they are in the coresponding folder e.g. `alpaka/api/onApi` for oneApi spezific code if it is used oneApi extensions else if it is pure sycl it is in the sycl folder.

ok from` include/alpaka/warp` to `include/alpaka/api/*/`: 

`UniformCudaHipBuiltIn.hpp  `-> api/unifiedCudaHip/warp.hpp 
` GenericSycl.hpp` ->  api/oneApi/warp.hpp   (not  api/syclGeneric/warp.hpp )
`SingleThread.hpp` ->  api/host/warp.hpp 
`Traits.hpp` -> stays (it's API-agnostic)

include/alpaka/warp/Traits.hpp
```diff
@@ -0,0 +1,291 @@
...
+#pragma once
+
+#include "alpaka/api/concepts/api.hpp"
+#include "alpaka/api/trait.hpp"
+#include "alpaka/core/common.hpp"
+#include "alpaka/tag.hpp"
+
+#include <cstdint>
+
+namespace alpaka::warp
```
psychocoderHPC: This file should be in the folder `alpaka/onAcc` and since it is a single file it does not require a separate folder. So please move it to `onAcc/internal/warp.hpp`
All these traits should be in `internal` because the user is not using them directly. The user interface is in `onAcc/Warp.hpp".
The namespace should also be `alpaka::onAcc::warp` since all warp functions will only be used within a kernel and never on the host.

include/alpaka/onAcc/warp.hpp
```diff
@@ -0,0 +1,273 @@
...
+ * Supplies a uniform warp API across CUDA, HIP, SYCL, and host-emulation accelerators.
+ *
+ * Some example usages:
+ * - consteval `alpaka::getWarpSize(api::Cuda{}, deviceKind::NvidiaGpu{})` for tag-driven compile-time logic.
+ * - runtime `alpaka::getWarpSize(deviceSpec)` when host code needs the width before launching work.
+ * - device-side `alpaka::onAcc::warp::getSize(acc)` to query the active warp inside a kernel.
+ * - device-side `alpaka::onAcc::warp::shfl(acc, 42, 0u) == 42`
+ */
+
+#pragma once
```
psychocoderHPC, line 15: Please rename this file to `warp.hpp`. We only name files with a capital letter in cases where it contains a class with the same name. This is a collection of interface functions there we start the file with a small letter.

include/alpaka/api/hip/Api.hpp
```diff
@@ -97,5 +97,14 @@ namespace alpaka
                 return 16u;
             }
         };
+
+        template<>
+        struct GetWarpSize::Op<api::Hip, deviceKind::AmdGpu>
+        {
+            consteval uint32_t operator()(api::Hip const, deviceKind::AmdGpu const) const
+            {
+                return 64u;
```
psychocoderHPC, line 167: please have a look to the mainline alpaak how the warp size is calculated. There are devices with warp size of 32 and some with warp size of 64.

include/alpaka/api/oneApi/Api.hpp
```diff
@@ -130,6 +139,15 @@ namespace alpaka
                 return 16u;
             }
         };
+
+        template<concepts::GpuType T_DeviceKind>
+        struct GetWarpSize::Op<api::OneApi, T_DeviceKind>
+        {
+            consteval uint32_t operator()(api::OneApi const, T_DeviceKind const) const
+            {
+                return 32u;
```
psychocoderHPC, line 148: This is wrong but keep it for now and conceptrate on the otehr changes. The reason is that in sycl you cn have different warp size. We need to change in sycl how a kernel is started and implement a synamic warp size dispatchment to have a compile time warp size in the kernel. This also requires that we must pass the `acc` to this trait be cause we need to store the warpsize in the acc dictionary.

psychocoderHPC (changes requested):
There are some required refactorings and after the refactoring we need to fix how warp size is handeled in sycl. Please push first the refactorings. The sycl change requires most likly a seperate PR before.
If some points are unclear please talk to me in the office.

mehmetyusufoglu:
> There are some required refactorings and after the refactoring we need to fix how warp size is handeled in sycl. Please push first the refactorings. The sycl change requires most likly a seperate PR before. If some points are unclear please talk to me in the office.

ok doing.

include/alpaka/warp/Traits.hpp
mehmetyusufoglu: done

include/alpaka/onAcc/warp.hpp
mehmetyusufoglu: done

include/alpaka/api/hip/Api.hpp
mehmetyusufoglu: I used warpSize already defined by cuda and hip headers like [here](https://github.com/alpaka-group/alpaka/blob/3aafde5d2b32853ef5362a581624d2c24e1604d9/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp#L39C17-L39C33) of mainline alpaka approach; but i had to change hip part a little bit because of rocm (especially rocm 7.0) failures and warnings. Since native CUDA headers expose `warpSize` as a true compile-time constant on the host its implementation was directly similar to alpaka mainline.

include/alpaka/api/trait.hpp
```diff
@@ -159,6 +172,18 @@ namespace alpaka
...
     }
 
+    /**
+     * Get the number of threads grouped into a warp (lock-step execution group).
+     *
+     * This consteval helper is meant for template code that only knows the API/device tags. When a concrete device or
+     * accelerator instance is available, prefer the runtime APIs (`DeviceProperties::getWarpSizes()`,
+     * `alpaka::onAcc::warp::getSize(acc)`) so backend-specific tuning remains accurate.
+     */
+    consteval uint32_t getWarpSize(concepts::Api auto const api, alpaka::concepts::DeviceKind auto const deviceType)
```
mehmetyusufoglu, line 182: @fwyzard  Do you think exposing `getWarpSize` from host and at compile-time is a good idea or should we keep warp size  access only from kernel and with acc argument (from kernel `alpaka::onAcc::warp::getSize(acc)`) ? I think this might be good as an additional feature; but i cannot think of usage cases.

Since cuda defined `warpSize` at one of its headers and we use it [in mainline alpaka](https://github.com/alpaka-group/alpaka/blob/3aafde5d2b32853ef5362a581624d2c24e1604d9/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp#L39)   and in alpaka3 (
https://github.com/alpaka-group/alpaka3/pull/339/files#diff-b14534dd0e79682dde5ee79dfee1d756d70f2f7089736ec37d7aac8829889940) , exposing it from host and compile time is reasonable for me. Alpaka3 will be exposing a backend variable in that sense while alpaka mainline was using but was not exposing explicitly.

psychocoderHPC: Please keep exposing compile time warp size for now only within the kernel. In Sycl a device supports different warp sizes and not only a single one, so it can not compile time and even if, keep the changes small and improvements can be done independent from this PR:

include/alpaka/api/hip/Api.hpp
```diff
@@ -97,5 +102,42 @@ namespace alpaka
...
+
+        template<>
+        struct GetWarpSize::Op<api::Hip, deviceKind::AmdGpu>
+        {
+            // Follow legacy Alpaka by sourcing HIP’s runtime `warpSize`, but guard it with a constant-evaluation
+            // check so host code still sees a compile-time value while device kernels receive the builtin result.
+#if ALPAKA_LANG_HIP
+            // HIP’s `warpSize` is a device builtin with a non-constexpr conversion, so we rely on
+            // `std::is_constant_evaluated()` to substitute documented wavefront defaults for compile-time requests.
+            // gfx9 hardware reports 64-lane wavefronts, while gfx10 and later default to 32 lanes per HIP docs.
```
psychocoderHPC: You comment is saying depending on the architecture you set the warpsize to 32 but I can not see it in the code. Please check the alpaka mainline code to see how it can be implemented.

psychocoderHPC (changes requested):

include/alpaka/api/trait.hpp
fwyzard: With the alpaka 2 approach, the warp size cannot be a compile-time constant on the host, because at compile time it's not possible to know what GPU one will be using.

E.g. different generation of AMD GPUs have different warp size, so making it a compile time constant would require using different _types_ for different GPU generations.

include/alpaka/api/unifiedCudaHip/warp.hpp
```diff
@@ -0,0 +1,429 @@
...
+#include "alpaka/core/common.hpp"
+#include "alpaka/core/config.hpp"
+#include "alpaka/onAcc/internal/warp.hpp"
+
+#include <cstdint>
+#include <type_traits>
+
+namespace alpaka::onAcc::warp::detail
+{
+#if ALPAKA_LANG_CUDA
```
psychocoderHPC, line 18: Please move the CUDA part into alpaka/api/cuda and the HIP part into alpaka/api/hip

include/alpaka/api/trait.hpp
fwyzard: I understand that in alpaka 3 the Accelerator objects can have more compile-time properties related to the execution space ?

Like, compile time constant group size ?
In that case it could be nice to also have a compile time warp size using the same approach.

psychocoderHPC (changes requested):

include/alpaka/api/trait.hpp
psychocoderHPC: Yes alpaka3 accelerators can carry compile time properties those can be used on within the kernel.
For sycl we need to implement that we can provide the warpsize at compile time this will be implemented in a different PR before we merge this.

mehmetyusufoglu: -  Runtime (host), `alpaka::getPreferredWarpSize(device)`, Preferred warp size for a specific device
-  Runtime (host), `alpaka::getWarpSizes(device)`, All supported warp SIZES (not size) for a specific device
-  Runtime (device), `alpaka::onAcc::warp::getSize(acc)`, Query active warp width inside a kernel
-  Compile-time, (consteval, host-side),` alpaka::getWarpSize(api, deviceKind)`, Tag-driven constant for template logic ( @psychocoderHPC  you dont want this ??)

include/alpaka/api/hip/Api.hpp
mehmetyusufoglu: I wrote it just as a comment, I wanted to explain what built-in variable `warpSize` should be expected in the comment.  Alpaka mainline is only using `warpSize` but it created problems at the CI therefore i had to come up with this solution.

mehmetyusufoglu: Ok I understand your point, i was checking an alpaka mainline not updated, Andrea added new [warp compile time commits](https://github.com/alpaka-group/alpaka/blob/3aafde5d2b32853ef5362a581624d2c24e1604d9/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp#L75) at September 29 :) Sorry.

include/alpaka/api/unifiedCudaHip/warp.hpp
mehmetyusufoglu: done, thanks.

include/alpaka/api/hip/Api.hpp
mehmetyusufoglu: done yesterday

psychocoderHPC:
PR is replaced by #353

PR #340, opened 2025-11-03
refactor device global memory

psychocoderHPC:
- add support for sycl oneAPI device global memory
- seperate internal code and user interface
- test only features those are supported by all APIs and devices


tests/unit/deviceGlobalMem/deviceGlobalMem.cpp
```diff
@@ -16,11 +16,16 @@ using namespace alpaka::onHost;
 
 using TestApis = std::decay_t<decltype(allBackends(enabledApis, onHost::example::enabledExecutors))>;
 
+/* We can not check passing the attributes 'const', 'static' or 'constexpr' because this is not supported by OneApi
+ * 2025.2 with AMD gpus.
+ * Therefore, we are testing only empty attributes and 'inline' attribute to fucus on supported code only.
```
ikbuibui:
```suggestion
 * Therefore, we are testing only empty attributes and 'inline' attribute to focus on supported code only.
```

ikbuibui (approved):

PR #342, opened 2025-11-05
fix MdSpanArray

psychocoderHPC:
- calling `data()` was not possible because of implicit cast issues
- dereverencing the current position of the MdSpanArray was broken because a CArray can not simply be dereferenced


ikbuibui (approved):
Looks okay, but I think you dont need to access mPtr/data() through `this`. 
If you want to keep the `this` that is also okay, just let me know and ill merge

psychocoderHPC:
it can stay but you are right that it is not required

PR #343, opened 2025-11-05
CI: add OneApi 2025.3

psychocoderHPC:
- add OneApi 2025.3
- increase OneApi 2025.1 and 2025.2 to the latest patch level


ikbuibui (approved):
Im not sure why the CI is stuck. I don't even see the stuck jobs in the `ci.yml`. Everything else passes andis fine by me.

psychocoderHPC:
ahh the reason is that these jobs are required but since i switched to the latest patch level the jobs habe new names. It required to update the required job list in github. will to it after mdspan is merged

PR #344, opened 2025-11-06
fix loop counter being incremented twice

ikbuibui:
The grayscale example can be configured to run multiple times, but the loop counter was incremented twice in every loop.
This PR fixes this. Counter is now only incremented in the for loop statement


psychocoderHPC:
The result of this bug should be that we get half of the real archived bandwidth.

psychocoderHPC (approved):

PR #350, opened 2025-11-11
fix single header

psychocoderHPC:
Some includes in the sycl/cuda/hip section was within `#ifdef` and avoid the
single header generator to create a valid header for CUDA/HIP.

- To solve some cyclic dependencies a new header `alpaka/api/math.hpp` is introduced.
- move all alpaka includes out of the language guards
  - This has the additional advantage of enabling CUDA/SYCL or HIP will not introduce new possible cyclic include issues.

The second commits add the clang OpenMP library to the tsan suppression files equally to what we did for GNU OpenMP.
The reason is that we got the same false positive we had already with GNU.

note: I tested CUDA and HIP gotbold


include/alpaka/api/math.hpp
```diff
@@ -0,0 +1,38 @@
...
+        };
+    };
+
+    template<typename T_Api>
+    constexpr decltype(auto) getMathImpl(T_Api const api)
+    {
+        return GetMathImpl::Op<T_Api>{}(api);
+    }
+} // namespace alpaka::trait
+#endif
```
ikbuibui: dead code

ikbuibui (approved):

PR #352, opened 2025-11-13
compile time warp size in kernel

psychocoderHPC:
- provide a way to query the warp size at compile time in the kernel

note: We will always only support one warp size per device, and this is the warp size from the device properties.
Generating the warp size (subgroup size) in sycl is very complex and requires a two-stage approach.
I am not happy with this, but from my testing, it guarantees that the kernel `parallel_for()` in sycl is annotated correctly with the compile-time values, and it is not falling back to any other default.

### DIFF
--- a/include/alpaka/api/cuda/computeApi.hpp
+++ b/include/alpaka/api/cuda/computeApi.hpp
@@ -1,0 +1,26 @@
+/* Copyright 2025 René Widera
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include "alpaka/api/unifiedCudaHip/ComputeApi.hpp"
+#include "alpaka/core/config.hpp"
+
+#include <type_traits>
+
+#if ALPAKA_LANG_CUDA
+
+namespace alpaka::onAcc::unifiedCudaHip::internal
+{
+    template<>
+    struct WarpSize::Get<alpaka::deviceKind::NvidiaGpu>
+    {
+        constexpr auto operator()() const
+        {
+            return std::integral_constant<uint32_t, 32u>{};
+        }
+    };
+} // namespace alpaka::onAcc::unifiedCudaHip::internal
+
+#endif
--- a/include/alpaka/api/oneApi/Queue.hpp
+++ b/include/alpaka/api/oneApi/Queue.hpp
@@ -9,6 +9,7 @@
 #include "alpaka/api/syclGeneric/Queue.hpp"
 #include "alpaka/api/syclGeneric/onAcc.hpp"
 #include "alpaka/core/config.hpp"
+#include "alpaka/core/syclConfig.hpp"
 #include "alpaka/onAcc/internal/globalMem.hpp"
 #include "alpaka/onHost/internal/interface.hpp"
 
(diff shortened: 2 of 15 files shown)


tests/unit/cvecKernelCall/cvecKernelCall.cpp
```diff
@@ -77,13 +83,19 @@ TEMPLATE_LIST_TEST_CASE("CVec frame extent kernel call", "", TestApis)
...
-    ALPAKA_FN_ACC void operator()(T const& acc, auto result) const
+    template<onAcc::concepts::Acc T_Acc>
+    ALPAKA_FN_ACC void operator()(T_Acc const& acc, auto result) const
     {
         // compile will fail if the type is silently cast to non CVec type
         alpaka::concepts::CVector auto blockThreadCount = acc[alpaka::layer::thread].count();
         static_assert(blockThreadCount.x() == 1u);
         result[0] = blockThreadCount.x() == 1u;
+
+        // warp size must be accessible at compile time, this is only possible if we us ethe accelertor type
```
ikbuibui:
```suggestion
        // warp size must be accessible at compile time, this is only possible if we use the accelerator type
```

tests/unit/cvecKernelCall/cvecKernelCall.cpp
```diff
@@ -36,6 +36,12 @@ struct KernelCVecFrameExtents
         static_assert(blockThreadCount2 == blockThreadCount);
         static_assert(frameExtent.x() == 43u);
         result[0] = frameExtent.x() == 43u;
+
+        // warp size must be accessible at compile time, this is only possible if we us ethe accelertor type
```
ikbuibui:
```suggestion
        // warp size must be accessible at compile time, this is only possible if we use the accelerator type
```

include/alpaka/api/cuda/computeApi.hpp
```diff
@@ -0,0 +1,25 @@
+/* Copyright 2025 René Widera
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include "alpaka/CVec.hpp"
```
ikbuibui:
```suggestion
```
To use integral_constant instead of CVec

include/alpaka/api/cuda/computeApi.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include "alpaka/api/unifiedCudaHip/ComputeApi.hpp"
+#include "alpaka/core/config.hpp"
+
```
ikbuibui, line 9:
```suggestion

#include <type_traits>

```
To use integral_constant instead of CVec

include/alpaka/api/cuda/computeApi.hpp
```diff
@@ -0,0 +1,25 @@
...
+#if ALPAKA_LANG_CUDA
+
+namespace alpaka::onAcc::unifiedCudaHip::internal
+{
+    template<>
+    struct WarpSize::Get<alpaka::deviceKind::NvidiaGpu>
+    {
+        constexpr auto operator()() const
+        {
+            return CVec<uint32_t, 32u>{};
```
ikbuibui:
```suggestion
            return std::integral_constant<uint32_t, 32u>{};
```

ikbuibui: I like this better because then you don't need to say `.x()`. But you will have to change this in a lot of places (I think 39)

include/alpaka/onAcc/Acc.hpp
```diff
@@ -65,6 +65,18 @@ namespace alpaka::onAcc
...
+
+        /** Get the warp size
+         *
+         * To be able to use the warp size as constexpr value you must call the static function via the type
+         * `Acc_t::getWarpSize()` else the warp size can only be used as runtime value.
+         * Alternative you can use the free function onAcc::getWarpSize<Acc_t>();
+         */
+        static constexpr uint32_t getWarpSize()
+        {
+            constexpr uint32_t w = ALPAKA_TYPEOF(std::declval<T_Storage>()[object::warpSize]){}.x();
```
ikbuibui:
```suggestion
            constexpr uint32_t w = ALPAKA_TYPEOF(std::declval<T_Storage>()[object::warpSize]){};
```
or 

```suggestion
            constexpr uint32_t w = ALPAKA_TYPEOF(std::declval<T_Storage>()[object::warpSize])::value;
```
or 

```suggestion
            constexpr uint32_t w = ALPAKA_TYPEOF(std::declval<T_Storage>()[object::warpSize]){}.value;
```

include/alpaka/api/syclGeneric/Queue.hpp
```diff
@@ -29,6 +29,122 @@ namespace alpaka::onHost
...
+                    return fn(CVec<uint32_t, 8u>{});
+                case 16u:
+                    return fn(CVec<uint32_t, 16u>{});
+                case 32u:
+                    return fn(CVec<uint32_t, 32u>{});
+                default:
+                    throw std::runtime_error(
+                        std::string("Sycl warp size runtime dispatch, unsupported warpSize: ")
+                        + std::to_string(warpSize));
+                    return fn(CVec<uint32_t, 1u>{});
```
ikbuibui: Won't return after throw, also in 3 places below

include/alpaka/api/oneApi/Queue.hpp
ikbuibui, line 387: Unrelated to this PR, but the `threadSpec` and `frameSpec` enqueues share a lot of code, it can possibly be shifter to a helper

ikbuibui (approved):

include/alpaka/api/oneApi/Queue.hpp
psychocoderHPC: offline discussed: joining the code could have performance implications and should be handled in a separate PR if needed. The same issue may exist in other Api implementations too.

psychocoderHPC:
@ikbuibui The PR is now ready

PR #353, opened 2025-11-14
warp operations

psychocoderHPC:
- provide warp function for all backends
- add tests
- add doxygen documentation

Compared to mainline alpaka voting function  like `any`, `all` returning `bool`.
The warp size and lane index is a unisgned 32bit integral value.

This PR based on #339, it removed special handling of executors using a single thread per block and applied many cleanups to keep the PR small.

All implementations are tested with CUDA 12.5, HIP 6.2 and OneApi 2025.2 (CUDA 12.6, HIP 6.3.4 - for Intel GPU, CPU, AMD GPU/gfx1100 and NVIDIA A40)

### DIFF
--- a/include/alpaka/alpaka.hpp
+++ b/include/alpaka/alpaka.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2024 René Widera
+/* Copyright 2024 René Widera, Mehmet Yusufoglu
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -10,7 +10,11 @@
 #include "alpaka/Vec.hpp"
 #include "alpaka/api/api.hpp"
 #include "alpaka/api/cpu.hpp"
+#include "alpaka/api/cuda/warp.hpp"
+#include "alpaka/api/hip/warp.hpp"
+#include "alpaka/api/host/warp.hpp"
 #include "alpaka/api/oneApi.hpp"
+#include "alpaka/api/oneApi/warp.hpp"
 #include "alpaka/api/unifiedCudaHip.hpp"
 #include "alpaka/apply.hpp"
 #include "alpaka/core/Dict.hpp"
@@ -33,6 +37,7 @@
 #include "alpaka/onAcc/interface.hpp"
 #include "alpaka/onAcc/memFence.hpp"
 #include "alpaka/onAcc/tag.hpp"
+#include "alpaka/onAcc/warp.hpp"
 #include "alpaka/onHost/Device.hpp"
 #include "alpaka/onHost/DeviceSelector.hpp"
 #include "alpaka/onHost/Queue.hpp"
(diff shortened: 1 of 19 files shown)


psychocoderHPC:
@ikbuibui The PR is now ready for a review

include/alpaka/onAcc/warp.hpp
```diff
@@ -0,0 +1,273 @@
...
+     * @{
+     */
+    template<concepts::Acc T_Acc>
+    constexpr uint32_t getSize()
+    {
+        return internal::getSize<T_Acc>();
+    }
+
+    template<concepts::Acc T_Acc>
+    constexpr uint32_t getSize(T_Acc const& acc)
```
ikbuibui, line 124:
```suggestion
    constexpr uint32_t getSize(T_Acc const&)
```

psychocoderHPC:
@ikbuibui ping

psychocoderHPC:
Merged it, if there is something to change, we can push it in a separate PR. I would like to start with a test for the warp level and provide worker descriptions for `makeIdxMap`

include/alpaka/onAcc/warp.hpp
```diff
@@ -0,0 +1,273 @@
...
+        using Api = ALPAKA_TYPEOF(acc[object::api]);
+        return internal::Activemask::Op<Acc, Api>{}(acc, Api{});
+    }
+
+    /** Return the lane index of the current thread within its warp. */
+    constexpr uint32_t getLaneIdx(alpaka::onAcc::concepts::Acc auto const& acc)
+    {
+        using Acc = ALPAKA_TYPEOF(acc);
+        using Api = ALPAKA_TYPEOF(acc[object::api]);
+        return internal::GetLanIdx::Op<Acc, Api>{}(acc, Api{});
```
ikbuibui, line 45:
```suggestion
        return internal::GetLaneIdx::Op<Acc, Api>{}(acc, Api{});
```
Otherwise it reads like a local area network idx (Needs to be fixed for all APIs)

include/alpaka/alpaka.hpp
```diff
@@ -10,7 +10,11 @@
 #include "alpaka/Vec.hpp"
 #include "alpaka/api/api.hpp"
 #include "alpaka/api/cpu.hpp"
+#include "alpaka/api/cuda/warp.hpp"
+#include "alpaka/api/hip/warp.hpp"
+#include "alpaka/api/host/warp.hpp"
```
ikbuibui, line 15: These warp includes should be moved to `include/alpaka/api/cuda.hpp"`, etc

ikbuibui (commented):
The tests are written with a race condition (all warps write to the same location in global memory), but this is not affecting the correctness of the tests.
No functional suggestions, only renaming and moving things around
The `WarpSize::Get` could also be possibly moved into the warp file instead of having a seperate `computeApi.hpp`

include/alpaka/onAcc/warp.hpp
psychocoderHPC: will be sneaked in with my next pr^^

include/alpaka/alpaka.hpp
psychocoderHPC: will try, I fought with the includes because of some cyclic issues, lets check again if I can move it. I do not remember whats was the exact reason

PR #355, opened 2025-11-17
Alpaka tuner

TimHanel00:
This tool is tightly integrated into Alpakas existing infrastructure by mirroring function calls for usability and utilizing existing concepts/structures (such as enqueue, or Vector, KernelBundle). Adds only one optional dependency -- [json/nlohmann](https://github.com/nlohmann/json)). Instruments kernel invocation by adjusting parameter of the frame specification, compile-time values and user defined parameters at each execution step. Provides the method to define and traverse multi-dimensional tuning spaces. Tracks timing measurement at runtime for a specific tuning Context (e.g. device,executor,kernel). Enables definitions of constraint relationships between one or multiple parameters. Allows persistent offline storage between tuning runs using the [json/nlohmann](https://github.com/nlohmann/json) dependency.


include/alpaka/onHost/tune/concepts.hpp
```diff
@@ -0,0 +1,281 @@
+//
+// Created by tim on 08.10.25.
+//
+
+#ifndef CONCEPTS_H
```
psychocoderHPC, line 5: please do not use macros to guard hpp files.

#pragma one 

is what should be used, see all other files in alpaka

psychocoderHPC (changes requested):

psychocoderHPC:
Thanks for your pull request. The review will take some time and we will most likely go also in person over the code. If I see it correctly it instandly doubles the amount of alpaka code.

psychocoderHPC:
note: all files need a license header! woth you as author

include/alpaka/onHost/tune/adjust/api.hpp
```diff
@@ -0,0 +1,10 @@
+//
```
psychocoderHPC, line 1: please remove tjis by the real alpaka header.

include/alpaka/onHost/tune/adjust/tunerAdjustGpu.hpp
```diff
@@ -0,0 +1,108 @@
+//
+// Created by tim on 05.03.25.
+//
+
+#ifndef TUNERGPU_HPP
+#define TUNERGPU_HPP
+
+#if defined(ALPAKA_LANG_CUDA) || defined(ALPAKA_LANG_HIP) || defined(ALPAKA_LANG_SYCL)
```
psychocoderHPC, line 8: we need to look into this. If there is no cuda code in there is no need to guard it by the language. IMO the code compiles even if only host api is available

psychocoderHPC:
The tuner has been moved into its own project. After @tim finished the full seperation within his reposetory it will be come available in https://github.com/alpaka-group

PR #356, opened 2025-11-20
HIP: add newest GPUs to the config

psychocoderHPC:


SimeonEhrig:
Why is the ordering inverted? Else biggest gfx version is on top.

include/alpaka/core/hipConfig.hpp
```diff
@@ -25,7 +25,20 @@
  * https://llvm.org/docs/AMDGPUUsage.html#processors
  */
 
-#        if defined(__gfx1153__)
+#        if defined(__gfx1200__)
+/* RDNA 4 dGPU (RX 9060 XT) */
```
SimeonEhrig:
```suggestion
/* RDNA 4 dGPU (RX 9060 / RX 9060 XT) */
```

include/alpaka/core/hipConfig.hpp
```diff
@@ -25,7 +25,20 @@
...
+#            define ALPAKA_AMDGPU_ARCH ALPAKA_VRRPP_TO_VERSION(120000)
+#        elif defined(__gfx1201__)
+/* RDNA 4 dGPU (RX 9070 / 9070 XT) */
```
SimeonEhrig:
```suggestion
/* RDNA 4 dGPU (RX 9070 / RX 9070 GRE / 9070 XT) */
```

include/alpaka/core/hipConfig.hpp
psychocoderHPC: updated

include/alpaka/core/hipConfig.hpp
psychocoderHPC: updated

psychocoderHPC:
> Why is the ordering inverted? Else biggest gfx version is on top.

The newest is mostly more important and this avoids that we need to scroll.

SimeonEhrig:
> > Why is the ordering inverted? Else biggest gfx version is on top.
> 
> The newest is mostly more important and this avoids that we need to scroll.

Because the rest of the list (existing entries) is ordered by the gfx level starting with highest value. I suggested it to be consistent.

psychocoderHPC:
@SimeonEhrig I fixed the order

SimeonEhrig (approved):

PR #357, opened 2025-11-21
fix using alpaka via CMake FetchContent

psychocoderHPC:
When FetchContent is used, it is not possible to set the languages within the target. The reason is that the language must be set in the top-level domain during the CMake configure time.

To add support for using alpaka with FetchContent, a helper macro is required.

Note: I have not found anything via Google that describes how you can enable languages within a library when using CMake FetchContent; therefore, I came up with my own solution. 
I was not able to integrate this into `alpaka_finalize()` or push this to `FetchContent`, therefore the user is required to perform an additional call.

### DIFF
--- a/docs/snippets/fetchContent/CMakeLists.txt
+++ b/docs/snippets/fetchContent/CMakeLists.txt
@@ -17,6 +17,10 @@ FetchContent_Declare(alpaka3 GIT_REPOSITORY https://github.com/alpaka-group/alpa
 # Make alpaka3 available for use in this project
 # This downloads, configures, and makes the library targets available
 FetchContent_MakeAvailable(alpaka3)
+# Finalize the alpaka FetchContent setup
+# this call is currently disabled because it the dev branch first needs the definition alpaka_FetchContent_Finalize()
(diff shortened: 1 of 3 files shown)


docs/snippets/fetchContent/CMakeLists.txt
```diff
@@ -17,6 +17,10 @@ FetchContent_Declare(alpaka3 GIT_REPOSITORY https://github.com/alpaka-group/alpa
 # Make alpaka3 available for use in this project
 # This downloads, configures, and makes the library targets available
 FetchContent_MakeAvailable(alpaka3)
+# Finalize the alpaka FetchContent setup
+# this call is currently disabled because it the dev branch first needs the definition alpaka_FetchContent_Finalize()
+# this call will be enabled after the current PR is merged.
+#alpaka_FetchContent_Finalize()
```
psychocoderHPC, line 23: This line will be uncommented in a second PR, because in this test we always fetch the old dev branch and there the definition is missing.

ikbuibui (approved):

PR #359, opened 2025-11-26
fix files/classes in wrong namespace

psychocoderHPC:
- move iterators from the generic alpaka namespace to `alpaak::onAcc`
- split `Iter.hpp` which contains many class definitions in seperate files
- move internal classes to the internal namespace

This PR is not changing any implementation; it moves only code around and keeps all names, even if some are bad.
The intention is only that `onAcc` classes should not be in the `alpaka/mem` folder.


ikbuibui:
Looks good.
Checked for changes using 
`git diff fix-OnAccIteratorsInGenericAlpakaNamespace^! --color-moved=dimmed-zebra --color-moved-ws=ignore-all-space`

ikbuibui (approved):

PR #362, opened 2025-11-27
fix `atomicDec()` and test all atomic operations

psychocoderHPC:
Moving over the never merged atomic test from alpaka mainline https://github.com/alpaka-group/alpaka/pull/1772

- [x] rebase against #361 and add relaxed to the atomicDec fix

### DIFF
--- a/include/alpaka/api/host/atomic.hpp
+++ b/include/alpaka/api/host/atomic.hpp
@@ -145,11 +145,11 @@ namespace alpaka::onAcc
                 isSupportedByAtomicAtomicRef<T>();
                 detail::atomic_ref<T> ref(*addr);
                 T old = ref;
-                T result = ((old >= value) ? 0 : static_cast<T>(old + 1));
-                while(!ref.compare_exchange_weak(old, result, detail::memory_order_relaxed))
+                T result;
+                do
                 {
-                    result = ((old >= value) ? 0 : static_cast<T>(old + 1));
-                }
+                    result = ((old >= value) ? T{0} : old + T{1});
+                } while(!ref.compare_exchange_weak(old, result, detail::memory_order_relaxed));
                 return old;
             }
         };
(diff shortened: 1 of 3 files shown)


include/alpaka/api/host/atomic.hpp
```diff
@@ -161,10 +161,11 @@ namespace alpaka::onAcc
                 isSupportedByAtomicAtomicRef<T>();
                 detail::atomic_ref<T> ref(*addr);
                 T old = ref;
-                T result = ((old >= value) ? 0 : static_cast<T>(old - 1));
+                T result = (old == static_cast<T>(0) || old > value) ? value : static_cast<T>(old - static_cast<T>(1));
```
psychocoderHPC: the outer cast in the last expression should be removed

include/alpaka/api/host/atomic.hpp
```diff
@@ -163,10 +163,10 @@ namespace alpaka::onAcc
                 isSupportedByAtomicAtomicRef<T>();
                 detail::atomic_ref<T> ref(*addr);
                 T old = ref;
```
ikbuibui, line 165: I like that this immediately benefits from the relaxed atomic ref loads :)

include/alpaka/api/host/atomic.hpp
```diff
@@ -163,10 +163,10 @@ namespace alpaka::onAcc
...
-                T result = ((old >= value) ? 0 : static_cast<T>(old - 1));
+                T result = (old == static_cast<T>(0) || old > value) ? value : (old - static_cast<T>(1));
                 while(!ref.compare_exchange_weak(old, result, detail::memory_order_relaxed))
                 {
-                    result = ((old >= value) ? 0 : static_cast<T>(old - 1));
+                    result = (old == static_cast<T>(0) || old > value) ? value : (old - static_cast<T>(1));
                 }
```
ikbuibui: This can be replaced with a do while, and the casts simply be direct initializations
```suggestion
                T result;
                do {
                    result = (old == T{0} || old > value) ? value : (old - T{1});
                } while (!ref.compare_exchange_weak(old, result, detail::memory_order_relaxed));
```

psychocoderHPC: I applied the suggestion to INc and Dec because in Inc there was a casting issue. This change can also be used for many other functions, but I do not want to sneak it into this PR

PR #363, opened 2025-11-28
cleanup HIP/CUDA atomic function detection

psychocoderHPC:
- remove `CLANG_CUDA_PTX_WORKAROUND` becasue is was only required for CUDA architectures we do not support anymore
- remove alpaka mainline macro usage


ikbuibui (commented):
Looks okay

PR #367, opened 2025-12-03
Add buffer.keepAlive

AntonReinhard:
This adds a `keepAlive` function for `SharedBuffers` that adds the buffer to a queue such that the buffer will stay alive *at least* until that point in the queue is reached.
This is different from the `destructorWaitFor` because it doesn't wait.

This uses a new `enqueueHostFnAsync`, and renames the original `enqueue` for host functions to `enqueueHostFn`.


tests/unit/mem/keepAlive.cpp
```diff
@@ -0,0 +1,90 @@
...
+    onHost::wait(queue);
+
+    // validate that the original buffer got the iota correctly copied
+    auto hostResult = onHost::allocHostLike(originalBuffer);
+
+    onHost::memcpy(queue, hostResult, originalBuffer);
+    onHost::wait(queue);
+
+    bool correct = true;
+    for(auto i = 0; i < N; ++i)
```
psychocoderHPC: warning: comparison of integer expressions of different signedness: ‘int’ and ‘const size_t’ {aka ‘const long unsigned int’} [-Wsign-compare]

psychocoderHPC (changes requested):

tests/unit/mem/keepAlive.cpp
AntonReinhard: Fixed

AntonReinhard:
I'm not sure why the oneapi test is failing... It works for me on hal

docs/snippets/example/05_device.cpp
```diff
@@ -50,6 +50,6 @@ TEST_CASE("host device", "[docs]")
 
     // Getting a queue to enqueue asynchronous work
     onHost::Queue hostQueue = hostDevice.makeQueue();
-    hostQueue.enqueue([]() { std::cout << "Hallo host task" << std::endl; });
+    hostQueue.enqueueHostFn([]() { std::cout << "Hallo host task" << std::endl; });
```
SimeonEhrig, line 53: Is a `enqueueAccFn` planed? Otherwise, maybe it is technically correct but I have no idea what the user should do with the extra information that it is a `HostTask` and not simply a `Task`

include/alpaka/api/unifiedCudaHip/Queue.hpp
```diff
@@ -360,14 +365,16 @@ namespace alpaka::onHost
                 auto data = std::unique_ptr<HostFuncData>(reinterpret_cast<HostFuncData*>(arg));
                 auto& queue = data->q;
                 auto f = queue.m_callBackThread.submit([d = std::move(data)] { d->t(); });
+                // std::cout << "f.wait " << std::this_thread::get_id() << std::endl;
```
SimeonEhrig: Leftover?

include/alpaka/api/unifiedCudaHip/Queue.hpp
```diff
@@ -360,14 +365,16 @@ namespace alpaka::onHost
...
                 f.wait();
+                // std::cout << "f waited" << std::endl;
```
SimeonEhrig: Leftover?

include/alpaka/api/unifiedCudaHip/Queue.hpp
```diff
@@ -360,14 +365,16 @@ namespace alpaka::onHost
...
             }
 
             void operator()(unifiedCudaHip::Queue<T_Device>& queue, T_Task const& task) const
             {
                 ALPAKA_LOG_FUNCTION(onHost::logger::queue);
                 using ApiInterface = typename unifiedCudaHip::Queue<T_Device>::ApiInterface;
-
+                // std::cout << "thread id: " << std::this_thread::get_id() << std::endl;
```
SimeonEhrig: Leftover?

docs/snippets/example/05_device.cpp
AntonReinhard: No, the idea is that the normal `enqueue` is just for kernels, because the semantics are very different between enqueuing a host task and a kernel, so now they get explicit different names instead of just being an overload that looks the same for both.
I'm not sure if we also want a `enqueueEvent`, because that still uses the same `enqueue` currently or if that's fine the way it is.

SimeonEhrig: Ah okay. I didn't saw in the GitHub diff, that `enqueue` still exist. Therefore it is fine.

SimeonEhrig:
> I'm not sure why the oneapi test is failing... It works for me on hal

Did you check Sycl CPU on Hal or only the Intel GPU?

AntonReinhard:
> > I'm not sure why the oneapi test is failing... It works for me on hal
> 
> Did you check Sycl CPU on Hal or only the Intel GPU?

Only the GPU, our hedgedoc still says that the openCL backend is broken on hal

SimeonEhrig:
> > > I'm not sure why the oneapi test is failing... It works for me on hal
> > 
> > 
> > Did you check Sycl CPU on Hal or only the Intel GPU?
> 
> Only the GPU, our hedgedoc still says that the openCL backend is broken on hal

It's on my never ending todo list :grimacing:

psychocoderHPC:
> > > I'm not sure why the oneapi test is failing... It works for me on hal
> > 
> > 
> > Did you check Sycl CPU on Hal or only the Intel GPU?
> 
> Only the GPU, our hedgedoc still says that the openCL backend is broken on hal

```
docker run --device /dev/dri:/dev/dri --gpus all --device=/dev/kfd --device=/dev/dri --security-opt seccomp=unconfined --group-add video -it -v "$HOME/workspace":/workspace intel/oneapi-hpckit:2025.2.0-0-devel-ubuntu24.04
```

note: `-v "$HOME/workspace"` provide access to the local folder 

in the container:

```
apt update
apt-get install -y cmake-curses-gui
export CXX=icpx
```

Than you have access to cpu and intel gpu

AntonReinhard:
I can remove some code duplication still, for now I want to see if the CI worksnow

include/alpaka/api/unifiedCudaHip/Queue.hpp
```diff
@@ -125,6 +125,11 @@ namespace alpaka::onHost
 
             friend struct onHost::internal::Enqueue;
 
+            void enqueueHostFnAsync(auto const& task)
+            {
+                m_callBackThread.submit(task);
```
psychocoderHPC: I assume this should be removed because you spezialized the trait `HostTaskAsync`

psychocoderHPC (changes requested):

include/alpaka/api/unifiedCudaHip/Queue.hpp
AntonReinhard: Right, forgot about that

AntonReinhard: Removed in 429ead9cbe1b2ff8b2c1c08269e2d85bf583c090

AntonReinhard:
The failed job needs to be restarted, but I can't do that. It's some `apt` issue

SimeonEhrig (approved):

include/alpaka/api/host/Queue.hpp
```diff
@@ -187,12 +187,17 @@ namespace alpaka::onHost
...
-            void enqueue(auto const& task)
+            void enqueueHostFn(auto const& task)
             {
                 ALPAKA_LOG_FUNCTION(onHost::logger::queue);
                 submit([task]() { task(); });
             }
 
+            void enqueueHostFnAsync(auto const& task)
+            {
+                m_workerThread.submit(task);
```
psychocoderHPC, line 199: please add above 

```
ALPAKA_LOG_FUNCTION(onHost::logger::queue);
```

include/alpaka/api/syclGeneric/Queue.hpp
```diff
@@ -278,26 +278,49 @@ namespace alpaka::onHost
...
-            [[maybe_unused]] sycl::event ev
-                = queue.m_queue.submit([&queue, task](sycl::handler& cgh)
-                                       { cgh.host_task([&queue, task]() { callHostTask(queue, task); }); });
+            [[maybe_unused]] sycl::event ev = queue.m_queue.submit(
+                [&queue, task](sycl::handler& cgh)
+                {
+                    cgh.host_task(
+                        [&queue, task]()
+                        {
+                            auto f = callHostTask(queue, task);
```
psychocoderHPC: maybe we should skip the indirection via the function, it is a single line only.

```
auto f = queue.m_callBackThread.submit([t = std::move(task)] { t(); });
```

include/alpaka/api/syclGeneric/Queue.hpp
```diff
@@ -278,26 +278,49 @@ namespace alpaka::onHost
...
+    {
+        // same as for Enqueue::HostTask, but not waiting for the task to finish
+        void operator()(syclGeneric::Queue<T_Device>& queue, T_Task const& task) const
+        {
+            ALPAKA_LOG_FUNCTION(onHost::logger::queue);
+            [[maybe_unused]] sycl::event ev = queue.m_queue.submit(
+                [&queue, task](sycl::handler& cgh)
+                {
+                    cgh.host_task([&queue, task]()
+                                  { HostTask<syclGeneric::Queue<T_Device>, T_Task>::callHostTask(queue, task); });
```
psychocoderHPC: same here we should simply call `queue.m_callBackThread.submit([t = std::move(task)] { t(); });`

include/alpaka/api/syclGeneric/Queue.hpp
psychocoderHPC: This also has the advantage that we have no additional copy for Task.

include/alpaka/api/host/Queue.hpp
```diff
@@ -187,12 +187,17 @@ namespace alpaka::onHost
...
              * dependency. In this case the destructure of the queue is not called.
              */
-            void enqueue(auto const& task)
+            void enqueueHostFn(auto const& task)
             {
                 ALPAKA_LOG_FUNCTION(onHost::logger::queue);
                 submit([task]() { task(); });
             }
 
+            void enqueueHostFnAsync(auto const& task)
```
psychocoderHPC: Should we maybe rename `enqueueHostFnAsync` to `enqueueHostFnDeferred`? We had this discussion in the past with `allocAsync()` what we renamed to `allocDeferred`. The reason for the rename was that if you call this method for a blocking queue it is not asynchronous, it blocks the execution stream.
Compared to `enqueueHostFn` only until the task starts and not until it finishes, what is also asynchronous but maybe the word `async`can confuse the user.

psychocoderHPC (changes requested):
Everything with `maybe` is optional and is mostly questions popping up during the review.

psychocoderHPC:
@AntonReinhard I tested the PR with `oneAPi` for AMD, NVIDIA and Intel. The test and scanf imroved works

psychocoderHPC:
observation: Short non scitific benchmark with 256MiB and scan_improved shows that for this thise `keepAlive()` is improving the performance and for the default size 16MiB it slightly reduces the performance, due to the overhead of callback functions.

include/alpaka/api/host/Queue.hpp
AntonReinhard: > The reason for the rename was that if you call this method for a blocking queue it is not asynchronous, it blocks the execution stream.

In the `allocDeferred` case it blocks until the alloc happened (with a blocking queue), but in this case it doesn't wait until the task actually ran, it only waits until the task started, even if the queue is blocking. So I think async is actually correct here unless I'm misunderstanding something

include/alpaka/api/syclGeneric/Queue.hpp
AntonReinhard: Done in 4d653ed3bfee46e8eb2983602ba4165af13e9c15

include/alpaka/api/syclGeneric/Queue.hpp
AntonReinhard: Done in 4d653ed3bfee46e8eb2983602ba4165af13e9c15

include/alpaka/api/host/Queue.hpp
AntonReinhard: Added in 4d653ed3bfee46e8eb2983602ba4165af13e9c15

tests/unit/mem/keepAlive.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+using namespace alpaka;
+
+using TestApis = std::decay_t<decltype(onHost::allBackends(onHost::enabledApis, onHost::example::enabledExecutors))>;
+using IdxType = std::size_t;
+using ValType = int;
+
+struct IotaKernel
+{
+    ALPAKA_FN_ACC void operator()(auto const& acc, concepts::MdSpan<ValType> auto out) const
```
psychocoderHPC: @AntonReinhard I merded 373 so concepts::MdSpan is removed, could you please use concepts::IMdSpan and rebase against dev branch

psychocoderHPC (changes requested):

psychocoderHPC (approved):

SimeonEhrig (approved):

PR #370, opened 2025-12-04
expose warps to the user for iterations via `makeIdxMap`

psychocoderHPC:
- provide `onAcc::range` predefinitions
- provide `onAcc::worker` predefinitions
- add missing warp function: `getWarpIdx()`
- separate interface specializations from the definition to avoid cyclic inlcude dependencies


ikbuibui (approved):
Looks good. Small fixes proposed [here](https://github.com/psychocoderHPC/alpaka3/pull/1)

PR #372, opened 2025-12-05
provided `ffs()` and support for `popcount()` 4 and 8 byte types

psychocoderHPC:
- intrinsic `ffs()`
  - add test
- `popcount()` add support for any 4 and 8 byte type

@LukeTheWalker This is maybe interesting for you.

### DIFF
--- a/include/alpaka/internal/intrinsic.hpp
+++ b/include/alpaka/internal/intrinsic.hpp
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <cstdint>
+
 namespace alpaka::internal::intrinsic
 {
     struct Popcount
@@ -11,7 +13,16 @@ namespace alpaka::internal::intrinsic
         template<typename T_IntrinsicImpl, typename T_Arg>
         struct Op
         {
-            auto operator()(T_IntrinsicImpl const, T_Arg const& val) const;
+            int32_t operator()(T_IntrinsicImpl const, T_Arg const& val) const;
+        };
+    };
+
+    struct Ffs
+    {
+        template<typename T_IntrinsicImpl, typename T_Arg>
+        struct Op
+        {
+            int32_t operator()(T_IntrinsicImpl const, T_Arg const& val) const;
         };
     };
 } // namespace alpaka::internal::intrinsic
--- a/include/alpaka/intrinsic.hpp
+++ b/include/alpaka/intrinsic.hpp
@@ -9,13 +9,28 @@
 #include "alpaka/api/trait.hpp"
 #include "internal/intrinsic.hpp"
 
+#include <climits>
+
 namespace alpaka
 {
-    constexpr auto popcount(auto const& arg)
+    /** Returns the number of bits set to 1. */
+    constexpr int32_t popcount(auto const& arg)
+        requires(sizeof(ALPAKA_TYPEOF(arg)) == 4u || sizeof(ALPAKA_TYPEOF(arg)) == 8u)
     {
-        auto const intrinsicImpl = alpaka::trait::getIntrinsicImpl(thisApi());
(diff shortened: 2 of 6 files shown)


tests/unit/intrinsic/ffs.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <catch2/catch_template_test_macros.hpp>
+
+#include <vector>
+
+using namespace alpaka;
+
+using TestBackends
+    = std::decay_t<decltype(onHost::allBackends(onHost::enabledApis, onHost::example::enabledExecutors))>;
+
+struct PopcountKernel
```
LukeTheWalker: This is probably a typo, should be something like FfsKernel

include/alpaka/internal/intrinsic.hpp
```diff
@@ -11,7 +11,16 @@ namespace alpaka::internal::intrinsic
         template<typename T_IntrinsicImpl, typename T_Arg>
         struct Op
         {
-            auto operator()(T_IntrinsicImpl const, T_Arg const& val) const;
+            int32_t operator()(T_IntrinsicImpl const, T_Arg const& val) const;
```
LukeTheWalker, line 16: why did we move from auto to int32_t?

psychocoderHPC: The standard is returning `int` too so I would like to be explicit. Typically we keep only `auto` if the value can be different for each implementation. This is e.g. the case for the warp function `ballot()` where the return type is 64bit in case the warp size is 64 or 32bit if the warp size is <=32

tests/unit/intrinsic/ffs.cpp
psychocoderHPC: copy past issue, fixed

psychocoderHPC:
I set the PR to draft. I will remove support for types larger 8 byte. It increases the complexity without a strong requirement.
I was today a little bit overmotivated ^^

psychocoderHPC:
@ikbuibui, ready to review, I reduced the PR to the necessary functionality only.

include/alpaka/intrinsic.hpp
```diff
@@ -9,13 +9,28 @@
...
+
 namespace alpaka
 {
-    constexpr auto popcount(auto const& arg)
+    /** Returns the number of bits set to 1. */
+    constexpr int32_t popcount(auto const& arg)
+        requires(sizeof(ALPAKA_TYPEOF(arg)) == 4u || sizeof(ALPAKA_TYPEOF(arg)) == 8u)
     {
-        auto const intrinsicImpl = alpaka::trait::getIntrinsicImpl(thisApi());
+        constexpr auto const intrinsicImpl = alpaka::trait::getIntrinsicImpl(thisApi());
```
ikbuibui:
```suggestion
        constexpr auto intrinsicImpl = alpaka::trait::getIntrinsicImpl(thisApi());
```
It is not a problem but just pointing out that constexpr implies const. This also happens below.

ikbuibui (approved):
Looks good. 
As a comment I don't know if adding support for all types with matching `sizeof` by using `bit_cast` is something useful or if it can be a source of errors since the users may not expect it, but it is fine for me.

include/alpaka/intrinsic.hpp
psychocoderHPC: fixed

psychocoderHPC:
> Looks good. As a comment I don't know if adding support for all types with matching `sizeof` by using `bit_cast` is something useful or if it can be a source of errors since the users may not expect it, but it is fine for me.

I also thought about it, the STL supports only unsigned types, in alpaak mainline we already supported integral types and casted internally, therefore I saw no reason to relax it. In principle, it is a bit function and there is no reason not to allow any type.

PR #375, opened 2025-12-08
add intrinsic `clz()`

psychocoderHPC:
- add `clz()`
- fix wrong datatype for CUDA/HIP `ffs()`
- fix wrong license for intrinsic tests

### DIFF
--- a/include/alpaka/api/unifiedCudaHip/intrinsic.hpp
+++ b/include/alpaka/api/unifiedCudaHip/intrinsic.hpp
@@ -43,11 +43,31 @@ namespace alpaka::internal::intrinsic
         {
             if constexpr(sizeof(T_Arg) == 4u)
             {
-                return __ffs(std::bit_cast<unsigned int>(val));
+                return __ffs(std::bit_cast<int>(val));
             }
             else if constexpr(sizeof(T_Arg) == 8u)
             {
-                return __ffsll(std::bit_cast<unsigned long long>(val));
+                return __ffsll(std::bit_cast<long long int>(val));
+            }
+            else
+                static_assert(!sizeof(T_Arg), "Unsupported data type, sizeof() must be 4 or 8");
+
+            ALPAKA_UNREACHABLE(int{});
+        }
+    };
+
+    template<typename T_Arg>
+    struct Clz::Op<alpaka::internal::CudaHipIntrinsic, T_Arg>
+    {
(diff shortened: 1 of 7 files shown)


include/alpaka/api/unifiedCudaHip/intrinsic.hpp
```diff
@@ -43,11 +43,31 @@ namespace alpaka::internal::intrinsic
         {
             if constexpr(sizeof(T_Arg) == 4u)
             {
-                return __ffs(std::bit_cast<unsigned int>(val));
```
ikbuibui, line 46: I am confused. I see in the docs that CUDA/HIP use signed values as the argument, but why was this not failing to compile?

tests/unit/intrinsic/clz.cpp
```diff
@@ -0,0 +1,83 @@
...
+    alpaka::onHost::memcpy(queue, hostOutput, devOutput);
+
+    // Wait for the queue to finish
+    alpaka::onHost::wait(queue);
+
+    // Verification
+    for(size_t i = 0; i < size; ++i)
+    {
+        auto val = hostInput[i];
+        int expected = std::countl_zero(val);
```
ikbuibui, line 80:
```
template<typename TValue>
static int32_t naiveClz(TValue value)
{
    if(value == 0)
        return sizeof(TValue) * CHAR_BIT;

    std::int32_t result = 0;
    const std::int32_t bits = sizeof(TValue) * CHAR_BIT;
    TValue mask = TValue(1) << (bits - 1);

    while((value & mask) == 0)
    {
        result++;
        mask >>= 1;
    }
    return result;
}
```
Compares `std::countl_zero` with itself. I don't mind it if you see the STL as a source of truth. 
Alternatively if you want, this naive implementation is similar to the one used with `ffs`.

ikbuibui (approved):
Looks okay

include/alpaka/api/unifiedCudaHip/intrinsic.hpp
psychocoderHPC: because of C++ implicit conversion

tests/unit/intrinsic/clz.cpp
psychocoderHPC: If there is a native STL function I take it and trust that this is tested. Only if there is jo native function or as for ffl you need to do some tricks I prefere a naive implementation

PR #377, opened 2025-12-10
fix usage of alpaka via `add_subdirectory()`

psychocoderHPC:
The same issue we saw when using alpaka with CMake's fetch content is present when using alpaka via `add_subdirectory()`.

- rename `alpakaFetchContentFinalize.cmake` to `cmake/alpakaPrepareForAddSubdirectoryUsage.cmake`
- add handling of optional load of `cmake/alpakaPrepareForAddSubdirectoryUsage.cmake` to `alpaka_finalize()`
- change `alpaka_finalize()` to a macro
- fix tiny message in the cuda cmake file

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ cmake_minimum_required(VERSION 3.25)
 project(alpaka)
 
 # This file's directory.
-set(_alpaka_ROOT_DIR ${CMAKE_CURRENT_LIST_DIR})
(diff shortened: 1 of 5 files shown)


psychocoderHPC:
I tested using alpaka with `add_subdirectory()`, via installation with `find_package()` and the fetch content example.

SimeonEhrig (approved):

SimeonEhrig:
I also tested `find_package`, `add_subdirectory` and `fetch_content`.

PR #381, opened 2025-12-10
fix CMake helper `alpaka_get_targets_recursive()`

psychocoderHPC:
If an external library foo is linking `alpaka` which means that alpaka is a transitive CMake dependency for the project which is using foo we have not detect that `alpaak_finalize()` was called on the project target.
This PR fixing the issue by recursivly scanning all target.

This fix is required for @TimHanel00 's project alpakaTune

### DIFF
--- a/cmake/finalize.cmake
+++ b/cmake/finalize.cmake
@@ -5,7 +5,7 @@
 
 ### Provide the alpaka target names linked to a target
 ##
-## The target alpaka::alpaka and alpaka will insecurely be resolved to the actual targets linked to it.
+## All targets will recursively be resolved to the actual targets linked to it.
 ## target names are: ALPAKA, CUDA, HIP, ONEAPI, HEADERS, HOST
 ##
 ## The output will be appended to the list variable provided as alpaka_target_list.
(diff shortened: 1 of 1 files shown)


SimeonEhrig:
@psychocoderHPC If this is working in @TimHanel00 project, I will approve. 

But to be honest, we should thinking about CMake integration tests. I think it make sense to implement it soon as possible. Each time, if you fix a bug in CMake, we need to test the CMake 3 times (`find_package`, `add_subdirectory` and `fetch_content`) to be sure that nothing breaks because the changes are not trivial.

psychocoderHPC:
> @psychocoderHPC If this is working in @TimHanel00 project, I will approve.

yes it works

> But to be honest, we should thinking about CMake integration tests. I think it make sense to implement it soon as possible. Each time, if you fix a bug in CMake, we need to test the CMake 3 times (find_package, add_subdirectory and fetch_content) to be sure that nothing breaks because the changes are not trivial.

Agree but would like to focus on documentation first because this is required for the upcoming workshop.
I only fix all the things that @TimHanel00 can go on with his task.

SimeonEhrig (approved):

PR #384, opened 2025-12-11
Add scan algorithms

AntonReinhard:
This adds the scan algorithm that was previously implemented as an example to the `alpaka::onHost` namespace.

The interface consists of 
- `getScanBufferSize` -> for an extent of a vector to scan, get the buffer size necessary (may be 0 for small scans)
- `exclusiveScan` -> perform an exclusive scan, with or without specifically passed intermediate buffer
- `inclusiveScan` -> perform an inclusive scan, with or without specifically passed intermediate buffer
- `exclusiveScanInPlace` -> perform an exclusive scan, with or without specifically passed intermediate buffer, only taking one buffer and using it as input+output, overwriting
- `inclusiveScanInPlace` -> perform an inclusive scan, with or without specifically passed intermediate buffer, only taking one buffer and using it as input+output, overwriting

The inplace versions need a separate overload because the optional intermediate buffer would otherwise make the functions ambiguous, but I think it's a good idea anyway to have more explicit names.

I added one relatively large test to have a case where the recursion needs two iterations, but it might be too large/take too long in CI.

### DIFF
--- a/include/alpaka/onHost/algo/internal/scan.hpp
+++ b/include/alpaka/onHost/algo/internal/scan.hpp
@@ -33,6 +33,7 @@ namespace alpaka::onHost::internal
     constexpr std::size_t numNvidiaBanks = 32u;
     constexpr std::size_t numAmdBanks = 32u;
     constexpr std::size_t numIntelBanks = 16u;
+    constexpr std::size_t chunkSize = 2048u;
 
     template<alpaka::concepts::DeviceKind TDeviceKind, typename T_Idx, typename T_Data>
     consteval T_Idx maximumMiniBlockSize()
@@ -52,7 +53,7 @@ namespace alpaka::onHost::internal
      * unknown/unimplemented device kinds, infinite memory banks are assumed, i.e., no padding is used.
      */
     template<typename TDeviceKind, typename T_Idx>
-    constexpr T_Idx conflictFreeAccess(auto const& n)
+    constexpr T_Idx conflictFreeAccess(T_Idx const& n)
     {
         if constexpr(TDeviceKind{} == deviceKind::nvidiaGpu)
             return n + n / static_cast<T_Idx>(numNvidiaBanks);
@@ -124,7 +125,7 @@ namespace alpaka::onHost::internal
     {
     public:
         ALPAKA_FN_ACC void operator()(
-            alpaka::onHost::concepts::Device auto const& acc,
+            auto const& acc,
             alpaka::concepts::IDataSource auto const& inputVec,
             alpaka::concepts::IMdSpan auto outputVec,
             auto... blockSums) const
@@ -145,21 +146,6 @@ namespace alpaka::onHost::internal
             constexpr auto LocalArrayLength = miniBlocksPerThread * miniBlockSize;
             using LocalArray = T_Data[LocalArrayLength];
 
-            ALPAKA_LOG_INFO(
-                onHost::logger::memory,
-                [&]()
-                {
-                    std::stringstream ss;
-                    ss << "scan kernel: {";
-                    ss << "frameExtent: " << frameExtent;
-                    ss << ", numThreadsPerBlock: " << numThreadsPerBlock;
-                    ss << ", elsPerThread: " << elsPerThread;
-                    ss << ", chunkExtent: " << chunkExtent;
-                    ss << ", miniBlockSize: " << miniBlockSize;
-                    ss << ", miniBlocksPerThread: " << miniBlocksPerThread;
-                    ss << "}" return ss.str();
-                });
-
             auto const validElementsInLastFrame = (numElements - T_Idx{1}) % chunkExtent + T_Idx{1};
 
             /* This kernel is called with 1-dimensional frame extents.
@@ -178,7 +164,7 @@ namespace alpaka::onHost::internal
 
                 auto tmp = onAcc::declareSharedMdArray<T_Data, uniqueId()>(
                     acc,
-                    CVec<T_Idx, conflictFreeAccess<DeviceType>(miniBlocksPerChunk - T_Idx{1}) + T_Idx{1}>{});
+                    CVec<T_Idx, conflictFreeAccess<DeviceType, T_Idx>(miniBlocksPerChunk - T_Idx{1}) + T_Idx{1}>{});
                 auto const frameOffset = chunkExtent * frameIdx;
 
                 for(auto frameElem : onAcc::makeIdxMap(
@@ -220,7 +206,8 @@ namespace alpaka::onHost::internal
                         miniBlockOffset += miniBlockSize)
                     {
                         // scan miniblock
-                        auto miniBlockSum = scanMiniBlock(regMem + miniBlockOffset, CVec<T_Idx, miniBlockSize>{});
+                        auto miniBlockSum
+                            = scanMiniBlock<T_Idx, T_Data>(regMem + miniBlockOffset, CVec<T_Idx, miniBlockSize>{});
 
                         // write miniblock sum into shared memory
                         tmp[conflictFreeAccess<DeviceType>((frameElem + miniBlockOffset) / miniBlockSize)]
@@ -355,7 +342,7 @@ namespace alpaka::onHost::internal
     {
     public:
         ALPAKA_FN_ACC void operator()(
-            alpaka::onHost::concepts::Device auto const& acc,
+            auto const& acc,
             alpaka::concepts::IMdSpan auto const& blockSums,
             alpaka::concepts::IMdSpan auto outputVec) const
         {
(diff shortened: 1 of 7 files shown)


AntonReinhard:
The oneapi execution seems inconsistent and sometimes segfaults, but I'm not yet sure if this is a problem with the scan implementation or with the recent `keepAlive`

include/alpaka/onHost/algo/internal/scan.hpp
```diff
@@ -0,0 +1,462 @@
...
+
+namespace alpaka::onHost::internal
+{
+    enum ScanType
+    {
+        EXCLUSIVE_SCAN,
+        INCLUSIVE_SCAN
+    };
+
+    constexpr std::size_t numNvidiaBanks = 32u;
```
psychocoderHPC: the warp size is now available at compile time and should be used here. see warp uni tests

tests/unit/algorithm/scan.cpp
```diff
@@ -0,0 +1,215 @@
...
+using namespace alpaka;
+
+using TestBackends
+    = std::decay_t<decltype(onHost::allBackends(onHost::enabledApis, onHost::example::enabledExecutors))>;
+
+template<typename T_Data>
+int validateResult(
+    auto queue,
+    concepts::IDataSource<T_Data> auto& inputData,
+    auto const& bufY,
```
SimeonEhrig:
```suggestion
    concepts::IDataSource<T_Data> auto const& bufY,
```

tests/unit/algorithm/scan.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+using namespace alpaka;
+
+using TestBackends
+    = std::decay_t<decltype(onHost::allBackends(onHost::enabledApis, onHost::example::enabledExecutors))>;
+
+template<typename T_Data>
+int validateResult(
+    auto queue,
+    concepts::IDataSource<T_Data> auto& inputData,
```
SimeonEhrig:
```suggestion
    concepts::IDataSource<T_Data> auto const&  inputData,
```

include/alpaka/onHost/algo/scan.hpp
```diff
@@ -0,0 +1,162 @@
...
+     */
+    template<typename T_Data>
+    void inclusiveScan(
+        alpaka::concepts::Executor auto& exec,
+        alpaka::onHost::concepts::Device auto& devAcc,
+        auto const& queue,
+        alpaka::concepts::IDataSource<T_Data> auto const& inputVec,
+        alpaka::concepts::IMdSpan<T_Data> auto outputVec,
+        auto... buffer)
+    {
```
SimeonEhrig:
```suggestion
    void inclusiveScan(
        alpaka::concepts::Executor auto& exec,
        alpaka::onHost::concepts::Device auto& devAcc,
        auto const& queue,
        alpaka::concepts::IDataSource auto const& inputVec,
        alpaka::concepts::IMdSpan auto outputVec,
        auto... buffer)
        requires std::same_as<std::remove_cv<ALPAKA_TYPE_OF(inputVec)::value_type>, std::remove_cv<ALPAKA_TYPE_OF(outputVec)::value_type>>
    {
```

I understand, that you want to have the `T_Data` type to make the concepts more explicit. But this requires that you need to set the data type manually for each function, because we cannot deduce it in the function signature. I think it is better to check if buffer `value_types` are equal via required clause.

By the way, I was also thinking about to add a `raw_type` to `IDataSource` that you can easily compare the value type, without the need to remove a possible `const`. @psychocoderHPC What are you thinking about?

include/alpaka/onHost/algo/scan.hpp
```diff
@@ -0,0 +1,162 @@
...
+     * buffer needs to be. If omitted, it will be allocated and destructed on the fly.
+     */
+    template<typename T_Data>
+    void inclusiveScan(
+        alpaka::concepts::Executor auto& exec,
+        alpaka::onHost::concepts::Device auto& devAcc,
+        auto const& queue,
+        alpaka::concepts::IDataSource<T_Data> auto const& inputVec,
+        alpaka::concepts::IMdSpan<T_Data> auto outputVec,
+        auto... buffer)
```
SimeonEhrig: If use the variadic argument `auto... buffer` only for a single optional buffer, please a overload instead. This is easier to understand.

include/alpaka/onHost/algo/scan.hpp
```diff
@@ -0,0 +1,162 @@
...
+     * @param queue The queue to enqueue to.
+     * @param inputVec The input data. Can be const.
+     * @param outputVec The output data. To perform an in-place scan, use the overload with only one data object.
+     * @param buffer (optional) The internally used buffer. Use the scanBufferSize() function to check how big the
+     * buffer needs to be. If omitted, it will be allocated and destructed on the fly.
+     */
+    template<typename T_Data>
+    void inclusiveScan(
+        alpaka::concepts::Executor auto& exec,
+        alpaka::onHost::concepts::Device auto& devAcc,
```
SimeonEhrig: You can extract the device from the queue (`queue.getDevice()`). But it also looks like you never uses the device.

SimeonEhrig (commented):
I checked only the interfaces yet. I will check the algorithm itself in a second review. But I'm not sure, if I will finish this today. Therefore I split the review.

include/alpaka/onHost/algo/scan.hpp
AntonReinhard: True, I currently never use the device, but I wanted to have it in the interface because we might want to call vendor library code or similar later when available, so I wanted to prepare the interface for that.

SimeonEhrig: The question is, is there a use case, where we use a different device, than the device which is bound the to the queue?

psychocoderHPC (changes requested):

include/alpaka/onHost/algo/scan.hpp
AntonReinhard: I talked with Rene, I will use just the queue for now and we will worry about vendor implementations some other time anyway.

include/alpaka/onHost/algo/internal/scan.hpp
```diff
@@ -0,0 +1,475 @@
...
+    {
+        if constexpr(TDeviceKind{} == deviceKind::nvidiaGpu)
+            return static_cast<T_Idx>(8);
+        else if constexpr(TDeviceKind{} == deviceKind::amdGpu)
+            return static_cast<T_Idx>(8);
+        else if constexpr(TDeviceKind{} == deviceKind::intelGpu)
+            return static_cast<T_Idx>(8);
+        else
+            return static_cast<T_Idx>(32768) / sizeof(T_Data);
+    }
```
SimeonEhrig, line 46: Just a question, why is the block size for each GPU 8?

SimeonEhrig:
I had a lock in the actual algorithm. Looks good but I cannot check all the details because it takes to much time to introduce myself. So short question, how do you verify, that the optimizations works? I assume temporary code to measure the execution time in the test code. If found an `#include <chrono>` ;-)

include/alpaka/onHost/algo/internal/scan.hpp
AntonReinhard: These are miniblock sizes, i.e. "tiny" serial scans that each thread does. On GPUs you don't want to do too many to not make the kernel too large, on a CPU you want to do very large serial scans.

include/alpaka/onHost/algo/scan.hpp
AntonReinhard: Done.

include/alpaka/onHost/algo/scan.hpp
AntonReinhard: I added static_asserts in the internal implementation

include/alpaka/onHost/algo/internal/scan.hpp
AntonReinhard: Done in 413ff47affd73b8117ab52dba0a42c3ca464bfb8

AntonReinhard:
I've opened an issue about the keepAlive + allocDeferred problem: https://github.com/alpaka-group/alpaka3/issues/394
For now I have reverted to using normal `alloc` with `keepAlive` in the scan test, which seems to work consistently. Using the deferred version creates deadlocks on oneAPI.

SimeonEhrig (approved):

psychocoderHPC:
> I had a lock in the actual algorithm. Looks good but I cannot check all the details because it takes to much time to introduce myself. So short question, how do you verify, that the optimizations works? I assume temporary code to measure the execution time in the test code. If found an `#include <chrono>` ;-)

@AntonReinhard did the performance validation in the past with the example and chrono. This PR is shaping an interface around the improved scan from the example.

@AntonReinhard After this PR is merged, we should update the example and use this simple interface in the improved scan example. I think we should not repeat the code in examples we use internally. If someone is interested, the code is available.

psychocoderHPC (approved):

psychocoderHPC (approved):

PR #393, opened 2025-12-16
CI: add cuda 13.1.0

psychocoderHPC:


psychocoderHPC:
CUDA 13.1.0 is boken and can not handle C++20 concpets correctly:~~ https://godbolt.org/z/vGafaa9Gc ~~

I will set this PR to draft. 
I opened a bug report for NVIDIA: ID 5748816

psychocoderHPC:
The examples was wrong here are the right reproducer:
Cuda: https://godbolt.org/z/qTT69rxn9
clang/gcc: https://godbolt.org/z/7c8xe3eWz

SimeonEhrig:
done in PR #470

PR #395, opened 2025-12-17
refactor `enabledExecutors`

psychocoderHPC:
- deprecate `alpaka::onHost::example::enabledExecutors`
- the new variable will be  `alpaka::cmake::enabledExecutors`
  - the reason is that the name space `onHost` was wrong
  - the namespace `example` was misleading too

I was very long struggling with myself to expose this variable to the user, therefore it was never included from `alpaka.hpp`.
IMO with the new namespace it is clearer how to control what's enabled.

issue #396: takes care that we will remove `alpaka::onHost::example::enabledExecutors` before the release.
I deprecated the variable to avoid invalidating the current one, and I may have used Godbolt examples.

### DIFF
--- a/include/alpaka/executor.hpp
+++ b/include/alpaka/executor.hpp
@@ -8,6 +8,7 @@
 #include "alpaka/api/hip/executor.hpp"
 #include "alpaka/api/host/executor.hpp"
 #include "alpaka/api/oneApi/executor.hpp"
+#include "alpaka/core/PP.hpp"
 
 namespace alpaka::exec
 {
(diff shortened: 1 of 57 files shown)


SimeonEhrig:
Two questions:
1. The `alpaka::cmake` namespace is also available if no CMake is used?
2. Enabling and disabling the executors is actual only a feature, which we use for our examples or is it also designed to be used in production code?

In the case, the function is in general available, maybe we can use the `alpaka::exec` namespace. So `alpaka::exec::enabledExecutors`.

psychocoderHPC:
> Two questions:
> 
> 1. The `alpaka::cmake` namespace is also available if no CMake is used?

Yes :-( The doc string mentioned that there are two ways to control this variable. Via CMake or via defines.
Do you have maybe a better name?

> 2. Enabling and disabling the executors is actual only a feature, which we use for our examples or is it also designed to be used in production code?
> 
> In the case, the function is in general available, maybe we can use the `alpaka::exec` namespace. So `alpaka::exec::enabledExecutors`.

With this change, it became always usable but not mandatory.
Should I move it it `alpaka::exec` this sounds betetr than `alpaka::cmake`

SimeonEhrig:
> > Two questions:
> > 
> > 1. The `alpaka::cmake` namespace is also available if no CMake is used?
> 
> Yes :-( The doc string mentioned that there are two ways to control this variable. Via CMake or via defines. Do you have maybe a better name?
> 
> > 2. Enabling and disabling the executors is actual only a feature, which we use for our examples or is it also designed to be used in production code?
> > 
> > In the case, the function is in general available, maybe we can use the `alpaka::exec` namespace. So `alpaka::exec::enabledExecutors`.
> 
> With this change, it became always usable but not mandatory. Should I move it it `alpaka::exec` this sounds betetr than `alpaka::cmake`

If the feature is available in general, we should move it to the `alpaka::exec` namespace.

include/alpaka/executor.hpp
```diff
@@ -19,3 +20,39 @@ namespace alpaka::exec
      */
     constexpr auto allExecutors = std::make_tuple(gpuCuda, gpuHip, oneApi, cpuOmpBlocks, cpuTbbBlocks, cpuSerial);
 } // namespace alpaka::exec
+
+namespace alpaka::cmake
```
SimeonEhrig:
```suggestion
namespace alpaka::exec
```

SimeonEhrig (changes requested):

include/alpaka/executor.hpp
```diff
@@ -18,4 +19,37 @@ namespace alpaka::exec
      * executor is used.
      */
     constexpr auto allExecutors = std::make_tuple(gpuCuda, gpuHip, oneApi, cpuOmpBlocks, cpuTbbBlocks, cpuSerial);
+
+    /** list of enabled executors
+     *
+     * - executors can be dis/en-abled by the CMake define alpaka_EXEC_<ExecutorName>
```
AntonReinhard:
```suggestion
     * - executors can be dis-/enabled by the CMake define alpaka_EXEC_<ExecutorName>
```

AntonReinhard (approved):

SimeonEhrig (approved):

include/alpaka/executor.hpp
psychocoderHPC: done

PR #400, opened 2025-12-17
SIMD where expression

psychocoderHPC:
- Simd: fix `SimdOrScalar`, missed support for floating point numbers
- add concept `SimdMask`
- implement `SimdWhereExpr` which can be created with `alpaka::where()`

Currently, only the non-const where is implemented, `std::simd` defines a constant where expression too, which can be implemented later.
The `SimdMask` is currently a `Simd<bool>` most likely a bitmask can be beneficial.

### DIFF
--- a/include/alpaka/SimdWhereExpr.hpp
+++ b/include/alpaka/SimdWhereExpr.hpp
@@ -1,0 +1,79 @@
+/* Copyright 2025 René Widera
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+/** @file This file provides a basic implementation of a SIMD vector.
+ *
+ * The implementation is based on the class Vec:
+ *   - the storge policy should become the native SIMD implementation e.g. std::simd
+ *   - load/ store and simd specifis should be implemented in the storage policy
+ *   - the name of storage policy should be changed
+ *
+ *   The current operator operations relay on compilers auto vectorization.
+ */
+
+#pragma once
+
+#include "alpaka/Simd.hpp"
+
+namespace alpaka
+{
+    template<concepts::SimdMask Mask, concepts::Simd T_Simd>
+    struct SimdWhereExpr
+    {
+        Mask const& mask;
+        T_Simd& value;
+
+        constexpr SimdWhereExpr(Mask const& m, T_Simd& v) : mask(m), value(v)
+        {
+        }
+
+        // disable copy and move constructors/operators to avoid pointing to invalid references.
+        constexpr SimdWhereExpr(SimdWhereExpr const&) = delete;
+        constexpr SimdWhereExpr(SimdWhereExpr&&) = delete;
+        constexpr SimdWhereExpr& operator=(SimdWhereExpr const&) = default;
+        constexpr SimdWhereExpr& operator=(SimdWhereExpr&&) = default;
+
+        using value_type = typename T_Simd::type;
+
+        constexpr void operator=(concepts::Simd auto const& rhs)
+        {
+            value.update(mask, rhs);
+        }
+
+        constexpr void operator=(concepts::LosslesslyConvertible<value_type> auto const& rhs)
+        {
+            value.update(mask, rhs);
+        }
+
+#define ALPAKA_SIMD_EXPR_ASSIGN_OP(op_name, op)                                                                       \
+    constexpr void operator op_name(concepts::Simd auto const& rhs)                                                   \
+    {                                                                                                                 \
+        value.update(mask, value op rhs);                                                                             \
+    }                                                                                                                 \
+    constexpr void operator op_name(concepts::LosslesslyConvertible<value_type> auto const& rhs)                      \
+    {                                                                                                                 \
+        value.update(mask, value op rhs);                                                                             \
+    }
+
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(+=, +)
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(-=, -)
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(/=, -)
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(*=, *)
+
+
+#undef ALPAKA_SIMD_EXPR_ASSIGN_OP
+    };
+
+    /** Conditional editional update
+     *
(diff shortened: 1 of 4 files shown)


tests/unit/simd/where.cpp
```diff
@@ -0,0 +1,53 @@
+/* Copyright 2024 René Widera
```
AntonReinhard: 😉 
```suggestion
/* Copyright 2025 René Widera
```

include/alpaka/SimdWhereExpr.hpp
```diff
@@ -0,0 +1,79 @@
...
+        value.update(mask, value op rhs);                                                                             \
+    }
+
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(+=, +)
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(-=, -)
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(/=, -)
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(*=, *)
+
+
+#undef ALPAKA_SIMD_EXPR_ASSIGN_OP
```
AntonReinhard, line 65: This formatting is pretty weird

include/alpaka/SimdWhereExpr.hpp
```diff
@@ -0,0 +1,79 @@
+/* Copyright 2025 René Widera
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+/** @file This file provides a basic implementation of a SIMD vector.
+ *
+ * The implementation is based on the class Vec:
+ *   - the storge policy should become the native SIMD implementation e.g. std::simd
+ *   - load/ store and simd specifis should be implemented in the storage policy
```
AntonReinhard:
```suggestion
 *   - load/ store and simd specifics should be implemented in the storage policy
```

include/alpaka/SimdWhereExpr.hpp
```diff
@@ -0,0 +1,79 @@
...
+ *   - the name of storage policy should be changed
+ *
+ *   The current operator operations relay on compilers auto vectorization.
```
AntonReinhard:
```suggestion
 *   The current operator operations rely on compilers auto vectorization.
```

include/alpaka/SimdWhereExpr.hpp
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+#pragma once
+
+#include "alpaka/Simd.hpp"
+
+namespace alpaka
+{
+    template<concepts::SimdMask Mask, concepts::Simd T_Simd>
+    struct SimdWhereExpr
```
AntonReinhard, line 22: This is only really supposed to be constructed through the `alpaka::where()`  function, right? Maybe this should be in an internal namespace, especially when it doesn't have a doc string

include/alpaka/SimdWhereExpr.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+
+#undef ALPAKA_SIMD_EXPR_ASSIGN_OP
+    };
+
+    /** Conditional editional update
+     *
+     * @param mask SIMD pack where each component is true for the element in v which should be overwritten with the
+     * value assigned to the returned expression
+     * @param value value on which the mask is applied on
```
AntonReinhard:
```suggestion
     * @param mask SIMD pack of booleans, where each component is true for the element in v which should be overwritten with the
     * value assigned to the returned expression
     * @param value SIMD vector to which the mask is applied
```

include/alpaka/SimdWhereExpr.hpp
```diff
@@ -0,0 +1,79 @@
...
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(+=, +)
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(-=, -)
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(/=, -)
+        ALPAKA_SIMD_EXPR_ASSIGN_OP(*=, *)
+
+
+#undef ALPAKA_SIMD_EXPR_ASSIGN_OP
+    };
+
+    /** Conditional editional update
```
AntonReinhard: This definitely needs more explanation and probably an example like in the tests

AntonReinhard (changes requested):

include/alpaka/SimdWhereExpr.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+        constexpr SimdWhereExpr(Mask const& m, T_Simd& v) : mask(m), value(v)
+        {
+        }
+
+        // disable copy and move constructors/operators to avoid pointing to invalid references.
+        constexpr SimdWhereExpr(SimdWhereExpr const&) = delete;
+        constexpr SimdWhereExpr(SimdWhereExpr&&) = delete;
+        constexpr SimdWhereExpr& operator=(SimdWhereExpr const&) = default;
+        constexpr SimdWhereExpr& operator=(SimdWhereExpr&&) = default;
```
SimeonEhrig: Only the constructors are disable. The copy and move assignment are enable. Is the code or comment wrong?

include/alpaka/SimdWhereExpr.hpp
psychocoderHPC: no it is correct, it is part of the class but the macro `ALPAKA_SIMD_EXPR_ASSIGN_OP` has no relation to the class.
I could move the macro out of the class body but I like to keep the scope visible after the usage I delete the definition to avoid possible side effects within the code.

tests/unit/simd/where.cpp
psychocoderHPC: time is relative

include/alpaka/SimdWhereExpr.hpp
psychocoderHPC: Thats a bug, the code is wrong the comment is correct.

include/alpaka/SimdWhereExpr.hpp
psychocoderHPC: I kept it in the user space because the user is actively using the interface of this object.

psychocoderHPC:
I add the review comments and wait for the CI not to fail with download issues

AntonReinhard (approved):

PR #403, opened 2025-12-18
IMdSpan: relaxed return value requirements of the access operator

### LINKED ISSUE #402: IMdSpan doesn't work with std::atomic<DataType>
As the title says, the IMdSpan concept doesn't accept an MdSpan of `std::atomic<DataType>`, because the `AssignableTo` statement fails. See this [godbolt link](https://godbolt.org/z/TaKzo61ah).

The allocation works as intended, and the assigning works as well, for `DataTypes`, just not for `std::atomic<DataType>` which is what the concept tries to check against. I'm not sure how to go about fixing this.

SimeonEhrig:
- before the return value of the access operator of a IMdSpan object needed to be copy able
- now also non-copyable objects like std::atomic are allowed
- bad side effect, the IDataSource concept is harder to read now

fix: #402


SimeonEhrig:
@AntonReinhard Can you please check if this fixes your problem?

AntonReinhard (approved):
Yeah this seems to work 👍

psychocoderHPC (approved):

PR #410, opened 2026-01-07
ManagedDealloc fixes

ikbuibui:
Removes unnecessary mutex lock from destruction of `ManagedDealloc`. 
Fixes a typo in the `ManagedDealloc` filename.

Sneaks in a change of removing an unused function `demangledNameShort`, which was calling a non existent function `simplifyDemangedName`.


include/alpaka/onHost/demangledName.hpp
```diff
@@ -79,13 +79,4 @@ namespace alpaka::onHost
         return simplified;
     }
 
-    /** Get a simplified demangled name of an object
-     *
-     * Template parameters will be left out and the alpaka namespace will be removed.
-     */
-    inline std::string demangledNameShort(auto const& demangledName)
```
psychocoderHPC, line 86: The function is above but looks like it was renamed at some point.
Never the less I will keep this in because we do not use this function somewhere.

psychocoderHPC (approved):

PR #412, opened 2026-01-08
fix warning in scan algorithm

psychocoderHPC:
If we not compile with debug symbols the scan implementation is showing a warning due to an unsued type definition.


ikbuibui (approved):
Typename everywhere!

PR #418, opened 2026-01-14
add device property: `globalMemCapacityBytes`

psychocoderHPC:
note: The possibility to query the amount of free memory will be provided in a separate PR.

### DIFF
--- a/include/alpaka/api/host/sysInfo.hpp
+++ b/include/alpaka/api/host/sysInfo.hpp
@@ -144,7 +144,7 @@ namespace alpaka::onHost
     //! \return The total number of bytes of global memory.
     //! Adapted from David Robert Nadeau:
     //! http://nadeausoftware.com/articles/2012/09/c_c_tip_how_get_physical_memory_size_system
-    inline auto getTotalGlobalMemSizeBytes() -> std::size_t
+    inline auto getGlobalMemCapacityBytes() -> std::size_t
     {
 #if ALPAKA_OS_WINDOWS
         MEMORYSTATUSEX status;
@@ -174,7 +174,7 @@ namespace alpaka::onHost
         std::uint64_t size(0);
         std::size_t sizeLen{sizeof(size)};
         if(sysctl(mib, 2, &size, &sizeLen, nullptr, 0) < 0)
-            throw std::logic_error("getTotalGlobalMemSizeBytes failed calling sysctl!");
+            throw std::logic_error("getGlobalMemCapacityBytes failed calling sysctl!");
         return static_cast<std::size_t>(size);
 
 #    elif defined(_SC_AIX_REALMEM) // AIX.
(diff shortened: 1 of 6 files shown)


include/alpaka/api/host/sysInfo.hpp
```diff
@@ -196,12 +196,12 @@ namespace alpaka::onHost
...
         if(sysctl(mib, 2, &size, &sizeLen, nullptr, 0) < 0)
-            throw std::logic_error("getTotalGlobalMemSizeBytes failed calling sysctl!");
+            throw std::logic_error("getGlobalMemCapacityBytes failed calling sysctl!");
         return static_cast<std::size_t>(size);
 #    endif
 
 #else
-#    error "getTotalGlobalMemSizeBytes not implemented for this system!"
+#    error "getGlobalMemCapacityBytes not implemented for this system!"
 #endif
```
AntonReinhard, line 205: Doesn't this mean that I can't compile alpaka at all on such a system, even if I don't want to use this feature? Wouldn't it make more sense to throw here or return zero and log a warning or something like that instead?

include/alpaka/onHost/DeviceProperties.hpp
```diff
@@ -11,16 +11,32 @@
...
         auto getName() const
         {
             return name;
         }
 
+        /** The total amount of global device memory in bytes.
+         *
+         * @attention It is **not** the amount of free memory!
+         */
+        size_t globalMemCapacityBytes;
```
AntonReinhard, line 33: What exactly is the memory of the device? Does it include all memory accessible on the device or just memory that is physically on that device? For example, the memory for Metal GPUs on Apple can be completely shared with the CPU, would that mean this is zero?

The answer should be documented here I think.

include/alpaka/api/host/sysInfo.hpp
psychocoderHPC: It is taken from mainline alpaka. 
Yes if we do not support the system you can not compile for it.
Thats fine becaue as soon as you create a CPU device what is done in case you allocate host memory you call this method and this will enforce that it is implemented.

include/alpaka/onHost/DeviceProperties.hpp
psychocoderHPC: I will add that it is the amount of physical memory.
In your metal example I would expect that an API is returning the amount of host memory even for the CPU.

AntonReinhard: Okay that would mean that system memory is then reported twice, once on the CPU and once on the GPU... But I guess that's just what it is then

include/alpaka/api/host/sysInfo.hpp
AntonReinhard: Oh okay. I thought this was just something to call when a user specifically needs it.

AntonReinhard (approved):

PR #422, opened 2026-01-19
Add example readix sort implementation

Wer-Wolf:
WIP

### DIFF
--- a/example/radixsort/src/radixSort.cpp
+++ b/example/radixsort/src/radixSort.cpp
@@ -3,13 +3,134 @@
  */
 
 #include <alpaka/alpaka.hpp>
+#include <alpaka/onAcc/tag.hpp>
 
 #include <boost/program_options.hpp>
+#include <cstddef>
+#include <cstdint>
 #include <iostream>
 #include <vector>
 
 namespace alpaka::example::radixSort
 {
+    constexpr auto chunkSize = CVec<uint32_t, 16>{};
+
+    struct RadixCountKernel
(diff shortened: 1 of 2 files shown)


example/radixsort/src/radixSort.cpp
```diff
@@ -0,0 +1,178 @@
...
+                            sharedCountTable[i] = 0;
+                        }
+                    }
+                }
+
+                onAcc::syncBlockThreads(acc);
+
+                for(auto frameElemIdxMD : onAcc::makeIdxMap(acc, onAcc::worker::linearThreadsInBlock, IdxRange{acc[frame::extent].product()}))
+                {
+                    auto const globalDataIdxMD = frameIdxMD * acc[frame::count] + frameElemIdxMD;
```
psychocoderHPC: BUG, this is the correct solution:

```C++
auto const globalDataIdxMD = frameIdxMD * acc[frame::extent] + frameElemIdxMD;
```

You must use the frame extent and not the number of frames.

```C++
// offset in elements to the begin of the frame data
frameIdxMD * acc[frame::extent]
```

Wer-Wolf: Good catch, now it works as expected. What is the difference between `frame::count` and `frame::extent`?

SimeonEhrig: The `frame::extent` is the "size" of a frame. The `frame::count` is the number of frames.

PR #428, opened 2026-01-26
clean README

psychocoderHPC:
- move old documantion from the README.md to read the docs
- remove duplicated documentation


SimeonEhrig (changes requested):
The CMake part should be part of the Advanced CMake section (rename it from `CMake Arguments` to `CMake`).

First merge #429

SimeonEhrig (approved):

PR #431, opened 2026-01-27
add CMake option `alpaka_COMPILE_PEDANTIC`

psychocoderHPC:
Add new CMake option `alpaka_COMOILE_PEDANTIC` to apply stricter compiler flags like `-Wall`, `-Werror`, depending on the compiler.
Typically, this flag will be used in the CI, but it can be used for local testing too.

Currently, we have fewer strict flags compared to alpaka mainline. 
More pedantic flags will be added within in separate PRs.

Note: The PR prioritizes using `alpaka::unsed()` over `[[maybe_unsued]] in the function signature.
Because it is for the caller not important if the parameter is internally used.
In seldem cases e.g. where `T_Acc` is used as type or `onAcc::concepts::Acc` is part of the function interface the instance name is removed to avoid the warning unused parameter, this is mostly only applied in test cases.


include/alpaka/api/unifiedCudaHip/Device.hpp
```diff
@@ -99,7 +99,7 @@ namespace alpaka::onHost
 
             friend struct onHost::internal::GetNativeHandle;
 
-            [[nodiscard]] uint32_t getNativeHandle() const noexcept
+            [[nodiscard]] int getNativeHandle() const noexcept
```
psychocoderHPC, line 102: note: sneaked this into this PR. The native type for a device is an integer. If I not change it here I need to cast the handle to an integer to avoid a warning but this is the correct fix.

benchmark/babelstream/CMakeLists.txt
```diff
@@ -29,12 +29,5 @@ target_link_libraries(${_TARGET_NAME} PRIVATE Catch2::Catch2WithMain)
...
 #Run as a ctest
-if(alpaka_CI)
-    # Only run for release builds since this is a benchmark
-    if(CMAKE_BUILD_TYPE STREQUAL "Release")
-        add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME})
-    endif()
-else()
-    # For a normal benchmark test, number of samples should be equal to the default value.
-    add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME})
-endif()
```
SimeonEhrig, line 40: Only a comment: the code exists, because the debug build takes to much time in CI to be executed (CI timeout).
At the moment, we test only release builds.

SimeonEhrig (approved):

benchmark/babelstream/CMakeLists.txt
psychocoderHPC: good to know, but in alpak3 we do not define `alpaka_CI` at all and we do not have slow accelerators (aka. executors) like we have it with the Threads accelerator in alpaka mainline.

SimeonEhrig: Let's see if the problem is coming back or is solved, because we have no slow executor anymore.

PR #432, opened 2026-01-28
Span support + minor bugfix for getExtent

TimHanel00:
This PR adds support for std::span such that it can be used in as a source or destination type for memory allocations.
concretely  support for std span (getExtent + getPitched overloads).
Additionally this fixes a bug that caused getExtent to fail on certain allocation (internal namespace was missing missing)

### DIFF
--- a/docs/snippets/cheatsheet/cheatsheet.cpp
+++ b/docs/snippets/cheatsheet/cheatsheet.cpp
@@ -287,7 +287,22 @@ auto main() -> int
 
             unused(hostView, deviceView);
         }
-
+        {
+            // BEGIN-CHEATSHEET-allocLike
+            // This allocLike + memcpy pattern is not specific to std::span; it also works with std::vector/std::array
+            // and with alpaka Buffers/Views.
+            DataType arr[]{0, 1, 2, 3, 4, 5, 6, 7, 8};
+            // construct a span to an existing container
+            std::span myHostSpan = std::span{arr};
+            // create a one-dimensional deviceBuffer with the same extent as 'myHostSpan'
+            auto deviceBuffer = alpaka::onHost::allocLike(device, myHostSpan);
+            // Copy host -> device directly from the span into the allocated device buffer.
+            // Note: if the queue is asynchronous, ensure the source memory container stays alive until the copy
+            // completes.
+            alpaka::onHost::memcpy(blockingQueue, deviceBuffer, myHostSpan);
+            // END-CHEATSHEET-allocLike
+            unused(myHostSpan, deviceBuffer);
+        }
         {
             concepts::IBuffer auto buffer = onHost::allocHost<DataType>(numElements);
             buffer.destructorWaitFor(queue);
--- a/docs/source/basic/cheatsheet.rst
+++ b/docs/source/basic/cheatsheet.rst
@@ -252,6 +252,14 @@ Copy multidimensional buffer/view or span data
     :end-before: END-CHEATSHEET-memcpy
     :dedent:
 
+Allocate a buffer with the same extents from a std::span, std::vector or std::array
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  .. literalinclude:: ../../snippets/cheatsheet/cheatsheet.cpp
+    :language: cpp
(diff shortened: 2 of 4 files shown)


docs/snippets/cheatsheet/cheatsheet.cpp
```diff
@@ -286,7 +286,15 @@ auto main() -> int
 
             unused(hostView, deviceView);
         }
+        {
+            // BEGIN-CHEATSHEET-useAStdSpan
+            int a[]{0, 1, 2, 3, 4, 5, 6, 7, 8};
+            std::span span = std::span{a};
+            auto deviceBuffer = alpaka::onHost::allocLike(device, span);
```
SimeonEhrig:
```suggestion
            // the deviceBuffer is 1D
            auto deviceBuffer = alpaka::onHost::allocLike(device, span);
```

If you check the API of `allocLike` it is clear, but my first thought was: Is the span creating a 1D buffer or does the span describe the dimension of the buffer? In this case it means a 9D buffer

docs/snippets/cheatsheet/cheatsheet.cpp
```diff
@@ -286,7 +286,15 @@ auto main() -> int
...
+            alpaka::onHost::memcpy(queue, deviceBuffer, span);
```
SimeonEhrig:
```suggestion
            alpaka::onHost::memcpy(queue, deviceBuffer, span);
            // END-CHEATSHEET-useAStdSpan
```

SimeonEhrig (changes requested):

docs/snippets/cheatsheet/cheatsheet.cpp
TimHanel00: Based on this comment and after realizing that `allocLike` had not been mentioned in the documentation before, I slightly rephrased and generalized this docu snippet. I would appreciate any suggestions regarding the accuracy and wording of this change.

docs/source/basic/cheatsheet.rst
```diff
@@ -251,7 +251,13 @@ Copy multidimensional buffer/view or span data
     :start-after: BEGIN-CHEATSHEET-memcpy
     :end-before: END-CHEATSHEET-memcpy
     :dedent:
-
+Allocate and copy from an existing buffer or view
```
SimeonEhrig:
```suggestion
Allocate a buffer with the same extents from a std::span, std::vector or std::array
```

docs/snippets/cheatsheet/cheatsheet.cpp
```diff
@@ -286,7 +286,22 @@ auto main() -> int
 
             unused(hostView, deviceView);
         }
-
+        {
+            // BEGIN-CHEATSHEET-allocLike
+            // This allocLike + memcpy pattern is not specific to std::span; it also works with std::vector/std::array
+            // and with Alpaka Buffers/Views.
```
SimeonEhrig:
```suggestion
            // This allocLike + memcpy pattern is not specific to std::span; it also works with std::vector/std::array
            // and with alpaka Buffers/Views.
```

TimHanel00: I mainly wanted to avoid implying that `std::span` is required here — since `allocLike` also works directly with other containers/views (e.g., std::vector, std::array, Alpaka buffers/views).
That said, I’m fine keeping this snippet solely std::span-focused if you prefer; in that case, the suggested changes fit well.

SimeonEhrig: ~Do you mean `alpaka::alocLike` is already specialized for `std::vector` and `std::array`?~

Okay, I see. Please change the heading and make `alpaka` lowercase. Then I will merge.

psychocoderHPC:
@TimHanel00 Please also rebase against the current dev branch since we have now more strict compiler flags in the CI.

see: #431

docs/snippets/cheatsheet/cheatsheet.cpp
TimHanel00: well yes since both containers provide a trait for 'getExtent', they should be directly usable as arguments for 'allocLike', if I am not mistaken.

SimeonEhrig: @TimHanel00 Can you please check my suggestions.

SimeonEhrig (approved):

PR #434, opened 2026-02-04
improve FrameSpec

SimeonEhrig:
- make constructors constexpr
- move `m_numFrames`, `m_frameExtent` and `m_threadSpec` to private scope and add getters
- rename member type `type` to `value_type`

### DIFF
--- a/include/alpaka/onHost/FrameSpec.hpp
+++ b/include/alpaka/onHost/FrameSpec.hpp
@@ -40,50 +40,65 @@ namespace alpaka::onHost
         alpaka::concepts::Vector<typename T_NumFrames::type, T_NumFrames::dim()> T_ThreadExtents>
     struct FrameSpec
     {
-        using type = typename T_NumFrames::type;
+        using index_type = typename T_NumFrames::type;
 
         using NumFramesVecType = T_NumFrames;
         using FrameExtentsVecType = T_FrameExtents;
         using ThreadExtentsVecType = T_ThreadExtents;
         using ThreadSpecType = ThreadSpec<T_NumFrames, T_ThreadExtents>;
 
-        static consteval uint32_t dim()
-        {
-            return T_FrameExtents::dim();
-        }
-
-        T_NumFrames m_numFrames;
-        T_FrameExtents m_frameExtent;
+    private:
+        NumFramesVecType m_numFrames;
+        FrameExtentsVecType m_frameExtents;
         ThreadSpecType m_threadSpec;
 
-        FrameSpec(T_NumFrames const& numFrames, T_FrameExtents const& frameExtent)
+    public:
+        constexpr FrameSpec(T_NumFrames const& numFrames, T_FrameExtents const& frameExtent)
             : m_numFrames(numFrames)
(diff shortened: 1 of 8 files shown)


include/alpaka/onHost/FrameSpec.hpp
```diff
@@ -81,10 +82,25 @@ namespace alpaka::onHost
         {
         }
 
-        auto getThreadSpec() const
+        [[nodiscard]] constexpr NumFramesVecType getNumFrames() const noexcept
+        {
+            return m_numFrames;
+        }
+
+        [[nodiscard]] constexpr FrameExtentsVecType getFrameExtent() const noexcept
```
psychocoderHPC: This should be `getFrameExtents` with an `s` at the end.
Some as in alpaka mainline the small `s` is saying that a vector type is coming back.

note: We discussed it in the dev meeting a while ago and decided to keep `s` instead of `*MD`

include/alpaka/onHost/FrameSpec.hpp
```diff
@@ -40,37 +40,38 @@ namespace alpaka::onHost
         alpaka::concepts::Vector<typename T_NumFrames::type, T_NumFrames::dim()> T_ThreadExtents>
     struct FrameSpec
     {
-        using type = typename T_NumFrames::type;
+        /// value_type of all vector components (T_NumFrames, T_FrameExtents and T_ThreadExtents)
+        using value_type = typename T_NumFrames::type;
```
psychocoderHPC: I am not sure but maybe this should be `index_type` because this is reflaction which type should be used for index operations.
In the class `Vec` there is a `type` which will be renamed soon in `value_type` and a type whcih should be used for the `operator[]` which is always `uint32_t`.

This is only a suggestion. I am also fine with `value_type`

psychocoderHPC (changes requested):
One small required renaming and one point where I am unsure what is the better naming.

include/alpaka/onHost/FrameSpec.hpp
SimeonEhrig: Sure. I was not sure, what the type means because it was named `type` without doc string. But when you wrote, that the type should be named `index_type`, it make sense what the type actual means.

I copied `value_type` simply from `Vec`.

include/alpaka/onHost/FrameSpec.hpp
SimeonEhrig: I renamed it

include/alpaka/onHost/FrameSpec.hpp
SimeonEhrig: I also removed the doc string. I think `index_type` is self explaining.

psychocoderHPC: > I copied value_type simply from Vec.

For `Vec` `value_type` makes sense because it is the type stored in the struct. `Vec` also has an `index_type` which is `uint32_t` and defines the type which is used if you index e.g. via the `operator[]` into the vector.

psychocoderHPC (approved):

PR #436, opened 2026-02-05
introduce `std::simd` suppport

psychocoderHPC:
- add support for `std::simd`
- refactor `Simd`
- introduce `SimdMask`
- add CI tests to enforce using `alpaka_SIMD=EMULATION` because in most cases `std::simd` will be used
- remove some methods in `Simd` which make no sense and were copied over from the `Vec` implementation
- change `constexpr`  simd tests into runtime tests because not all compilers can initialize simd vectors as constexpr. (should be no big deal in real live)

Note: I am unhappy with the implementation because there are many code duplication. I will focus in seperate PRs to solve this. This PR is focusing on introducing std::simd without breaking the perfromance for existing code


psychocoderHPC:
The issues in the CI are real issues, I removed the alignment from the type `Simd` and therefore I can not write the EmuSimd copy the way I wrote it.
I will handle it after my vacations

cmake/alpakaCommon.cmake
```diff
@@ -229,7 +229,7 @@ try_compile(
     CXX_STANDARD_REQUIRED TRUE
     CXX_EXTENSIONS FALSE
 )
-if(alpaka_HAS_STD_ATOMIC_REF AND (NOT alpaka_ACC_CPU_DISABLE_ATOMIC_REF))
```
psychocoderHPC, line 232: Left over from alpak mainline copy past (sneaked into this PR)

psychocoderHPC:
@SimeonEhrig @ikbuibui ready for review

psychocoderHPC:
@SimeonEhrig I set it again to draft because I realized a performance drop with bable stream.
Somehow under CUDA vector loads will not be used.

cmake/tests/StdSimd.cpp
```diff
@@ -0,0 +1,25 @@
+/* Copyright 2022 Jan Stephan
```
SimeonEhrig: Did you copy it from alpaka mainline?

tests/unit/algorithm/concurrent.cpp
```diff
@@ -42,7 +42,7 @@ struct StencilAddMin
         alpaka::concepts::Simd auto aSimd = a.load();
         alpaka::concepts::Simd auto bSimd = b.load();
         for(uint32_t idx = 0u; idx < simdWidth; ++idx)
-            aSimd[idx] += math::min(aSimd[idx], bSimd[idx]);
+            aSimd[idx] += math::min(+aSimd[idx], +bSimd[idx]);
```
SimeonEhrig, line 48: What does this line? I know, the `+` can be used for some type cast tricks.

Is this the way, you user needs to write functors for `concurrent`?

psychocoderHPC:
@SimeonEhrig Now the PR is ready

cmake/tests/StdSimd.cpp
```diff
@@ -0,0 +1,25 @@
...
+#    if !defined(ALPAKA_HAS_STD_SIMD)
+#        define ALPAKA_HAS_STD_SIMD 1
+#    endif
+#endif
+
+
+auto main() -> int
+{
+    alpakaStdSimd::native_simd<float> simdFloat = 0.0f;
+    return EXIT_SUCCESS;
```
ikbuibui, line 25: is this really working? EXIT_SUCCESS is not defined

cmake/alpakaCommon.cmake
```diff
@@ -246,6 +246,41 @@ if(NOT alpaka_HAS_STD_ATOMIC_REF)
...
+set_property(CACHE alpaka_SIMD PROPERTY STRINGS "DEFAULT;STDSIMD;EMULATION")
+
+if(alpaka_SIMD STREQUAL "STDSIMD" OR alpaka_SIMD STREQUAL "DEFAULT")
+    # Check for C++ std::simd
+    # we only check the CXX compiler, equal to OpenMP which is checked for the CXX compiler only too.
+    try_compile(
+        alpaka_HAS_STD_SIMD # Result stored here
+        "${PROJECT_BINARY_DIR}/alpakaFeatureTests" # Binary directory for output file
+        SOURCES
+            "${_alpaka_FEATURE_TESTS_DIR}/StdSimd.cpp" # Source file
```
ikbuibui, line 263: I'm not sure but you might need to pass CMake flags into this. For example if `-stdlib=libc++` is missing, it might cause a problem

cmake/alpakaCommon.cmake
```diff
@@ -246,6 +246,41 @@ if(NOT alpaka_HAS_STD_ATOMIC_REF)
...
+        CXX_STANDARD 20
+        CXX_STANDARD_REQUIRED TRUE
```
ikbuibui: Breaks compatibility with projects which might be using C++23/26 standard.

ikbuibui (commented):
Initial review, will continue tomorrow

cmake/tests/StdSimd.cpp
psychocoderHPC: yes, because as soon as it compiles we have `std::simd` support.
We only check from CMake if the example is compilable.

cmake/alpakaCommon.cmake
psychocoderHPC: It is using the CMake language CXX, so everything you set as flags for your normal CXX application will be set here too. It is equal to the `atomic_ref` check where we do not forward flags by hand.

cmake/alpakaCommon.cmake
psychocoderHPC: Why does this breaks using C++23+?
`CXX_STANDARD_REQUIRED` is saying that C++20 is mendentory and the compiler is not allowed to fall back to older standards.

https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD_REQUIRED.html

cmake/tests/StdSimd.cpp
psychocoderHPC: discussed it offline `#include <cstdlib>` is missing

cmake/alpakaCommon.cmake
psychocoderHPC: we should use `alpaka_CXX_STANDARD` standard instead of hard code 20

include/alpaka/Simd.hpp
```diff
@@ -318,14 +232,24 @@ namespace alpaka
...
+         *
+         * @return The value type, by copy.
+         */
+        constexpr type operator[](std::integral auto const idx) const
         {
-            return Storage::operator[](idx);
+            return asStorage()[idx];
         }
 
         /** named member access
```
SimeonEhrig: This documentation is not rendered in doxygen.

CMakeLists.txt
```diff
@@ -126,6 +126,7 @@ install(DIRECTORY ${_alpaka_TESTING_DIR} DESTINATION ${CMAKE_INSTALL_DATAROOTDIR
 
 # File is required to validate the C++20 feature std::atomic_ref
```
ikbuibui:
```suggestion
# Files required to validate optional features std::atomic_ref and std::simd
```

include/alpaka/Simd.hpp
SimeonEhrig, line 294: Is this a group of functions for an SIMD object with 4 elements?
Is `r()`, `g()`, `b()` and `a()` also group of functions for a SIMD object with 4 elements?
Is `r(),g(),b(),a()` an alias for `x(),y(),z(),w()`?

SimeonEhrig: Write a line, that the functions are available depending of the number of elements in the SIMD object.

include/alpaka/simd/internal/EmuSimd.hpp
```diff
@@ -0,0 +1,249 @@
...
+
+            static constexpr auto fill(T_Type const& value)
+            {
+                EmuSimd result([&value](uint32_t const) { return static_cast<T_Type>(value); });
+                return result;
+            }
+
+            template<
+                typename F,
+                std::enable_if_t<std::is_invocable_v<F, std::integral_constant<uint32_t, 0u>>, uint32_t> = 0u>
```
SimeonEhrig: Can this replaced with a concept?

include/alpaka/SimdMask.hpp
```diff
@@ -0,0 +1,537 @@
...
+        SimdMask() = default;
+
+        /** Initialize via a generator expression
+         *
+         * The generator must return the value for the corresponding index of the component which is passed to the
+         * generator.
+         */
+        template<
+            typename F,
+            std::enable_if_t<std::is_invocable_v<F, std::integral_constant<uint32_t, 0u>>, uint32_t> = 0u>
```
SimeonEhrig: Concept?

include/alpaka/Simd.hpp
```diff
@@ -144,45 +92,47 @@ namespace alpaka
 
         constexpr Simd() = default;
 
+        using Storage::asNativeType;
+
+        constexpr auto& asStorage()
```
SimeonEhrig, line 104: Please add a small docstring

include/alpaka/SimdMask.hpp
```diff
@@ -0,0 +1,537 @@
...
+        {
+            return Storage::operator[](idx);
+        }
+
+        constexpr type operator[](std::integral auto const idx) const
+        {
+            return static_cast<type>(Storage::operator[](idx));
+        }
+
+        /** named member access
```
SimeonEhrig, line 254: Same problem with the doc like in `std::Simd`. See other comment.

tests/unit/algorithm/transform.cpp
```diff
@@ -58,7 +58,7 @@ struct ScalarOpWithAcc
     constexpr auto operator()(onAcc::concepts::Acc auto const&, auto const& a, auto const& b) const
     {
         using ValueType = trait::GetValueType_t<ALPAKA_TYPEOF(a)>;
-        return math::min(a + ValueType{1}, b);
+        return math::min(a + ValueType{1}, +b);
```
SimeonEhrig, line 63: Is this user code?

include/alpaka/Simd.hpp
SimeonEhrig: GitHub is broken. This is the correct code link: https://github.com/psychocoderHPC/alpaka3/blob/b51b270de60b253c7714bc2a569b9a8b391b3fe6/include/alpaka/Simd.hpp#L255-L264

cmake/tests/StdSimd.cpp
psychocoderHPC: No but copied a file because of laziness to copy the header^^

tests/unit/algorithm/concurrent.cpp
psychocoderHPC: I added a comment. The problem is that `bSimd[idx]` can be a reference wrapper and we need to cast this to a value.

include/alpaka/Simd.hpp
psychocoderHPC: fixed but doxygen sub grouping is not working. I is not forwarding the subgroup doc string to all signatures in the subgroup, only to the first.

tests/unit/algorithm/concurrent.cpp
psychocoderHPC: I added a comment

include/alpaka/Simd.hpp
psychocoderHPC: subgrouping is not working as expected, therefore I changes the doc string

include/alpaka/simd/internal/StdSimd.hpp
```diff
@@ -0,0 +1,177 @@
...
+
+            static constexpr auto fill(T_Type const& value)
+            {
+                BaseType ret(value);
+                return StdSimd{ret};
+            }
+
+            constexpr void copyFrom(T_Type const* data, alpaka::concepts::Alignment auto alignment)
+            {
+                if constexpr((alignment.template get<T_Type>() % alignof(ALPAKA_TYPEOF(*this))) == 0u)
```
ikbuibui:
```suggestion
                if constexpr((alignment.template get<T_Type>() % alpakaStdSimd::memory_alignment_v<BaseType>) == 0u)
```

include/alpaka/simd/internal/StdSimd.hpp
```diff
@@ -0,0 +1,177 @@
...
+                    this->asNativeType().copy_from(data, alpakaStdSimd::vector_aligned);
+                else
+                    this->asNativeType().copy_from(data, alpakaStdSimd::element_aligned);
+            }
+
+            constexpr void copyTo(auto* data, alpaka::concepts::Alignment auto alignment) const
+            {
+                if constexpr((alignment.template get<T_Type>() % alignof(ALPAKA_TYPEOF(*this))) == 0u)
```
ikbuibui:
```suggestion
                if constexpr((alignment.template get<T_Type>() % alpakaStdSimd::memory_alignment_v<BaseType>) == 0u)
```

include/alpaka/simd/internal/EmuSimd.hpp
psychocoderHPC: yes, it is copied from the vector class, which was original C++17 code

include/alpaka/simd/internal/StdSimdMask.hpp
```diff
@@ -0,0 +1,179 @@
...
+            {
+                if constexpr((alignment.template get<T_Type>() % alignof(ALPAKA_TYPEOF(*this))) == 0u)
+                    this->asNativeType().copy_from(data, alpakaStdSimd::vector_aligned);
+                else
+                    this->asNativeType().copy_from(data, alpakaStdSimd::element_aligned);
+            }
+
+            constexpr void copyTo(auto* data, alpaka::concepts::Alignment auto alignment) const
+            {
+                if constexpr((alignment.template get<T_Type>() % alignof(ALPAKA_TYPEOF(*this))) == 0u)
```
ikbuibui:
```suggestion
                if constexpr((alignment.template get<T_Type>() % alpakaStdSimd::memory_alignment_v<BaseType>) == 0u)
```

include/alpaka/simd/internal/StdSimdMask.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+            static constexpr auto fill(T_Type const& value)
+            {
+                BaseType ret(value);
+                return StdSimdMask{ret};
+            }
+
+            constexpr void copyFrom(T_Type const* data, alpaka::concepts::Alignment auto alignment)
+            {
+                if constexpr((alignment.template get<T_Type>() % alignof(ALPAKA_TYPEOF(*this))) == 0u)
```
ikbuibui:
```suggestion
                if constexpr((alignment.template get<T_Type>() % alpakaStdSimd::memory_alignment_v<BaseType>) == 0u)
```

include/alpaka/simd/internal/StdSimdMask.hpp
```diff
@@ -0,0 +1,179 @@
...
+                    this->asNativeType()[i] = static_cast<bool>(initArgs[i]);
+            }
+
+            template<typename... T_Args>
+            requires(sizeof...(T_Args) == T_width && (std::same_as<T_Args, bool> && ...))
+            constexpr StdSimdMask(T_Args&&... args) : BaseType{}
+            {
+                std::array<bool, T_width> const initArgs{ALPAKA_FORWARD(args)...};
+                for(uint32_t i = 0u; i < T_width; ++i)
+                    this->asNativeType()[i] = initArgs[i];
```
ikbuibui, line 73:
```suggestion
this->asNativeType().copy_from(initArgs.data(), alpakaStdSimd::element_aligned);
```

ikbuibui (commented):
Some thoughts and comments.
- I thought passing the simd mask by value could possibly be better (discussed offline that most probably not)
- instead of for loops we could use copy_n, but this was also discussed as most likely slow, as it has a runtime `n` and the for loop can be unrolled by the compiler.
- the functions like `revert`, `-`, `rshrink`, `min` etc can be implemented as generators, avoiding creating an intermediate object, for example
  ```
  constexpr Simd revert() const
  {
      return Simd([&](auto const i) constexpr { return (*this)[T_width - 1u - i]; });
  }
  ```

tests/unit/algorithm/transform.cpp
psychocoderHPC: Yes it is user code. I added a comment why this is required.
With a functor which is not using auto all would be fine without the casting trick.

```C++
struct ScalarOpWithAcc
{
    constexpr auto operator()(onAcc::concepts::Acc auto const&, float const& a, float const& b) const
    {
        return math::min(a + float{1}, b);
    }
};
```

include/alpaka/simd/internal/StdSimd.hpp
psychocoderHPC: done, offline discussed I missed that std::simd are "proxy" types and that the memory layout of the data is not required to be equal to the layout of the std::simd instance.

ikbuibui (approved):

tests/unit/algorithm/transform.cpp
SimeonEhrig: We discussed it offline. I'm still no sure, if we maybe can implement something, which cast the type automatically. But for this PR it is fine and improving the semantic is something for a possible upcoming PR.

SimeonEhrig (approved):

PR #438, opened 2026-02-09
refactored ThreadSpec

SimeonEhrig:
- rename member type `type` to `value_type`
- make member variables `m_numBlocks` and `m_numThreads` private and add getters
- ThreadSpec is not mutable anymore

I think it make more sense that ThreadSpec is immutable to avoid undefined behavior/bugs. There was only one use case, where the value of a ThreadSpec was changed. I replaced it by constructing a new threadSpec. It is in an example not performance critical.

### DIFF
--- a/include/alpaka/onHost/ThreadSpec.hpp
+++ b/include/alpaka/onHost/ThreadSpec.hpp
@@ -18,22 +18,34 @@ namespace alpaka::onHost
         alpaka::concepts::Vector<typename T_NumBlocks::type, T_NumBlocks::dim()> T_NumThreads>
     struct ThreadSpec
     {
-        using type = typename T_NumBlocks::type;
+        using index_type = typename T_NumBlocks::type;
         using NumBlocksVecType = typename T_NumBlocks::UniVec;
         using NumThreadsVecType = T_NumThreads;
 
-        static consteval uint32_t dim()
-        {
-            return T_NumThreads::dim();
-        }
-
+    private:
         NumBlocksVecType m_numBlocks;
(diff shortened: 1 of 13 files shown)


include/alpaka/onHost/ThreadSpec.hpp
```diff
@@ -18,23 +18,35 @@ namespace alpaka::onHost
...
         NumThreadsVecType m_numThreads;
 
+    public:
         constexpr ThreadSpec(T_NumBlocks const& numBlocks, T_NumThreads const& numThreadsPerBlock)
             : m_numBlocks(numBlocks)
             , m_numThreads(numThreadsPerBlock)
         {
         }
+
+        [[nodiscard]] constexpr NumThreadsVecType const& getNumThreads() const noexcept
```
SimeonEhrig, line 36: First, I returned a value but the kernel `acc` object wants to keep a `std::cref` on the value.
https://github.com/alpaka-group/alpaka3/pull/438/changes#diff-4fd6c7f00bee18d3a2e2570440844a44da6b86c00d559bbe4760aa5daaac5683R43

I think it makes sense in terms of performance (register usage). Therefore maybe we should return also const references instead values in the FrameSpec getters.

psychocoderHPC: To remove the `std::cref` you can use `alpaka::unwrapp()`.
Note there is one `p` to much in the word `unwrapp`^^ but this will be changed soon.

psychocoderHPC (changes requested):

psychocoderHPC (approved):

PR #439, opened 2026-02-16
Implement begin() and end() iterator for IdxRange

SimeonEhrig:
- The begin() and end() function can be used on host side, to iterate all positions.
- On host side, it implements the FlatIdxContainer iterator for a serial executor and a contiguous layout.
- On accelerator side, still use makeMapIdx to iterate in parallel.

### DIFF
--- a/include/alpaka/mem/IdxRange.hpp
+++ b/include/alpaka/mem/IdxRange.hpp
@@ -8,6 +8,7 @@
 #include "alpaka/core/PP.hpp"
 #include "alpaka/core/common.hpp"
 #include "alpaka/mem/BoundaryIter.hpp"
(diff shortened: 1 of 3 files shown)


psychocoderHPC (approved):

include/alpaka/mem/IdxRange.hpp
```diff
@@ -75,6 +76,35 @@ namespace alpaka
             return m_end - m_begin;
         }
 
+        /** Begin iterator to iterate all positions in the range. It first iterates the fastest index (the one on the
```
psychocoderHPC, line 79: btw: Doxygen comments start with a brief description. Then an empty line and a long description.

psychocoderHPC (approved):

PR #442, opened 2026-02-17
Fix broken documentation `std::span` snippet

TimHanel00:
`std::span` does not provide an API therefore it cannot be used with `alpaka::memcpy` directly. I modified the docu snippet that was using `std::span` - it now shows how to use a host container (`std::vector`) in combination with `allocLike` and `memcpy` instead. This version has been verified through dedicated testing across multiple backends and devices.

### DIFF
--- a/docs/snippets/cheatsheet/cheatsheet.cpp
+++ b/docs/snippets/cheatsheet/cheatsheet.cpp
@@ -289,19 +289,18 @@ auto main() -> int
         }
         {
             // BEGIN-CHEATSHEET-allocLike
-            // This allocLike + memcpy pattern is not specific to std::span; it also works with std::vector/std::array
+            // This allocLike + memcpy pattern is not specific to std::vector; it also works with std::array
             // and with alpaka Buffers/Views.
-            DataType arr[]{0, 1, 2, 3, 4, 5, 6, 7, 8};
-            // construct a span to an existing container
-            std::span myHostSpan = std::span{arr};
-            // create a one-dimensional deviceBuffer with the same extent as 'myHostSpan'
-            auto deviceBuffer = alpaka::onHost::allocLike(device, myHostSpan);
-            // Copy host -> device directly from the span into the allocated device buffer.
+            // Construct a host container (here: std::vector) with arbitrary values.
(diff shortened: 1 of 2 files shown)


SimeonEhrig:
How does the `std::span` version pass the CI? The code should be tested automatically.

I will check it.

SimeonEhrig:
@TimHanel00 I don't understand. I tested the original code and it is working. The following code produces the output: `0 1 2 3 4 5 6 7 8`

```c++
        {
            // BEGIN-CHEATSHEET-allocLike
            // This allocLike + memcpy pattern is not specific to std::span; it also works with std::vector/std::array
            // and with alpaka Buffers/Views.
            DataType arr[]{0, 1, 2, 3, 4, 5, 6, 7, 8};
            // construct a span to an existing container
            std::span myHostSpan = std::span{arr};
            // create a one-dimensional deviceBuffer with the same extent as 'myHostSpan'
            auto deviceBuffer = alpaka::onHost::allocLike(device, myHostSpan);
            // Copy host -> device directly from the span into the allocated device buffer.
            // Note: if the queue is asynchronous, ensure the source memory container stays alive until the copy
            // completes.
            alpaka::onHost::memcpy(blockingQueue, deviceBuffer, myHostSpan);
            // END-CHEATSHEET-allocLike
            auto hostBuffer = alpaka::onHost::allocHostLike(deviceBuffer);
            alpaka::onHost::memcpy(blockingQueue, hostBuffer, deviceBuffer);
            for(auto e : hostBuffer)
            {
                std::cout << e << " ";
            }
            std::cout << std::endl;
            unused(myHostSpan, deviceBuffer);
        }
 ```

psychocoderHPC (approved):

psychocoderHPC:
Note: The problem is only triggered for non-host memcopies where we need to use the API to find the copy direction.

psychocoderHPC:
I opened an issue https://github.com/alpaka-group/alpaka3/issues/443 to ensure that we improve our concepts for memcopy to avoid such issues.

TimHanel00:
Apologies @SimeonEhrig - I should have made it clearer in the PR description that the error only occurred for non-host memcopies.

PR #446, opened 2026-02-19
refactor doxygen for `GetSubView()` and `concepts::IView`

psychocoderHPC:
Due to the refactoring IDE's can help with providing doxygen help during coding.
Never the less it is usefull that a concepts has also documenation to allow others to implement a class following the concept, thats why the concepts reference back to the example implemenation documantion.

### DIFF
--- a/include/alpaka/mem/concepts/IView.hpp
+++ b/include/alpaka/mem/concepts/IView.hpp
@@ -31,26 +31,14 @@ namespace alpaka::concepts
 
             /** @brief Creates a sub view to a part of the memory.
              *
-             * The sub view has the same dimension as the original.
+             * @see alpaka::View::getSubView
              *
-             * @param extents Number of elements for each dimension. Each number must be less than or equal to
-             * the number of elements in the original dimension.
-             * @return View which is pointing only to a part of the original view.
+             * @{
              */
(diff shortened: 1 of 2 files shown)


include/alpaka/mem/concepts/IView.hpp
```diff
@@ -31,26 +31,14 @@ namespace alpaka::concepts
 
             /** @brief Creates a sub view to a part of the memory.
              *
-             * The sub view has the same dimension as the original.
+             * @see alpaka::View::getSubView
```
psychocoderHPC, line 34: I do not point directly to the signature with one and two arguments because it would be 

```
@see alpaka::View::getSubView(alpaka::concepts::VectorOrScalar auto const&, alpaka::concepts::VectorOrScalar auto const&)
```

In case we refactor interfaces and change the e.g. the concept or the concept name we will definitely miss to fix the documentation. 
We could avoid passing doxygen runs if there is a missing or wrong reference but currently IMO this is slowing down our development to much.

SimeonEhrig: For the moment it is fine. If we enable that the doc build CI job fails if there is an error in doxygen or shpinx doc, we can discuss this again.

SimeonEhrig (approved):

PR #447, opened 2026-02-19
fix `FrameSpec` and `ThreadSpec` can not be mutated

psychocoderHPC:
Add operators to mutate the state.

With the last refactoring, we removed the possibility of generating the frame specification with `onHost::getFrameSpec()` and then post mortem change it. 
I use this very often for performance optimizations or testing effects like different numbers of thread blocks and frames.

### DIFF
--- a/include/alpaka/onHost/FrameSpec.hpp
+++ b/include/alpaka/onHost/FrameSpec.hpp
@@ -86,12 +86,27 @@ namespace alpaka::onHost
             return m_numFrames;
         }
 
+        [[nodiscard]] constexpr NumFramesVecType& getNumFrames() noexcept
+        {
+            return m_numFrames;
+        }
+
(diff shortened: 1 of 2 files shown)


SimeonEhrig:
I changed it on purpose. I found only one case in the CUDA kernel like example, where it was required: https://github.com/alpaka-group/alpaka3/pull/438/changes#diff-bd477ff53625d14a57324e319e90b8eef4622082709696b1940c0fefe948f001

Therefore it I decided to make it immutable because that I think the guaranty that ThreadSpec is immutable is more valuable than making this corner case nice. If you have a good case, we can change it back.

SimeonEhrig:
In a offline discussion, we agreed that we keep it, as it is until we have a really good use case. At the moment, the guaranty that a frameSpec object cannot be modified is a bigger benefit, that a less verbose code.

Note: If the members variables should be mutable, we should make the variables public, remove the `m_` prefix and delete the getters.

Issue #449, opened 2026-02-23
CallBackThread refactoring with `std::jthread`


psychocoderHPC:
The Callback thread should be refactored and `std::jthread` should be used to avoid manual error-prone shutdown implementations and self-joining issues.

https://github.com/alpaka-group/alpaka3/blob/dev/include/alpaka/core/CallbackThread.hpp

PR #450, opened 2026-02-23
fix `allocDeffered()` and `keepAlive()` usage

### LINKED ISSUE #394: Deadlock using keepAlive on an allocDeferred buffer
In PR #384 we noticed that the `buffer.keepAlive()` sometimes deadlocks when used together with a `allocDeferred` buffer, specifically with the syclGeneric/oneAPI backend.

psychocoderHPC:
fix #394

OneApi Sycl deadlocked in cases where `allocDefeered()` was used together with buffers `keepAlive()`.
The reason was that a native sycl host task was enqueued within an sycl host task, which is not allowed.

Together with this PR I opened #449 to refactor the `CallBackThread` with modern C++ jthreads.

Note: I kept `alloc` instead of `allocDeferred` in the example and in the algorithms scan implementation.
If `allocDeferred` is only called once on CUDA the latency is very high due to the per stream caching pool setup time of CUDA.
IMO for the algorithm scan we should provide the user the possibility to define which type of memory he prefers for the temporary required memory.

### DIFF
--- a/include/alpaka/core/CallbackThread.hpp
+++ b/include/alpaka/core/CallbackThread.hpp
@@ -17,6 +17,11 @@
 
 namespace alpaka::core
 {
+    /** A thread queue executing tasks asynchronously.
+     *
+     * This object should be used as members of objects which are secured by smart pointers to avoid that a task is
+     * taking over the ownership of the callback thread and therefore can destroy itself before all tasks are executed.
+     */
     class CallbackThread
     {
 #if ALPAKA_COMP_CLANG
@@ -67,10 +72,21 @@ namespace alpaka::core
             {
                 if(std::this_thread::get_id() == m_thread.get_id())
                 {
-                    std::cerr << "ERROR in ~CallbackThread: thread joins itself" << std::endl;
-                    std::abort();
+                    // at this point there should be no tasks in the queue
+                    if(!m_tasks.empty())
+                    {
+                        std::cerr << "ERROR in ~CallbackThread: internal queue is not empty but object is destructed."
+                                  << std::endl;
+                        std::abort();
+                    }
+                    /* We can not joint our self.
+                     * We can only end here if a tasks the callback thread is executing is capturing the object which
+                     * is holding the callback thread.
+                     */
+                    m_thread.detach();
                 }
-                m_thread.join();
+                else
+                    m_thread.join();
             }
         }
 
--- a/include/alpaka/core/CallbackThread.hpp
+++ b/include/alpaka/core/CallbackThread.hpp
@@ -19,7 +19,7 @@ namespace alpaka::core
 {
     /** A thread queue executing tasks asynchronously.
      *
-     * This object should be used as members of objects which are secured by smart pointers to avoid that a task is
+     * This object should be used as a member of objects which are secured by smart pointers to avoid that a task is
      * taking over the ownership of the callback thread and therefore can destroy itself before all tasks are executed.
      */
     class CallbackThread
@@ -79,9 +79,9 @@ namespace alpaka::core
                                   << std::endl;
                         std::abort();
                     }
-                    /* We can not joint our self.
-                     * We can only end here if a tasks the callback thread is executing is capturing the object which
-                     * is holding the callback thread.
+                    /* We can not join ourselves.
+                     * We can only end here if a task that the callback thread is executing is capturing the object
+                     * which is holding the callback thread.
                      */
                     m_thread.detach();
                 }
--- a/include/alpaka/onHost/algo/internal/scan.hpp
+++ b/include/alpaka/onHost/algo/internal/scan.hpp
@@ -458,6 +458,9 @@ namespace alpaka::onHost::internal
     {
         using T_Data = ALPAKA_TYPEOF(inputVec)::value_type;
 
+        /* We do not use allocDeferred here since we measured up to a factor 40 higher latency compared to alloc for
+         * CUDA 12.8 on an A30 for the first call. The reason is the cuda per stream caching pool setup time.
+         */
         auto buf = onHost::alloc<char>(devAcc, scanBufferSize<T_Data>(inputVec.getExtents()));
 
         scan<SCAN_TYPE>(queue, devAcc, exec, buf, outputVec, inputVec);
--- a/tests/unit/mem/keepAlive.cpp
+++ b/tests/unit/mem/keepAlive.cpp
@@ -55,7 +55,7 @@ TEMPLATE_LIST_TEST_CASE("keep alive", "", TestApis)
 
     {
         // enqueue everything in this scope
-        auto scopedBuffer = onHost::alloc<int>(device, N);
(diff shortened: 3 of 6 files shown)


include/alpaka/onHost/algo/internal/scan.hpp
```diff
@@ -458,6 +458,9 @@ namespace alpaka::onHost::internal
     {
         using T_Data = ALPAKA_TYPEOF(inputVec)::value_type;
 
+        /* We do not use allocDeferred here since we measured up to a factor 40 higher latency compared to alloc for
+         * CUDA 12.8 on an A30 for the first call. The reason is the cuda per stream caching pool setup time.
+         */
         auto buf = onHost::alloc<char>(devAcc, scanBufferSize<T_Data>(inputVec.getExtents()));
```
AntonReinhard, line 464: Is this a measurement only of the specific allocDeferred vs alloc? I think for this specific use case, the latency hiding of the actual computation might still make the whole scan call faster overall.

include/alpaka/core/CallbackThread.hpp
```diff
@@ -67,10 +72,21 @@ namespace alpaka::core
...
+                    // at this point there should be no tasks in the queue
+                    if(!m_tasks.empty())
+                    {
+                        std::cerr << "ERROR in ~CallbackThread: internal queue is not empty but object is destructed."
+                                  << std::endl;
+                        std::abort();
+                    }
+                    /* We can not joint our self.
+                     * We can only end here if a tasks the callback thread is executing is capturing the object which
+                     * is holding the callback thread.
```
AntonReinhard:
```suggestion
                    /* We can not join ourself.
                     * We can only end here if a task that the callback thread is executing is capturing the object which
                     * is holding the callback thread.
```

include/alpaka/core/CallbackThread.hpp
```diff
@@ -17,6 +17,11 @@
 
 namespace alpaka::core
 {
+    /** A thread queue executing tasks asynchronously.
+     *
+     * This object should be used as members of objects which are secured by smart pointers to avoid that a task is
+     * taking over the ownership of the callback thread and therefore can destroy itself before all tasks are executed.
```
AntonReinhard:
```suggestion
     * This object should be used as a member of objects which are secured by smart pointers to avoid that a task is
     * taking over the ownership of the callback thread and therefore can destroy itself before all tasks are executed.
```

AntonReinhard:
> IMO for the algorithm scan we should provide the user the possibility to define which type of memory he prefers for the temporary required memory.

The overloads accepting the buffer memory already allow this, no?

include/alpaka/onHost/algo/internal/scan.hpp
SimeonEhrig: @psychocoderHPC Do you use a different queue? Normally you pass the same queue to scan, which you use also to allocate the memory for the `inputVec` and `outputVec`. Therefore the stream should already exists.

psychocoderHPC: I used the same queue, it is not about the queue creation, it is about the caching allocator creation CUDA is using within each stream. The cache is only initialized on the first usage.

> Is this a measurement only of the specific allocDeferred vs alloc? I think for this specific use case, the latency hiding of the actual computation might still make the whole scan call faster overall.

Yes, it is the comparison between `allocDeferred` and `alloc`. Even with `allocDeferred`, there is no latency hiding, the call is using `cudaMallocAsync` which only means the memory will be available for kernel after the call within the stream, it is equal to the cpu side lazy allocation of memory on first touch. 
The big benefit is coming if you use multiple calls.

The question is now if we schould you by default `allocDefered` even if it is slow for the first call. In that case we need to add a dummy allocation to our scan examples else the performance will be very bad. Additionally, we need to add a doc string to scan which points to this behaviour.

SimeonEhrig: I think we have 3 options.
1. Keep alloc -> did you meassured performance benefits on other platforms like CPU, ROCm or Sycl?
2. hard code specialization for CUDA, which uses alloc instead allocDefered (maybe via `if constexpr`)
3. add policy to control it and set useful defaults per backend

psychocoderHPC: > 1. Keep alloc -> did you meassured performance benefits on other platforms like CPU, ROCm or Sycl?

No I checke only CUDA. For CPU or Sycl there can not be any difference because it simply maps to normal malloc calls. 

> 2. hard code specialization for CUDA, which uses alloc instead allocDefered (maybe via if constexpr)

The point is in general `allocDefered` has lower latency but not for the first call. 

> 3. add policy to control it and set useful defaults per backend

Yes this would be an option to expose it to the user. We we would do this this will not be part of this PR.

So the real question is, should I switch to `allocDeferred` and extent the doc string and point to the first call overhead or should I keep `alloc` equal to what it is currently?

psychocoderHPC:
> > IMO for the algorithm scan we should provide the user the possibility to define which type of memory he prefers for the temporary required memory.
> 
> The overloads accepting the buffer memory already allow this, no?



> > IMO for the algorithm scan we should provide the user the possibility to define which type of memory he prefers for the temporary required memory.
> 
> The overloads accepting the buffer memory already allow this, no?

Thats a valid point, the user can already provide the memory of their choice, so no policy is required.

include/alpaka/onHost/algo/internal/scan.hpp
SimeonEhrig: We keep `alpaka::alloc`. Like @AntonReinhard mentioned, the user can manually control how the memory is allocated and pass it via argument.

In a offline discussion with @psychocoderHPC , he explained me that I misunderstand the usage of `alpaka::allocDerred`. I thought this is simply an `alpaka::alloc` without implicit `alpaka::wait`. But the function is designed to be fast in hot loop, means repeatedly allocate the same chunk of memory. Therefore we should assume that the default case is outside the hot loop, therefore the latency of the first allocation matters. If the user use it in a hot loop, the user should care about the memory allocation by itself and pass the temporary memory to the `scan` function.

@psychocoderHPC Please add a note the doc string, which describes the allocation behavoir and what the user should do to achive better performance.

psychocoderHPC:
I added the review comments and extended the documentation

tests/unit/mem/keepAlive.cpp
```diff
@@ -55,7 +55,7 @@ TEMPLATE_LIST_TEST_CASE("keep alive", "", TestApis)
 
     {
         // enqueue everything in this scope
-        auto scopedBuffer = onHost::alloc<int>(device, N);
+        auto scopedBuffer = onHost::allocDeferred<int>(queue, N);
```
psychocoderHPC, line 58: This reverts #448, #448
 was added to avoid random CI crashes due to the known issue

SimeonEhrig (approved):

PR #451, opened 2026-02-25
introduce concept IndexVec

SimeonEhrig:
Check whether the specified type is a multidimensional index. The dimension needs to match the expected dim and the index type needs to be convertible without precision lost.

I need to check, if we can use the concept somewhere else in the code. Ether I do it in this PR or in a following up PR.

### DIFF
--- a/tests/unit/precision/precision.cpp
+++ b/tests/unit/precision/precision.cpp
@@ -25,16 +25,12 @@ using namespace alpaka::onHost;
 
 using TestApis = std::decay_t<decltype(allBackends(enabledApis, exec::enabledExecutors))>;
 
-template<typename T_LoopIdxType>
+template<typename>
 struct IotaKernelND
 {
     ALPAKA_FN_ACC void operator()(auto const& acc, auto out) const
     {
         using MemScalarType = typename alpaka::trait::GetValueType_t<ALPAKA_TYPEOF(out)>::type;
-        for(auto i : onAcc::makeIdxMap<T_LoopIdxType>(acc, onAcc::worker::threadsInGrid, IdxRange{out.getExtents()}))
-        {
-            out[i] = pCast<MemScalarType>(i);
-        }
 
         // rerun the tests with automatic derived loop index type
         for(auto i : onAcc::makeIdxMap(acc, onAcc::worker::threadsInGrid, IdxRange{out.getExtents()}))
@@ -44,7 +40,7 @@ struct IotaKernelND
     }
 };
 
-template<typename T_MemIdxType, typename T_LoopIdxType>
+template<typename T_MemIdxType, typename T_LoopIdxType, template<typename> class T_Kernel>
 void iotaTest(auto& queue, auto exec, auto const extents, auto frameSize)
 {
     auto dBuff = onHost::alloc<Vec<T_MemIdxType, ALPAKA_TYPEOF(extents)::dim()>>(queue.getDevice(), extents);
@@ -52,11 +48,13 @@ void iotaTest(auto& queue, auto exec, auto const extents, auto frameSize)
 
     using KenelIdxScalarType = typename ALPAKA_TYPEOF(frameSize)::type;
 
+    T_Kernel<T_LoopIdxType> kernel;
+
     onHost::wait(queue);
     queue.enqueue(
         exec,
         FrameSpec{pCast<KenelIdxScalarType>(extents) / frameSize, frameSize},
-        KernelBundle{IotaKernelND<T_LoopIdxType>{}, dBuff});
+        KernelBundle{kernel, dBuff});
     onHost::memcpy(queue, hBuff, dBuff);
     onHost::wait(queue);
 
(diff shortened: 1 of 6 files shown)


SimeonEhrig:
I opened a bug report for the [segmentation fault problem](https://github.com/alpaka-group/alpaka3/pull/451/changes/2e4bdf18a677c1f917dafa30d5679e74874a73c5): https://developer.nvidia.com/bugs/5930490

SimeonEhrig:
> I opened a bug report for the [segmentation fault problem](https://github.com/alpaka-group/alpaka3/pull/451/changes/2e4bdf18a677c1f917dafa30d5679e74874a73c5): https://developer.nvidia.com/bugs/5930490

I got an answer. The bug is already fixed in nvcc 13.2-rc

tests/unit/precision/precision.cpp
```diff
@@ -25,16 +25,12 @@ using namespace alpaka::onHost;
 
 using TestApis = std::decay_t<decltype(allBackends(enabledApis, exec::enabledExecutors))>;
 
-template<typename T_LoopIdxType>
+template<typename>
```
psychocoderHPC: please remove the template

tests/unit/precision/precision.cpp
```diff
@@ -44,19 +40,21 @@ struct IotaKernelND
     }
 };
 
-template<typename T_MemIdxType, typename T_LoopIdxType>
+template<typename T_MemIdxType, typename T_LoopIdxType, template<typename> class T_Kernel>
```
psychocoderHPC: `template<typename> class T_Kernel` -> `class T_Kernel` but you can also path the kernel via argument than there is no need for a template parameter

tests/unit/precision/precision.cpp
```diff
@@ -44,19 +40,21 @@ struct IotaKernelND
...
 void iotaTest(auto& queue, auto exec, auto const extents, auto frameSize)
 {
     auto dBuff = onHost::alloc<Vec<T_MemIdxType, ALPAKA_TYPEOF(extents)::dim()>>(queue.getDevice(), extents);
     auto hBuff = onHost::allocHostLike(dBuff);
 
     using KenelIdxScalarType = typename ALPAKA_TYPEOF(frameSize)::type;
 
+    T_Kernel<T_LoopIdxType> kernel;
```
psychocoderHPC: `<T_LoopIdxType>` not required if the kernel is passed as full type

tests/unit/precision/precision.cpp
```diff
@@ -202,14 +262,90 @@ void callTests(auto cfg)
...
+    }
+
+    Device device = devSelector.makeDevice(0);
+    Queue queue = device.makeQueue();
+
+    INFO(
+        "device=" << device.getName() << " api=" << deviceSpec.getApi().getName()
+                  << " exec=" << onHost::demangledName(exec));
+
+    iotaTest<T_MemIdx, T_LoopIdx, IotaKernelNDForceCast>(
```
psychocoderHPC: please path always the full kernel signature 

```C++
iotaTest<T_MemIdx, IotaKernelNDForceCast<T_LoopIdx>>(
```

tests/unit/precision/precision.cpp
psychocoderHPC: `typename T_LoopIdxType` is not required after the change

tests/unit/precision/precision.cpp
```diff
@@ -74,27 +72,33 @@ void callTests(auto cfg)
...
     Queue queue = device.makeQueue();
 
-    std::cout << "exec=" << onHost::demangledName(exec) << std::endl;
+    INFO(
+        "device=" << device.getName() << " api=" << deviceSpec.getApi().getName()
+                  << " exec=" << onHost::demangledName(exec));
 
     {
         using MemIdxType = std::conditional_t<T_signedMemIdx, std::make_signed_t<uint32_t>, uint32_t>;
         using LoopIdxType = std::conditional_t<T_signedLoopIdx, std::make_signed_t<uint32_t>, uint32_t>;
```
psychocoderHPC: `LoopIdxType` definition can be removed.

tests/unit/precision/precision.cpp
```diff
@@ -202,14 +262,90 @@ void callTests(auto cfg)
 
 TEMPLATE_LIST_TEST_CASE("kernel_precision_iotaNd", "", TestApis)
 {
+    constexpr bool SIGNED = false;
```
psychocoderHPC: I like the named definitions more than `true` `false`

but maybe 

```
    enum class Sign : bool
    {
        SIGNED = true,
        UNSIGNED = false
    };
```

would allow to change

```
template<Sign T_signedMemIdx, Sign T_signedLoopIdx, Sign T_signedKernelIdx>
void callTests(auto cfg)
```

psychocoderHPC (changes requested):
Could you please remove the template template signature of the kernel, this is making clear that one kernel is not taking the loop index type into account.

tests/unit/precision/precision.cpp
psychocoderHPC: not e if you use clas enums you must compare them like 

```C++
if constexpr (T_signedMemIdx == Sign::SIGNED)
```

SimeonEhrig: I used a normal enum, because the values can be directly cast to bool

psychocoderHPC (approved):

PR #406, opened 2025-12-19
vendor API abstraction

psychocoderHPC:
This PR is a prove of concept if we could with minimal overhead integrate vendor functions.

- use scan as example

```C++
auto scanFn = vendor::onHost::scanFn(dev);
if constexpr(scanFn)
{
    size_t tmpBufferSize = scanFn.getBufferSize(queue, scanFn.exclusive, bufX, bufY);
    auto tmp = onHost::alloc<char>(dev, tmpBufferSize);
    scanFn(queue, scanFn.exclusive, tmp, bufX, bufY);
    alpaka::onHost::wait(queue);
}
```

note: The function call `getBufferSize()` is currently blocking. I was not sure how to design the interface to the vendor function, if I should expose the full C interface or if I should allow passing alpaka objects.
This is only a draft anf should show one possible way and provides a platform for discussions.

please ignore the CI I have only checked the scan example.

PR #452, opened 2026-02-26
add alpaka function interface

psychocoderHPC:
This PR provides an alternative interface compared to #406

Provide an interface to dispatch function calls to third party function interfaces. This can be used to call functionality that is not available in alpaka or in cases where e.g. the vendor implementation provides better performance. 
alpaka's function interface provides fallback possibilities to user implemented generic implemenations to ensure the portability of the user application.
alpaka's function interface can be used onHost and onAcc.
A function symbol can be used as a kernel within alpaka too, which simplifies mixing user-optimized native CUDA code and alpaka kernel.

- implement function interface
- provide an example on how to use the interface
- introduce a concept `concepts::DeviceSpec`

Note: there is currently no example available where the function interface is used as kernel, this is coming soon ... in a separate PR.


psychocoderHPC:
@chillenzer The CI shows green

example/vendorApi/src/main.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+    // Create a queue on the device
+    onHost::Queue queue = devAcc.makeQueue();
+
+    // Allocate 3 host memory buffers
+    auto iotaBuffer_h = onHost::allocHost<Data>(extent);
+    auto output_h = onHost::allocHostLike(iotaBuffer_h);
+
+    // C++14 random generator for uniformly distributed numbers in {1,..,42}
+    std::random_device rd{};
```
psychocoderHPC: should be removed

SimeonEhrig: Do you mean replacing with something like `alpaka::onHost::random()`?

example/vendorApi/src/main.cpp
```diff
@@ -0,0 +1,185 @@
...
+    auto input0_d = onHost::allocLike(devAcc, iotaBuffer_h);
+    auto input1_d = onHost::allocLike(devAcc, iotaBuffer_h);
+    auto output_d = onHost::allocLike(devAcc, iotaBuffer_h);
+
+    // copy input data to device and clear device output buffer
+    onHost::memcpy(queue, input0_d, iotaBuffer_h);
+    onHost::memcpy(queue, input1_d, iotaBuffer_h);
+    onHost::fill(queue, output_d, Data{0});
+
+    vendorExample::Transform::call(queue, output_d, std::plus{}, input0_d, input1_d);
```
SimeonEhrig, line 105: What is the idea here? I would expect that I call `alpaka::onHost::transform` and the function calls hidden an optimized version of transform (e.g. from thrust lib), which I registered.

Here, if I replace the function call with my own function, why I need the complex replace mechanism? I could simply call my optimized version and implement my own easy fallback algorithm.

include/alpaka/concepts.hpp
```diff
@@ -50,6 +50,13 @@ namespace alpaka
          */
         template<typename T>
         concept Pointer = std::is_pointer_v<T>;
+
+        /** @brief Concept to check that a device specification with an API and device kind can be extracted. */
+        template<typename T>
+        concept DeviceSpec = requires(T t) {
+            { internal::getApi(t) } -> alpaka::concepts::Api;
+            { internal::getDeviceKind(t) } -> alpaka::concepts::DeviceKind;
```
SimeonEhrig, line 58: I think, this will introduce a circular dependency. An interface concept should define member functions and never, which traits it does specialize.

example/vendorApi/src/main.cpp
psychocoderHPC: random is not used in this example since we use iota to initialize the data

example/vendorApi/src/main.cpp
psychocoderHPC: > I would expect that I call alpaka::onHost::transform and the function calls hidden an optimized version of transform (e.g. from thrust lib), which I registered.

This is a way we have been taught about in the past, but it will most likely be hard to maintain. There will be third party libraries we can not support but the user would like to use. This new approach is giving the user maximum flexibility.

>  I could simply call my optimized version and implement my own easy fallback algorithm.

Calling the optimized version means you need to guard the cases, maybe by `#ifdef HAVE _OPTIMIZED_VERSION` and this makes the user code more complicated.

include/alpaka/concepts.hpp
psychocoderHPC: At the moment, there are no circular dependencies because I use `internal` and not the user interface `alpaka::getDeviceKind()`. `internal::getDeviceKind(t)` later maps to the trait.

psychocoderHPC:
@chillenzer and @ikbuibui  I updated the implementation. My last changes are in the last commit, but I suggest reviewing the complete changes at once.

example/vendorApi/src/cudaFn.hpp
```diff
@@ -0,0 +1,50 @@
...
+            thrust::cuda::par.on(queue.getNativeHandle()),
+            input0.data(),
+            input0.data() + input0.getExtents().x(),
+            input1.data(),
+            output.data(),
+            binaryOp);
+    }
+
+    /** @} */
+#endif
```
ikbuibui: You can move this endif after the end of the namespace and completely remove defining and using `#if ALPAKA_EXAMPLE_HAS_THRUST`. The whole thing can be guarded by `if __has_include`

example/vendorApi/src/fn.hpp
```diff
@@ -0,0 +1,67 @@
...
+        queue.enqueueHostFn(
+            [=]()
+            {
+                std::transform(
+                    input0.data(),
+                    input0.data() + input0.getExtents().x(),
+                    input1.data(),
+                    outPtr,
+                    binaryOp);
+            });
```
ikbuibui, line 64:
```suggestion
        queue.enqueueHostFn(
            [in0 = input0.data(),
             extX = input0.getExtents().x(),
             in1 = input1.data(),
             out = output.data(),
             op = binaryOp]()
            {
                std::transform(
                    in0,
                    in0 + extX,
                    in1,
                    out,
                    op);
            });
```
Not a required change, just an optimization to only capture the necessary params and not the full mdspan objects etc.
Also should this `std::transform` use an execution policy?

include/alpaka/fn.hpp
```diff
@@ -0,0 +1,319 @@
...
+            return T_registrationPolicy == Registration::alwaysTrue
+                   || concepts::FnRegistered<ALPAKA_TYPEOF(spec(getApi(any), getDeviceKind(any)))>
+                   || ((T_fallbackPolicy != Fallback::none)
+                       && concepts::FnRegistered<ALPAKA_TYPEOF(spec(api::Alpaka{}, getDeviceKind(any)))>);
+        }
+
+        /** Call function overload if defined for the given device specification. */
+        template<alpaka::concepts::DeviceSpec T_Any, typename... Args>
+        requires concepts::DispatchedFnInvocable<
+            ALPAKA_TYPEOF(spec(getApi(std::declval<T_Any>()), getDeviceKind(std::declval<T_Any>()))),
```
ikbuibui: This is super long and repeated in many places, maybe something like helps readability 
```C++
template<alpaka::concepts::DeviceSpec T_Any>
using SpecFor = decltype(spec(getApi(std::declval<T_Any>()), getDeviceKind(std::declval<T_Any>())));
```

ikbuibui (commented):
No major comments. Things are okay. I opened a PR against your branch with some typo fixes.

example/vendorApi/src/fn.hpp
psychocoderHPC: I do not want to blow up the example, therefore I will stay with capturing by copy but you are right you suggestion is the safer way.
I do not use the execution policy of std::transform because with older compilers, we will get many warnings that e.g. the policy parallel is not implemented. Since this example is using small buffers and is not about performance, it should be fine. The key message of this example is to show how you can dispatch to third-party APIs.

include/alpaka/fn.hpp
```diff
@@ -0,0 +1,311 @@
...
+            return typename T_FnClass::template Spec<ALPAKA_TYPEOF(getApi(any)), ALPAKA_TYPEOF(getDeviceKind(any))>{};
+        }
+
+        template<alpaka::concepts::Api T_Api, alpaka::concepts::DeviceKind T_DeviceKind>
+        static constexpr auto spec(T_Api api, T_DeviceKind deviceKind)
+        {
+            alpaka::unused(api, deviceKind);
+            return typename T_FnClass::template Spec<T_Api, T_DeviceKind>{};
+        }
+
```
chillenzer, line 173: Why doesn't the first internally call the second? It's two places to change in case anything changes in how that function is defined. Not a showstopper because it's currently easy to see that both must be changed but surprised to not find the familiar pattern.

include/alpaka/fn.hpp
```diff
@@ -0,0 +1,311 @@
...
+         *
+         * See the call operator().
+         * @attention call() is a static function where the call operator required an instance of this class.
+         */
+        static constexpr decltype(auto) call(alpaka::concepts::DeviceSpec auto&& any, auto&&... args)
+        {
+            static_assert(
+                std::is_trivially_constructible_v<T_FnClass>,
+                "Function class must be trivially constructible to use call().");
+            T_FnClass{}(ALPAKA_FORWARD(any), ALPAKA_FORWARD(args)...);
```
chillenzer:
```suggestion
            return T_FnClass{}(ALPAKA_FORWARD(any), ALPAKA_FORWARD(args)...);
```

chillenzer (changes requested):
Looks good. There are still a few typos in the docstrings but I'll open a separate PR for this (or you can run an agent over it yourself).

There's one missing `return` keyword that I feel still blocks this. Sorry.

include/alpaka/fn.hpp
psychocoderHPC: Because I readded the second later after I did a refactoring ask by @ikbuibui and than I have not seen it. In general the current version should be better from the point of compile time because it is creating one signarure less.

include/alpaka/fn.hpp
psychocoderHPC: ohh yes thanks for spotting the bug, I will fix it tomorrrow

include/alpaka/fn.hpp
chillenzer: That sounds like a premature optimization. As I said, it's fine with me to not do it right now but without a compile-time benchmark showing that it really hurts, I won't buy that.

psychocoderHPC: I will update it with the missing return statement together

include/alpaka/fn.hpp
psychocoderHPC: updated

include/alpaka/fn.hpp
psychocoderHPC: updated

include/alpaka/fn.hpp
ikbuibui: I thought this was on purpose, because you need kernels to return void :sweat_smile:

chillenzer (approved):
Looks good!

PR #453, opened 2026-03-02
fixed `index_type` of `MdSpanArray`

SimeonEhrig:
The index_type was hard coded to `size_t`, even an extents which was used to construct it had a different `value_type`. No `MdSpanArray` has the same `index_type` like the `extents`.

If the function `alpaka::onAcc::declareSharedMdArray(acc, extents);` is used, the get shared memory, the index type of the extents was ignored. Instead the index type of the SharedMdArray was `size_t` all the time. This caused problems, when I added the concept `alpaka::concepts::IndexVec` to the access operator of `alpaka::SharedMdArray`.

The following code throws an compiler error without fix because of potential precession lost:

```c++
auto m = alpaka::onAcc::declareSharedMdArray(acc, alpaka::CVec<int, 1>{});
// 0 is an `int`, therefore it can be negative
// the index_type of m is `size_t` 
m[0];
```

# Site note

I added CMake test target for the unit tests, that the macros `REQUIRE_MESSAGE` and `CHECK_MESSAGE` are available in each test case.

### DIFF
--- a/include/alpaka/mem/MdSpanArray.hpp
+++ b/include/alpaka/mem/MdSpanArray.hpp
@@ -14,6 +14,7 @@
 #include "alpaka/mem/concepts/detail/InnerTypeAllowedCast.hpp"
 #include "alpaka/mem/trait.hpp"
 #include "alpaka/trait.hpp"
+#include "concepts/IndexVec.hpp"
 
 #include <concepts>
 #include <type_traits>
@@ -53,7 +54,7 @@ namespace alpaka
         using type = T[T_Extents{}[0u]];
     };
 
-    template<typename T_ArrayType, concepts::Alignment T_MemAlignment = Alignment<>>
+    template<typename T_ArrayType, std::integral T_IndexType, concepts::Alignment T_MemAlignment = Alignment<>>
     struct MdSpanArray
     {
         static_assert(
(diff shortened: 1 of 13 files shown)


include/alpaka/mem/MdSpanArray.hpp
```diff
@@ -201,9 +208,9 @@ namespace alpaka
 
         constexpr auto getExtents() const
         {
-            auto const createExtents = []<auto... T_extent>(std::index_sequence<T_extent...>)
+            auto const createExtents = []<uint32_t... T_extent>(std::integer_sequence<uint32_t, T_extent...>)
```
psychocoderHPC, line 210: should this not be `index_type` instead of `uint32_t`?

include/alpaka/mem/MdSpanArray.hpp
```diff
@@ -201,9 +208,9 @@ namespace alpaka
...
             { return CVec<index_type, std::extent_v<T_ArrayType, T_extent>...>{}; };
-            return createExtents(std::make_integer_sequence<index_type, dim()>{});
+            return createExtents(std::make_integer_sequence<uint32_t, dim()>{});
```
psychocoderHPC, line 212: `index_type` instead of `uint32_t`

include/alpaka/onAcc/internal/globalMem.hpp
```diff
@@ -213,7 +213,9 @@ namespace alpaka::onAcc::internal
 
         constexpr decltype(auto) get() const requires(std::is_array_v<type>)
         {
-            return alpaka::MdSpanArray<type>{T_Storage::get(thisApi())};
+            // std::array uses size_t for the access operator
+            // a static array of type C also uses size_t to define its length
+            return alpaka::MdSpanArray<type, size_t>{T_Storage::get(thisApi())};
```
psychocoderHPC, line 217: could you please change this to `size_t` to `uint32_t`. Since we have now the possibility to control the precision it should be small to reduce register usage for index computations. Shared memory is always small and fits into 32bit.

psychocoderHPC (changes requested):
Thanks for the fix, there are only 2 small points

include/alpaka/mem/MdSpanArray.hpp
SimeonEhrig: `uint32_t` is the type of the dim

include/alpaka/onAcc/internal/globalMem.hpp
psychocoderHPC: this is wrong, it should be `size_t` to allow `uint32_t` as index type for the `operator[]`.

psychocoderHPC (approved):

PR #458, opened 2026-03-09
add missing `Queue::enqueue()` signature

psychocoderHPC:
- add enqueue signature without an executor
- update and add few documenation strings

### DIFF
--- a/include/alpaka/onHost/Queue.hpp
+++ b/include/alpaka/onHost/Queue.hpp
@@ -108,9 +108,9 @@ namespace alpaka::onHost
             auto const& f,
             auto&&... args) const
         {
-            return internal::enqueue(
+            internal::enqueue(
                 *m_queue.get(),
-                std::move(executor),
+                executor,
                 domainSpec,
                 KernelBundle{f, onHost::makeAccessibleOnAcc(ALPAKA_FORWARD(args))...});
         }
@@ -127,10 +127,10 @@ namespace alpaka::onHost
          */
         void enqueue(onHost::concepts::ThreadOrFrameSpec auto const& domainSpec, auto const& f, auto&&... args) const
         {
-            auto executor = supportedExecutors(internal::getDevice(*m_queue.get()), exec::allExecutors);
-            return internal::enqueue(
+            auto executors = supportedExecutors(internal::getDevice(*m_queue.get()), exec::allExecutors);
+            internal::enqueue(
                 *m_queue.get(),
-                std::get<0>(executor),
+                std::get<0>(executors),
                 domainSpec,
                 KernelBundle{f, onHost::makeAccessibleOnAcc(ALPAKA_FORWARD(args))...});
         }
@@ -149,8 +149,8 @@ namespace alpaka::onHost
             onHost::concepts::ThreadOrFrameSpec auto const& domainSpec,
             KernelBundle<TKernelFn, TArgs...> const& kernelBundle) const
         {
-            auto executor = supportedExecutors(internal::getDevice(*m_queue.get()), exec::allExecutors);
-            internal::enqueue(*m_queue.get(), std::get<0>(executor), domainSpec, kernelBundle);
+            auto executors = supportedExecutors(internal::getDevice(*m_queue.get()), exec::allExecutors);
+            internal::enqueue(*m_queue.get(), std::get<0>(executors), domainSpec, kernelBundle);
         }
 
         /** Enqueue a kernel functor to a queue.
@@ -179,9 +179,7 @@ namespace alpaka::onHost
          */
         void enqueueHostFn(auto const& task) const
         {
-            return internal::Enqueue::HostTask<ALPAKA_TYPEOF(*m_queue.get()), ALPAKA_TYPEOF(task)>{}(
-                *m_queue.get(),
-                task);
+            internal::Enqueue::HostTask<ALPAKA_TYPEOF(*m_queue.get()), ALPAKA_TYPEOF(task)>{}(*m_queue.get(), task);
         }
 
         /** Enqueue an operation which is executed asynchronously on the host side
@@ -194,7 +192,7 @@ namespace alpaka::onHost
          */
         void enqueueHostFnDeferred(auto const& task) const
         {
-            return internal::Enqueue::HostTaskDeferred<ALPAKA_TYPEOF(*m_queue.get()), ALPAKA_TYPEOF(task)>{}(
+            internal::Enqueue::HostTaskDeferred<ALPAKA_TYPEOF(*m_queue.get()), ALPAKA_TYPEOF(task)>{}(
                 *m_queue.get(),
                 task);
         }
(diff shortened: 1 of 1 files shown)


SimeonEhrig (approved):

SimeonEhrig:
@ikbuibui René and I cannot review anymore, because we pushed both :grimacing:

psychocoderHPC:
or @AntonReinhard needs to take over the merge.

include/alpaka/onHost/Queue.hpp
```diff
@@ -84,73 +84,98 @@ namespace alpaka::onHost
...
         void enqueue(
             auto const executor,
-            onHost::concepts::ThreadOrFrameSpec auto const& blockCfg,
+            onHost::concepts::ThreadOrFrameSpec auto const& domainSpec,
             auto const& f,
             auto&&... args) const
         {
             return internal::enqueue(
                 *m_queue.get(),
                 std::move(executor),
```
ikbuibui: Not in this PR, but you are moving a const object. This is simply a copy

include/alpaka/onHost/Queue.hpp
```diff
@@ -84,73 +84,98 @@ namespace alpaka::onHost
...
+         * @param domainSpec Thread or frame specification which provides a chunked description of the thread or frame
+         * index domain.
+         * @param f The compute kernel functor.
+         * @param args Arguments passed to the kernel functor.
          */
         void enqueue(
             auto const executor,
-            onHost::concepts::ThreadOrFrameSpec auto const& blockCfg,
+            onHost::concepts::ThreadOrFrameSpec auto const& domainSpec,
             auto const& f,
```
ikbuibui, line 108:
```suggestion
            auto && f,
```
and then forwarding it below is better since `f` may be a heavy lambda with captures and currently the code creates a copy

include/alpaka/onHost/Queue.hpp
```diff
@@ -84,73 +84,98 @@ namespace alpaka::onHost
...
          *
-         * @param specification thread or frame specification which provides a chunked description of the thread or
-         * frame index domain
-         * @param kernelBundle the compute kernel and there arguments
+         * @param domainSpec Thread or frame specification which provides a chunked description of the thread or frame
+         * index domain.
+         * @param f The compute kernel functor.
+         * @param args Arguments passed to the kernel functor.
          */
+        void enqueue(onHost::concepts::ThreadOrFrameSpec auto const& domainSpec, auto const& f, auto&&... args) const
```
ikbuibui, line 128: same as above `auto&& f`

include/alpaka/onHost/Queue.hpp
```diff
@@ -84,73 +84,98 @@ namespace alpaka::onHost
...
+        {
+            auto executor = supportedExecutors(internal::getDevice(*m_queue.get()), exec::allExecutors);
+            return internal::enqueue(
+                *m_queue.get(),
+                std::get<0>(executor),
```
ikbuibui:
```suggestion
            auto executors = supportedExecutors(internal::getDevice(*m_queue.get()), exec::allExecutors);
            return internal::enqueue(
                *m_queue.get(),
                std::get<0>(executors),
```
and the same below, so that we know it is a tuple of executors.
also the comment says that ``The default executor is the executor with the highest parallelism/performance.``  Is it guaranteed that the `get<0>(executors)` will be the one with the highest parallelism?

include/alpaka/onHost/Queue.hpp
```diff
@@ -174,13 +199,23 @@ namespace alpaka::onHost
...
         {
             return internal::Enqueue::Event<ALPAKA_TYPEOF(*m_queue.get()), ALPAKA_TYPEOF(*event.get())>{}(
                 *m_queue.get(),
                 *event.get());
         }
 
+        /** Wait until all operations in this queue are finished.
+         *
+         * The caller will be blocked until all previously queued operations have been completed.
+         */
```
ikbuibui, line 216: Is this correct? Is this blocking the current thread on a event or synchronizing the queue to wait on an event?

include/alpaka/onHost/Queue.hpp
```diff
@@ -174,13 +199,23 @@ namespace alpaka::onHost
...
 
+        /** Enqueue an event
+         *
+         * The event will be signaled after all preceding operations in the queue are finished.
+         *
+         * @param event Event that is to be enqueue in the queue of operations.
+         */
         void enqueue(Event<Device<T_Api, T_DeviceKind>> const& event) const
         {
             return internal::Enqueue::Event<ALPAKA_TYPEOF(*m_queue.get()), ALPAKA_TYPEOF(*event.get())>{}(
```
ikbuibui: Returning in a void function. Can we not skip this, otherwise I am tempted to check to ensure that the called function returns void as well.
Same thing below in `waitFor`

include/alpaka/onHost/Queue.hpp
```diff
@@ -84,73 +84,98 @@ namespace alpaka::onHost
...
+         * @param args Arguments passed to the kernel functor.
          */
+        void enqueue(onHost::concepts::ThreadOrFrameSpec auto const& domainSpec, auto const& f, auto&&... args) const
+        {
+            auto executor = supportedExecutors(internal::getDevice(*m_queue.get()), exec::allExecutors);
+            return internal::enqueue(
+                *m_queue.get(),
+                std::get<0>(executor),
+                domainSpec,
+                KernelBundle{f, onHost::makeAccessibleOnAcc(ALPAKA_FORWARD(args))...});
```
ikbuibui: Could also simply call the enqueue function defined above.

include/alpaka/onHost/Queue.hpp
psychocoderHPC: This is by intention, we capture it by reference and perform the copy as late as possible.
In this case, during the creation of the kernel bundle where we need to copy anyway.

include/alpaka/onHost/Queue.hpp
psychocoderHPC: ahh thank, I will change it in this PR, it is anyway a tag which should always be copied.

include/alpaka/onHost/Queue.hpp
psychocoderHPC: same as above, it is by itention

include/alpaka/onHost/Queue.hpp
psychocoderHPC: Yes it is documented above `allExecutors`

```
    /** list of all executors supported by alpaka
     *
     * The order is from high parallelism to low parallelism for executors which are falling into the same category.
     * This list is used at places where a function can be called without an executor. In this case the first available
     * executor is used.
     */
```

include/alpaka/onHost/Queue.hpp
psychocoderHPC: Yes, it blocks the current caller thread until the event is processed.

include/alpaka/onHost/Queue.hpp
psychocoderHPC: removed it, saw it in the past too and yes thats bad.

include/alpaka/onHost/Queue.hpp
psychocoderHPC: I will keep it at the moment as it is because all other function interfaces call internal implementations too. Changing only this is hard to spot if we were hunting bugs.

PR #459, opened 2026-03-10
fix `CallbackQueue`

psychocoderHPC:
follow-up of #448

The reason for this PR is that CI showed the error `ERROR in ~CallbackThread: internal queue is not empty, but object is destroyed.` because the last task is still in the queue when the thread is destroyed. This is fine because the tasks finished already, but are not removed from the internal queue at this moment.

Due to the design that two classes with different lifetimes use the same shared data for the queue implementation, we sometimes run into the case where the task executor thread destroyed the callback class, but the last task was not finished.
By separating the shared data, we now have a clear live-time management.

Note: The implementation could most likely be refactored to become nicer, but we messed up with the implementation in the past already, to avoid introducing larger bugs, the changes were kept small and focused.

### DIFF
--- a/include/alpaka/core/CallbackThread.hpp
+++ b/include/alpaka/core/CallbackThread.hpp
@@ -59,27 +59,31 @@ namespace alpaka::core
 
         using TaskPackage = std::pair<std::unique_ptr<Task>, std::promise<void>>;
 
+        struct State
+        {
+            std::queue<TaskPackage> m_tasks;
+            std::mutex m_mutex;
+            std::condition_variable m_cond;
+        };
+
     public:
+        CallbackThread() : m_state(std::make_shared<State>())
+        {
+        }
+
         ~CallbackThread()
         {
             {
-                std::unique_lock<std::mutex> lock{m_mutex};
+                std::unique_lock<std::mutex> lock{m_state->m_mutex};
                 m_thread.request_stop();
                 // wakeup the thread in case it is waiting
-                m_cond.notify_one();
+                m_state->m_cond.notify_one();
             }
 
             if(m_thread.joinable())
             {
                 if(std::this_thread::get_id() == m_thread.get_id())
                 {
-                    // at this point there should be no tasks in the queue
-                    if(!m_tasks.empty())
-                    {
-                        std::cerr << "ERROR in ~CallbackThread: internal queue is not empty but object is destructed."
-                                  << std::endl;
-                        std::abort();
-                    }
                     /* We can not join ourselves.
                      * We can only end here if a task that the callback thread is executing is capturing the object
                      * which is holding the callback thread.
@@ -92,7 +96,6 @@ namespace alpaka::core
         }
 
         //! It is guaranteed that the task is fully destroyed before the future's result is set.
-        //! @{
         template<typename NullaryFunction>
         auto submit(NullaryFunction&& nf) -> std::future<void>
         {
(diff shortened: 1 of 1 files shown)


include/alpaka/core/CallbackThread.hpp
```diff
@@ -104,41 +107,30 @@ namespace alpaka::core
...
-        std::queue<TaskPackage> m_tasks;
+        /** Hold data shared between this call and the thread processing the tasts. */
+        std::shared_ptr<State> m_state;
 
         auto startWorkerThread() -> void
         {
+            auto queueState = m_state;
             m_thread = std::jthread(
-                [this](std::stop_token st)
+                [state = std::move(queueState)](std::stop_token st)
```
SimeonEhrig:
```suggestion
                [state = m_state](std::stop_token st)
```
Why not directly copy the value?

include/alpaka/core/CallbackThread.hpp
```diff
@@ -59,27 +59,31 @@ namespace alpaka::core
...
-                m_cond.notify_one();
+                m_state->m_cond.notify_one();
             }
 
             if(m_thread.joinable())
             {
                 if(std::this_thread::get_id() == m_thread.get_id())
                 {
-                    // at this point there should be no tasks in the queue
-                    if(!m_tasks.empty())
```
SimeonEhrig, line 77: Is it save to remove the code completely? At the moment, it detects a bug.

include/alpaka/core/CallbackThread.hpp
psychocoderHPC: This is not possible in c++, it must be a variable in the function scope. I can capture by `[=]` but want to be explicit and show that I make a copy of the object.

include/alpaka/core/CallbackThread.hpp
psychocoderHPC: yes save, the worker is not stopping before all tasks are executed. While we are consume with the thread a task from the queue we move the task out of the queue and empty the leftover invalid object after the task is destroyed. This means the queue is not empty but the last task is invalid.
By separating the shared data, we do not have this problem anymore, we only need to guarantee that the thread is not stopping before it has consumed all tasks, this is done in the thread loop itself.

include/alpaka/core/CallbackThread.hpp
SimeonEhrig: I don't understand it. I tested the change. It is working. Instead creating a copy and move it to the lambda, I suggest to directly use the copy. Therefore the move is simply avoided.

psychocoderHPC: it is not allowed

```C++
error: capture of non-variable ‘alpaka::core::CallbackThread::m_state’
  132 |                 [m_state](std::stop_token st)
```

psychocoderHPC:
```C++
auto startWorkerThread() -> void
        {
            m_thread = std::jthread(
                [m_state](std::stop_token st)
                {
                    while(true)
                    {
                        std::promise<void> taskPromise;
                        std::exception_ptr eptr;
//...
```

psychocoderHPC: ahh I missed that you wrote `state = m_state`, I tested direct capturing. Ok yours is better!

psychocoderHPC: I updated it

SimeonEhrig (approved):

PR #461, opened 2026-03-11
document idxMap

SimeonEhrig:


SimeonEhrig:
Close because nobody can explain frame, frame extents and frame spec in a reasonable time. Therefore also not `makeMapIdx()`.

PR #463, opened 2026-03-13
optimize CMake

psychocoderHPC:
When using `alpaka_finalize()` it was searching for each target for `atomic_ref` and SIMD support.

Guard the check to be performed only once to speedup CMake.
Most likely we need to guard any places where we update the alpaka targets. This would require more testing first; therefore, currently, only known working places are guarded.


psychocoderHPC:
@SimeonEhrig PR is ready for a review

SimeonEhrig (approved):

PR #472, opened 2026-03-24
disable fast math in the CMake presets

SimeonEhrig:
- fast math let some tests fail


psychocoderHPC:
discussed offline: we should fix the test and use `approximately_equal` to compare floating point numbers and not `==` in the SIMD test.

SimeonEhrig:
not required anymore because of PR #481

PR #474, opened 2026-03-25
mem ordered fences

ikbuibui:
Includes inline assembly fallbacks for CUDA if `<cuda/atomic>` isn't available.

Compared to mainline alpaka it isn't using `omp flush`. It directly uses the `std::atomic_thread_fence`.

Adds a concept for scope, similar to the one introduced for memory order.

I am not sure about the order of parameters in the fence (and in the future the atomic) interface. Most interfaces (STL, CUDA, HIP, SYCL) are written as `(order, scope)` but we use `(scope,order)`

I currently haven't added any default orderings for fences


psychocoderHPC:
> I am not sure about the order of parameters in the fence (and in the future the atomic) interface. Most interfaces (STL, CUDA, HIP, SYCL) are written as (order, scope) but we use (scope,order)

For now we keep the order, let us discuss it in the next developer VC.

include/alpaka/api/host/memoryOrder.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include <atomic>
+#include <type_traits>
+
+#ifdef ALPAKA_DISABLE_STD_ATOMIC_REF
+#    include <boost/atomic.hpp>
+#endif
+
+namespace alpaka::onAcc::internalCompute
+{
+#if defined(ALPAKA_DISABLE_STD_ATOMIC_REF)
```
psychocoderHPC: Could you please remeove one macro block and add

```C++
#if defined(ALPAKA_DISABLE_STD_ATOMIC_REF)
        namespace alpakaAtomic = boost;
#else
       namespace alpakaAtomic = std;
#endif

struct MemOrderHost
    {
        template<concepts::MemoryOrder TMemOrder>
        static constexpr auto get(TMemOrder const)
        {
            if constexpr(std::same_as<TMemOrder, order::SeqCst>)
            {
                return alpakaAtomic::memory_order::seq_cst;
            }
            if constexpr(std::same_as<TMemOrder, order::AcqRel>)
            {
                return alpakaAtomic::memory_order::acq_rel;
            }
            if constexpr(std::same_as<TMemOrder, order::Release>)
            {
                return alpakaAtomic::memory_order::release;
            }
            if constexpr(std::same_as<TMemOrder, order::Acquire>)
            {
                return alpakaAtomic::memory_order_acquire;
            }
            if constexpr(std::same_as<TMemOrder, order::Relaxed>)
            {
                return alpakaAtomic::memory_order_relaxed;
            }
        }
    };
```

psychocoderHPC: note `MemOrderBoost` was not used at all

include/alpaka/api/host/memFence.hpp
```diff
@@ -25,66 +27,29 @@ namespace alpaka::onAcc::internalCompute
...
-        }
+            using ScopeT = std::remove_cvref_t<decltype(scope)>;
 
-        inline void hostMemoryFenceImpl(exec::CpuTbbBlocks const&, scope::System const)
-        {
-            std::atomic_thread_fence(std::memory_order_acq_rel);
+            // Block scope requires no fence since threads within a block are simulated/single-threaded
+            if constexpr(!std::same_as<ScopeT, scope::Block>)
+            {
+                std::atomic_thread_fence(MemOrderStl::get(order));
```
psychocoderHPC: please use `MemOrderHost` to cover boost and std atomics

include/alpaka/api/host/memoryOrder.hpp
psychocoderHPC: Question: Are memory ordering depends on `std::atomic_ref` support? If not the boost fallback can be removed.

psychocoderHPC (changes requested):
small changes are required.

include/alpaka/api/host/memoryOrder.hpp
ikbuibui: Yes, I was being sneaky. It is not required for fences at all, but will be required for ordered atomics in the future. I will remove it for now and reintroduce it later.

psychocoderHPC (approved):

PR #481, opened 2026-04-02
implement approximately equal for alpaka::Simd in the unit tests

SimeonEhrig:


include/alpaka/SimdMask.hpp
```diff
@@ -92,6 +95,8 @@ namespace alpaka
          *
          * The generator must return the value for the corresponding index of the component which is passed to the
          * generator.
+         *
+         * @note The generator needs to have the function interface `bool generator(u_int32_t id)`.
```
psychocoderHPC: we are never using `u_int32_t`, is there a reason why you not use `uint32_t`

psychocoderHPC: The invoke us testing for `generator(std::integral_constant<uint32_t, INDEX>)` so this and for sure `generator(uint32_t)` is supported.

psychocoderHPC (changes requested):
if you can change `u_int32_t` it would be nice

include/alpaka/SimdMask.hpp
SimeonEhrig: > The invoke us testing for generator(std::integral_constant<uint32_t, INDEX>) so this and for sure generator(uint32_t) is supported.

I wrote it down, with `generator(auto id)`, the type of `id` is an wrapper around `uint_32_t`. So if you pass it for example to the access operator of an `alpaka::Simd`, you get an compiler error, because the wrapper does not fulfill the requirement `std::integral`.

SimeonEhrig: > we are never using u_int32_t, is there a reason why you not use uint32_t

done

psychocoderHPC (approved):

PR #483, opened 2026-04-03
fix function interface registration check

psychocoderHPC:
Fix the wrong implementation of the function interface registration check.

- introduce `hasRegisteredFallback()` to check for the fallback
  registration only
- add test cases, to guard against wrong behaviour
- add `getApi()` to `onHost::Device`

### DIFF
--- a/include/alpaka/fn.hpp
+++ b/include/alpaka/fn.hpp
@@ -176,7 +176,7 @@ namespace alpaka::fn
         /** Checks if a function overload is registered for the given device specification.
          *
          * You can use the result to optionally call the function overload and disable at compile time code sections
-         * similar to C** preprocessor guards.
+         * similar to C++ preprocessor guards.
          * @code
          * ALPAKA_FN_SYMBOL(Foo,alpaka::fn::Fallback::none, alpaka::fn::Registration::enforced);
          *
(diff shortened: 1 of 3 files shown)


psychocoderHPC:
~~I will keep it in draft for now, I need to clarify the intendet usage first.~~

psychocoderHPC:
@SimeonEhrig PR is ready

SimeonEhrig (approved):

SimeonEhrig:
For clarification, why the mechanism "registering" exist? Is it possible, that you cannot check, if a overload is add, if you call `ALPAKA_FN_SYMBOL`?

Means the following is not possible?

```c++
namespace vendorExample
{
    ALPAKA_FN_SYMBOL(Foo, alpaka::fn::Fallback::toAlpaka, alpaka::fn::Registration::enforced);

    constexpr void fnDispatch(
        Foo::Spec<alpaka::api::Host, alpaka::deviceKind::Cpu>, auto&& queue)
    {
    }
}

int main(){
  // ...
  static_assert(vendorExample::isRegistered(hostCpuDevice)); // does not work, because requires explicit registration
}
```

SimeonEhrig (approved):

psychocoderHPC:
> For clarification, why the mechanism "registering" exist? Is it possible, that you cannot check, if a overload is add, if you call `ALPAKA_FN_SYMBOL`?
> 
> Means the following is not possible?

Exactly we can not check if a signature is available if you not add all function paramaters. So the code you providid will fail because the symbol is not registered.
If you remove the enforced registration of the symbol you could check if you can call the method e.g. with `if constexp ( requires{})`.
The registration is an optional feature that you can enable/disable a complete code region based on the result at compile time. So If a symbol is not available for cuda you can skip all memory allocations which are maybe required to call the function.

You can have a look into https://github.com/alpaka-group/alpaka3/blob/0a4a410320efa4a7fc9ffa156c8ac3df388e3c06/example/scan/src/scan_executors.hpp#L200
The full code for the CUB preparation within the macro guard can be moved within a `constexpr if` where you check first if the overload for cuda is available and if not you do not call it.

Is this helping?

PR #488, opened 2026-04-09
Fix zero element iter

psychocoderHPC:
- fix `MdForwardIter` if the view range contains zero in at least one dimension
- extent subview test cases

### DIFF
--- a/test/unit/mem/subDataStorage.cpp
+++ b/test/unit/mem/subDataStorage.cpp
@@ -7,6 +7,9 @@
 #include <alpakaTest/testMacros.hpp>
 #include <catch2/catch_test_macros.hpp>
 
+#include <array>
+#include <type_traits>
+
 TEST_CASE("1D alpaka::View::getSubView function tests", "[mem][view][SubDataStorage]")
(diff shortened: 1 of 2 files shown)


test/unit/mem/subDataStorage.cpp
```diff
@@ -20,42 +23,56 @@ TEST_CASE("1D alpaka::View::getSubView function tests", "[mem][view][SubDataStor
...
+        REQUIRE(count == 0);
+    }
 }
 
 TEST_CASE("3D alpaka::View::getSubView function tests", "[mem][view][SubDataStorage]")
 {
     constexpr int z = 3;
     constexpr int y = 5;
     constexpr int x = 4;
-    alpaka::Vec total_extents{z, y, x};
```
psychocoderHPC, line 51: The diff is slightly larger because I fixed the wrong naming schema. We always use `camelCase`.
Since I did it at the end with search and replace, also lines that are not required for this PR get changes. sry.

SimeonEhrig (approved):

PR #489, opened 2026-04-09
fix `MdSpan` `::end()` and `::begin()` constness

psychocoderHPC:
- add `const` version for `end()` and `begin()`
- extent tests

### DIFF
--- a/test/unit/mem/mdIterator.cpp
+++ b/test/unit/mem/mdIterator.cpp
@@ -6,29 +6,125 @@
 
 #include <catch2/catch_test_macros.hpp>
 
+#include <array>
+#include <type_traits>
+#include <utility>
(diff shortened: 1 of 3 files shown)


test/unit/mem/mdIterator.cpp
```diff
@@ -6,29 +6,125 @@
...
+        auto view = buffer.getView();
+
+        int nextValue = 10;
+        for(int& value : view)
+        {
+            value = nextValue;
+            nextValue += 5;
+        }
+
+        REQUIRE(view[alpaka::Vec{0u, 0u}] == 10);
```
psychocoderHPC: should check buffer not view

SimeonEhrig (approved):

Issue #490, opened 2026-04-10
Using .cc file extensions causes language issues for CUDA/HIP backends


TimHanel00:
Renaming an entry source file from, for example, `main.cpp` to `main.cc` appears to cause language translation issues, especially for the CUDA and HIP APIs.

As a result, certain backends are no longer available at runtime.
I tested this with `vectorAdd` using `vectorAdd.cpp` as the entry point:
-> `Building CUDA object example/vectorAdd/CMakeFiles/vectorAdd.dir/...`
If the entry point is renamed to `vectorAdd.cc`, it is instead compiled only as a CXX object:
-> `Building CXX object example/vectorAdd/CMakeFiles/vectorAdd.dir/src/vectorAdd.cc.o`
I attached the outputs of both versions in a separate file:

[alpaka_extensionIssue_output.md](https://github.com/user-attachments/files/26630225/alpaka_extensionIssue_output.md)

Besides, this behavior can be reproduced and verified easily.

Since I see no reason why `.cpp` and `.cc` should not be treated equivalently, I do not think this should be considered expected behavior. Just because we use `.cpp` by convention does not mean other projects will not use `.cc`, nor should `.cpp` be effectively mandatory for such projects.

If this limitation is already known, or clearly documented somewhere for the user, then the issue can of course be closed. However, I am currently not aware of that.

TimHanel00:
Since PR #491 has been approved and merged this is no longer a pending issue.

PR #491, opened 2026-04-10
fix `.cc` file extension issue

TimHanel00:
This PR aims to resolve Issue #490, by adjusting branch conditions in `cmake/finalize.cmake` to ensure consistent handling of `.cc` and `.cpp` source files.


psychocoderHPC (approved):

psychocoderHPC:
Using `cc` equal to `cpp` is fine since I read that `cc` stands for `C with Classes` from early times and is also often by google.

PR #492, opened 2026-04-12
refactor/add tutorial to documentation

psychocoderHPC:
Complete the tutorial to cover most alpaka objects with real-world use case examples. A strong focus was set on the wording for the frame specification because it is easily misunderstood as a thread/blocks configuration; therefore, there are small adjustments in the doxygen strings of FrameSpec.
Documentation code examples have now been tested with all backend configurations to avoid showing code that runs with the serial executor on the host only. The exception is for the device section, where the user gets an introduction on how devices can be selected.

- This PR extends the rst documentation with a new include command `filteredliteralinclude` for full source code files while excluding labels used to mark sections to reference parts of the code in rst files. 

PR is created with the help of CodexAI.

# Review suggestion
- Pull the PR to apply suggestions directly in the code base and provide a PR against this PR
- Read first the rendered tutorial section from this PR (you can enable a rendered diff too, press `d` to enable the diff)
- Check the example source code

Note: I have already iterated many times over this tutorial and rewrote parts, but it is now time for someone else to review it.

CC-ing: @TimHanel00, @SimeonEhrig, @AntonReinhard, @ikbuibui

PR #494, opened 2026-04-13
update `FrameSpec` and `ThreadSpec` doxygen

psychocoderHPC:
- add docygen string for `ThreadSpec`
- extent doxygen for `FrameSpec`
- tiny adjustments in examples

Changes pulled from #492

### DIFF
--- a/example/nBody/src/nBody.cpp
+++ b/example/nBody/src/nBody.cpp
@@ -174,7 +174,7 @@ namespace alpaka::example::nBody
 
         auto const numChunks = divCeil(Vec{extents.x()}, chunkSize);
 
-        // The frame spec describes the size of blocks and the grid used on the accelerator.
+        // The frame spec describes the logical parallelism exposed to the kernel, not the exact backend block/grid.
         auto const frameSpec = FrameSpec{numChunks, chunkSize};
 
         // Make an instance of the kernel object to use later
--- a/example/nBody/src/nBody.cpp
+++ b/example/nBody/src/nBody.cpp
@@ -174,7 +174,8 @@ namespace alpaka::example::nBody
 
         auto const numChunks = divCeil(Vec{extents.x()}, chunkSize);
 
-        // The frame spec describes the logical parallelism exposed to the kernel, not the exact backend block/grid.
+        // The frame spec describes the logical parallelism exposed to the kernel, not the exact mapping on the threads
+        // and the blocks of the backend.
         auto const frameSpec = FrameSpec{numChunks, chunkSize};
 
         // Make an instance of the kernel object to use later
--- a/example/tutorial/src/06_cudaLikeKernel.cpp
+++ b/example/tutorial/src/06_cudaLikeKernel.cpp
@@ -110,12 +110,12 @@ void testVectorAddKernel(alpaka::onHost::concepts::Device auto device, auto comp
     // fill the output buffer with zeros; the size is known from the buffer objects
     alpaka::onHost::memset(queue, out_d, 0x00);
 
-    /* If you use thread specifications alpak is not automatically adjusting the umber of threads per block.
-     * The kernel will be called with the specification you configured.
-     * In contrast to frame specification where the number of threads and blocks will be adjusted by alpaka to be
-     * optimal.
+    /* If you use a thread specification alpaka does not automatically adjust the number of blocks or threads per
+     * block. The kernel is called with exactly the block/thread configuration you configure. In contrast, a frame
+     * specification only describes the logical parallelism exposed to the kernel; alpaka may map that to a different
+     * physical block/thread configuration.
      */
-    auto threadSpec = alpaka::onHost::ThreadSpec{32u, 32u};
+    alpaka::onHost::concepts::ThreadSpec auto threadSpec = alpaka::onHost::ThreadSpec{32u, 32u};
 
     // launch the 1-dimensional kernel with scalar size
     if constexpr(alpaka::isSeqExecutor(computeExec))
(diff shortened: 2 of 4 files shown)


example/nBody/src/nBody.cpp
```diff
@@ -174,7 +174,7 @@ namespace alpaka::example::nBody
 
         auto const numChunks = divCeil(Vec{extents.x()}, chunkSize);
 
-        // The frame spec describes the size of blocks and the grid used on the accelerator.
+        // The frame spec describes the logical parallelism exposed to the kernel, not the exact backend block/grid.
```
SimeonEhrig:
```suggestion
        // The frame spec describes the logical parallelism exposed to the kernel, not the exact mapping on the threads and the blocks of the backend.
```

include/alpaka/onHost/FrameSpec.hpp
SimeonEhrig, line 23: `K <= P` Are you sure, that's correct? Maybe I misunderstand something. Therefore how I understand it:

Lets assume we have 100 elements (P) and and GPU with 32 Threads per core. If K should not bigger than P, I have the following cases:

- `K < P`, e.g. use 3 Frames with Frame Extent 32 -> K=96, I would assume that 3 cores handles the work and one core has to process as second chunk with 4 elements.
- `K == P`, e.g. use 4 Frames with Frame Extent 25 -> K=100, 4 Cores run in parallel and each core uses only 25 of 32 threads
- `K > P`, e.g. use 4 Frames with Frame Extent 32 -> K=128, I would assume that 4 cores handles the work and the last core processes only 4 elements.

SimeonEhrig:
That's a good improvement. FrameSpec something which is hard to understand but I required. @ikbuibui and I missed that during the documentation of PR #492.

include/alpaka/onHost/FrameSpec.hpp
psychocoderHPC: Yes the formulation is correct. For reasonable large P on a many core system, K schould be much smaller than P to have often the best performance. Much smaller means that you should at least keep the SIMD (S) width for your device into account.
You can also say P/S <= K. 
alpaka algorithm using more values and the upper threshold for K is smaller.
If you use hierarchical/tiled kernels you should also choose the frame extent wise but than shared memory is kicking in too.
There is no easy general rule but `K<=P` is often fine.
Rule of thump: less is more

Your cases:
case 1: Is for reasonable large P mostly the best. Best balanced work. Theoretically the best choice would be to have as many frames as you have multiprocessors. In practice oversubscribing multiprocessors with frames aby a factor 2 to 4 is running best. On Cpu you could get vectorized code from the auto vectorizer.
case 2: Is fine if we ignore gpus where 25 is not the best frame extent. Never the less you have no chance for using a grid strided loop. Vectorization is possible because on CPU we use 1 thread per frame but now it depends strongly on the frame extent.
case 3. Is valid and will work but same as case 2 most likely no good vectorizarion on CPU.

psychocoderHPC: Never the less the parameters depends on the kernel implementation and you should always start with something reasonable based e.g. on the rule in the doc string and as mentioned in the [tutorial PR you start to review](https://alpaka3--492.org.readthedocs.build/en/492/tutorial/kernel.html#choosing-the-correct-frame-specification), change it based on profiling later if it shows that the kernel became a bottleneck.

Or you use the alpaka auto tuner but also after using first something reasonable.

example/nBody/src/nBody.cpp
psychocoderHPC: fixed

example/tutorial/src/06_cudaLikeKernel.cpp
```diff
@@ -110,12 +110,12 @@ void testVectorAddKernel(alpaka::onHost::concepts::Device auto device, auto comp
...
     alpaka::onHost::memset(queue, out_d, 0x00);
 
-    /* If you use thread specifications alpak is not automatically adjusting the umber of threads per block.
-     * The kernel will be called with the specification you configured.
-     * In contrast to frame specification where the number of threads and blocks will be adjusted by alpaka to be
-     * optimal.
+    /* If you use a thread specification alpaka does not automatically adjust the number of blocks or threads per
+     * block. The kernel is called with exactly the block/thread configuration you configure. In contrast, a frame
+     * specification only describes the logical parallelism exposed to the kernel; alpaka may map that to a different
+     * physical block/thread configuration.
```
ikbuibui:
```suggestion
    /* Use of a thread specification enforces that the kernel will be called with exactly the block/thread configuration you configure. 
     * In contrast, a frame specification only describes the logical parallelism exposed to the kernel; alpaka may automatically 
     * adjust the number of blocks or threads per block and map that to a different physical block/thread configuration.
```

example/tutorial/src/06_cudaLikeKernel.cpp
```diff
@@ -192,15 +192,15 @@ void testVectorAddKernel3D(alpaka::onHost::concepts::Device auto device, auto co
...
     alpaka::onHost::memset(queue, out_d, 0x00);
 
-    /* If you use thread specifications alpak is not automatically adjusting the umber of threads per block.
-     * The kernel will be called with the specification you configured.
-     * In contrast to frame specification where the number of threads and blocks will be adjusted by alpaka to be
-     * optimal.
+    /* If you use a thread specification alpaka does not automatically adjust the number of blocks or threads per
+     * block. The kernel is called with exactly the block/thread configuration you configure.
+     * In contrast, a frame specification only describes the logical parallelism exposed to the kernel; alpaka may map
+     * that to a different physical block/thread configuration.
```
ikbuibui:
```suggestion
    /* Use of a thread specification enforces that the kernel will be called with exactly the block/thread configuration you configure. 
     * In contrast, a frame specification only describes the logical parallelism exposed to the kernel; alpaka may automatically 
     * adjust the number of blocks or threads per block and map that to a different physical block/thread configuration.
```

ikbuibui (commented):
I think this wording is clearer

include/alpaka/onHost/FrameSpec.hpp
SimeonEhrig: Okay, I was not sure if `FE` instead `K` is correct. Thanks for clarification.

example/tutorial/src/06_cudaLikeKernel.cpp
psychocoderHPC: thx done

example/tutorial/src/06_cudaLikeKernel.cpp
psychocoderHPC: done

ikbuibui (approved):

PR #495, opened 2026-04-13
refactor tutorial documentation

psychocoderHPC:
split PR from #492

- Refactor tutorial structure to an exploding side bar.
- Rename alpaka to always lower case
- add helper include command `filteredliteralinclude` for full source code files while excluding labels used to mark sections to reference parts of the code in rst files.
- add section for views, execution, events, and mental models

Note: There are 4 dead links in the doc, but I do not want to remove them, else it will be hard to partially fix it with the follow-up. They will become valid with the next PRs.

rendered view of the PR: https://alpaka3--495.org.readthedocs.build/en/495/

### DIFF
--- a/docs/snippets/example/02_execution.cpp
+++ b/docs/snippets/example/02_execution.cpp
@@ -1,0 +1,62 @@
+/* Copyright 2026 René Widera
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#include <alpaka/alpaka.hpp>
+
+#include <catch2/catch_test_macros.hpp>
+
+#include <iostream>
+
+using namespace alpaka;
+
+TEST_CASE("tutorial enumerate backends and executors", "[docs]")
+{
+    // BEGIN-TUTORIAL-enumerateDeviceSpec
+    // Numa aware CPU: onHost::DeviceSpec{api::host, deviceKind::numaCpu};
+    // Nvidia GPU: onHost::DeviceSpec{api::cuda, deviceKind::nvidiaGpu};
+    // Amd GPU: onHost::DeviceSpec{api::hip, deviceKind::amdGpu};
+    // Intel GPU: onHost::DeviceSpec{api::oneApi, deviceKind::intelGpu};
+    // this call selects the host Cpu
+    auto deviceSpec = onHost::DeviceSpec{api::host, deviceKind::cpu};
+    auto selector = onHost::makeDeviceSelector(deviceSpec);
+
+    auto numDevices = selector.getDeviceCount();
+    REQUIRE(numDevices >= 1u);
+
+    auto properties = selector.getDeviceProperties(0u);
+    onHost::concepts::Device auto device = selector.makeDevice(0u);
+    // END-TUTORIAL-enumerateDeviceSpec
+
+    CHECK(properties.warpSize >= 1u);
+    CHECK(!device.getName().empty());
+
+    size_t numVisitedBackends = 0u;
+    // BEGIN-TUTORIAL-enumerateBackends
+    onHost::executeForEachIfHasDevice(
+        [&](auto const& backend)
+        {
+            ++numVisitedBackends;
+
+            auto backendDeviceSpec = backend[object::deviceSpec];
+            auto backendExec = backend[object::exec];
+            auto backendSelector = onHost::makeDeviceSelector(backendDeviceSpec);
+            onHost::concepts::Device auto backendDevice = backendSelector.makeDevice(0u);
+            onHost::Queue backendQueue = backendDevice.makeQueue();
+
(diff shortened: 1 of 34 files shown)


AntonReinhard (commented):
Sorry, but I can't review this much code right now with the scrutiny it would require, given that is was not written by a human.

I have some general comments though:
1) I would suggest not putting copyright on the files that were generated entirely. According to several precedence cases in the US at least, generated code is automatically public domain. It is also misleading to anyone reading the code.
2) In that same vein, the tutorial should mention explicitly and unmistakably that it is generated. This is imho just basic courtesy.
3) Is there a specific reason the code snippets seem to be arbitrarily numbered?
4) From what I looked at, the tutorial is very verbose in some places (many llm-typical three point bullet lists or lists of examples that are really not necessary), and very surface-level in other places (for example the views/subviews).

Generally, considering how long we sometimes discussed how to exactly document some of our interfaces/concepts in a way that is explicit, brief, and understandable, I don't know how much sense it makes to just slap an LLM on the problem and hope for the best... It will likely also be a nightmare to maintain all of this text, because nobody has a good overview in their mind of what's in it and when it needs changing. A review doesn't replace writing it.

I also think especially for something like a tutorial, someone needs to make sure that the order in which things are explained explicitly makes sense. This too, I think, only really will work by writing it yourself.

SimeonEhrig:
I'm on a big review. I read each site and try to understand it. I will continue tomorrow.
@psychocoderHPC We need to clarify what is the purpose of the tutorial section. Partial it simply repeats the Basic sections but with much less information and no reference.

Maybe we can remove a lot of bullet points with man page style, where is simply a `See also` section at the end of the page.

psychocoderHPC:
Thanks for you feedback.

> 1. I would suggest not putting copyright on the files that were generated entirely. According to several precedence cases in the US at least, generated code is automatically public domain. It is also misleading to anyone reading the code.

I am not a lawyer, and I do not know the right way since most snippets are based on other examples. Never the less I will change the license ti ISC because that the license we use for examples in alpaka. I copied the wrong header.

> 2. In that same vein, the tutorial should mention explicitly and unmistakably that it is generated. This is imho just basic courtesy.

It is not fully generated by ML, most of this PR was existing before and Codex drafted a full outline where a lot was written bei codex and later chaged by me. But yes, we could add some information at the beginning.

> 3. Is there a specific reason the code snippets seem to be arbitrarily numbered?

No, it was numbered arbitrarily before, so I started to give numbers the same as in old times labels during programming and kept holes to move in new files.

> Generally, considering how long we sometimes discussed how to exactly document some of our interfaces/concepts in a way that is explicit, brief, and understandable, I don't know how much sense it makes to just slap an LLM on the problem and hope for the best... It will likely also be a nightmare to maintain all of this text, because nobody has a good overview in their mind of what's in it and when it needs changing. A review doesn't replace writing it.

The problem is that if we do not use the help of LLM for this topic, we will never end up with at least basic documentation within the next 6 months. The problem with maintaining the documentation is that it is independent of LLM because this is a typical issue that documentation is always outdated because you fail to update it. I tagged this as good as possible by having all code snippets for the documentation within the CI, this is also a good starting point because whenever a doc code example is touched, the possibility that you look into the corresponding rst file is high.
 
> I also think especially for something like a tutorial, someone needs to make sure that the order in which things are explained explicitly makes sense. This too, I think, only really will work by writing it yourself.

IMO the order is fine and also a matter of taste. In general, the overall order is from very basic stuff, before a kernel can be launched, to kernel execution and small examples showing different features.

psychocoderHPC:
> We need to clarify what is the purpose of the tutorial section. Partial it simply repeats the Basic sections but with much less information and no reference.

IMO we talked already about it, for me the tutorial should be a path on how to learn about all features of alpaka by guided examples, inspired by the [RAJA tutorial](https://raja.readthedocs.io/en/develop/sphinx/user_guide/tutorial.html)
 
> Maybe we can remove a lot of bullet points with man page style, where is simply a `See also` section at the end of the page.

IMO the tutorial should not be a technical manual page, this is what Doxygen already provides.
As I know, beginner users would like to have a guided entry into the topic with many "practical" examples.

I know that all of the documentation is not perfect, but one reason is that since I wrote 90% of alpaka3 I have a complete baised view and know about all things it provides.
This is not nessesary helping when writing documentation. 
That's why I would like to have the tutorial online so that new users can start pointing out things that are not clear from their perspective, and we have a base where we can start to improve the documentation.

psychocoderHPC:
I set this Pr to draft. There is an overall structure discussion in https://github.com/alpaka-group/alpaka3/pull/492.
The full new structure will be shown first in #492 and later split into reviewable slices in separate PRs.

psychocoderHPC:
#500 is replacing this pr

PR #497, opened 2026-04-15
add `Queue::isEmpty()`

psychocoderHPC:
- introduce the queue method `isEmpty()` (was already present in alpaka mainline but missed in alpaka3)
- FIX: during the implementation I found out that our event implementation for `OneApi` was not thread safe. It was possible that two threads changed the internal state of an event without any mutex.
- add the queue empty test case from alpaka mainline

The Sycl queue is now holding an event to the last enqueued action/task, this is required because Sycl is not providing a way to querry if the queue is empty. This has the disadvantage that if alpaka and native code is mixed alpaka `isEmpty()` can return a false positive because we can not track the queue state outside of alpaka.

### DIFF
--- a/include/alpaka/api/host/Queue.hpp
+++ b/include/alpaka/api/host/Queue.hpp
@@ -79,6 +79,14 @@ namespace alpaka::onHost
             uint32_t m_numaIdx = 0u;
             core::CallbackThread m_workerThread;
             bool m_isBlocking{false};
+            /** Flag to show if a blocking tasks is executed
+             *
+             * This variable is only used if m_isBlocking == true.
+             *
+             * state: If true a thread is executing a blocking tasks, else false.
+             */
+            std::atomic<bool> m_isBlockingTaskExecuted{false};
+
             /** Mutex to ensure sequential execution of tasks and operation if the queue is blocking.
              *
              * For non-blocking queue @c m_workerThread is taking care of the execution order
@@ -99,6 +107,7 @@ namespace alpaka::onHost
                 if(m_isBlocking)
                 {
                     std::lock_guard<std::mutex> lk(m_mutex);
+                    m_isBlockingTaskExecuted = true;
                     fn();
                     // silent tsan warnings: The promise is fulfilled directly and only a future which is true is
                     // returned, there can not be a data race in between.
@@ -113,6 +122,7 @@ namespace alpaka::onHost
 #if defined(__GNUC__) && !defined(__clang__)
 #    pragma GCC diagnostic pop
 #endif
+                    m_isBlockingTaskExecuted = false;
                     // to keep the uniform interface with the non-blocking case,
                     // return by moving the f since it is move-only
                     return f;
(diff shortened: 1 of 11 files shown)


include/alpaka/api/syclGeneric/Queue.hpp
```diff
@@ -262,15 +263,64 @@ namespace alpaka::onHost
...
             sycl::queue m_queue;
+            // secure that two threads can change the event at the same time
+            mutable std::shared_mutex m_eventGuard;
+            /** Event which is representing the last enqueued task/action by alpaka
+             *
+             * @attention You should not use the event directly, use always getLastEvent() or setLastEvent().
+             * This event is not tracking tasks enqueued aoutside of alpaka by using the native queue handle.
+             *
+             */
+            mutable sycl::event m_lastEvent;
```
psychocoderHPC, line 324: note: The usage of `mutable` is on purpose; this is how we did it already in alpaka mainline

include/alpaka/api/host/Queue.hpp
```diff
@@ -234,6 +244,30 @@ namespace alpaka::onHost
                 return this->shared_from_this();
             }
 
+            friend struct internal::IsQueueEmpty;
+
+            /** Checks if the queue is empty
+             *
+             * If m_isBlocking is true, only tasks will be taken into account, events will be ignored they could not
+             * influence the usage of isQueueEmpty. if m_isBlocking is false, events will be taken into account because
+             * they are handled as normal tasks.
```
SimeonEhrig, line 253: What happens, if you enqueue a event in a blocking queue? Does queue wait until the event is reached?

include/alpaka/api/syclGeneric/Queue.hpp
```diff
@@ -262,15 +263,65 @@ namespace alpaka::onHost
...
                 if(isBlocking())
                     ev.wait_and_throw();
             }
 
+            friend struct internal::IsQueueEmpty;
+
+            /** Test of all tasks in the queue are finished
+             *
+             * @attention We sue for the test the event of the last enqueued tasks/action by alpaka. This method is not
+             * able to detect enqueued tasks/actions outside of alpaka.
```
SimeonEhrig:
```suggestion
             * @attention We are testing for the last event of last enqueued alpaka event or action. The function cannot check events that were queued directly into the native queue, bypassing alpaka. 
```

include/alpaka/api/host/Queue.hpp
psychocoderHPC: yes the thread which is enqueing the event is waiting because, blocking means that the caller is blocked until the action/task ist finsihed.

include/alpaka/api/syclGeneric/Queue.hpp
psychocoderHPC: done

include/alpaka/api/host/Queue.hpp
SimeonEhrig: Does it mean, if an event was enqueue, the caller can continue and enqueue a task. Then maybe the event still waits, the task is not started and the queue will return `true` for `isEmpty`?

psychocoderHPC: No, your scenario is not possible for a **blocking** queue because the caller is not continued until the event is finished.

On a **blocking** queue, the caller will be blocked when enqueuing an event `isEmpty()` can only be called after the event finishes because the caller will not be continued before. The result of `isEmpty()` , assuming no other threads enqueue anything, will be `true`.

On a **non-blocking** queue, the caller will not be blocked. Enqueue an event after a long-running kernel, the caller is instantly continuing and can query `isEmpty()`. In case the kernel is still running and the event is therefore not already finished the query will return `false`.

PR #499, opened 2026-04-21
sphinx doc: on local system, build single header only of a source file is changed

SimeonEhrig:
Make much easier to use `make html` to build the documentation. The single header is automatically rebuild if required.

The rules for a build are:
- On readthedocs.io -> build all the time
- If single header is missing -> build all the time
- If environment variable `ALPAKA_SINGLE_HEADER` -> build depending on the value (`ON` -> force build, `OFF` -> force skip)
- Auto detection if a file is changed*

* The mechanism is the following. It takes `git status` to check which files was modified. The set can be filtered by a filter, e.g. only files from the `include` folder. Hash each file and store it in cache file. If a hash changed, a file was added or removed, trigger rebuild of the single header.

# TODO:

- [x] stop building documentation if single header build failes
- [x] Show only single header output in a case of an error

# Following up tasks

I will also use the same mechanism for the doxygen build. Therefore I open an extra PR.

### DIFF
--- a/docs/source/sphinx_helper/file_cache.py
+++ b/docs/source/sphinx_helper/file_cache.py
@@ -24,12 +24,18 @@ def get_modified_files(
     """
     cache_file_path_obj = pathlib.Path(cache_file_path)
 
+    git_get_commit_hash_process = subprocess.run(
+        ["git", "rev-parse", "HEAD"], stdout=subprocess.PIPE, text=True, check=True
+    )
+    current_git_commit_hash = git_get_commit_hash_process.stdout.strip()
+
     old_changes: dict[str, str] = {}
     if cache_file_path_obj.exists():
         with open(cache_file_path_obj, "r", encoding="UTF-8") as f:
             old_changes: dict[str, str] = json.load(f)
 
     current_changes = get_hashed_files(path_filter_regex)
+    current_changes["commit"] = current_git_commit_hash
 
     def dict_diffs(A: dict, B: dict) -> dict:
         _sentinel = object()
--- a/docs/source/sphinx_helper/file_cache.py
+++ b/docs/source/sphinx_helper/file_cache.py
@@ -1,0 +1,97 @@
+"""Provied functionally to check if a file was changed since last doc build process."""
+
+import re
+import hashlib
+import json
+import pathlib
+import subprocess
+
+
+def get_modified_files(
+    cache_file_path: str, path_filter_regex: str = ""
+) -> dict[str, str]:
+    """Return all files, which changed since last documentation build. If empty, nothing changed.
+
+    Args:
+        cache_file_path (str): Name of the cache file.
+        path_filter_regex (str, optional): Filter which file paths should be included.
+            The paths are relative to the Git project root.
+            Defaults to "", which means all files in the repostiory.
+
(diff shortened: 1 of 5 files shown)


SimeonEhrig:
Fond a bug. A checkout of another commit will not trigger a rebuild.

docs/source/sphinx_helper/file_cache.py
```diff
@@ -0,0 +1,103 @@
+"""Provied functionally to check if a file was changed since last doc build process."""
```
psychocoderHPC: Provied -> Provide

psychocoderHPC: Maybe: `Provide functionality ...`

docs/source/sphinx_helper/file_cache.py
```diff
@@ -0,0 +1,103 @@
...
+def get_modified_files(
+    cache_file_path: str, path_filter_regex: str = ""
+) -> dict[str, str]:
+    """Return all files, which changed since last documentation build. If empty, nothing changed.
+
+    Args:
+        cache_file_path (str): Name of the cache file.
+        path_filter_regex (str, optional): Filter which file paths should be included.
+            The paths are relative to the Git project root.
+            Defaults to "", which means all files in the repostiory.
```
psychocoderHPC: repostiory -> repository

docs/source/sphinx_helper/file_cache.py
```diff
@@ -0,0 +1,103 @@
...
+    diff = dict_diffs(old_changes, current_changes)
+
+    with open(cache_file_path_obj, "w", encoding="UTF-8") as f:
+        json.dump(current_changes, f, indent=4)
+
+    return diff
+
+
+def get_hashed_files(path_filter_regex: str = "") -> dict[str, str]:
+    """Search for all files which are new or modified basing on git status in the repository.
```
psychocoderHPC: basing -> based

docs/source/sphinx_helper/single_header.py
```diff
@@ -0,0 +1,87 @@
...
+        return True
+
+    single_header_path = os.path.abspath(os.path.join(app.builder.outdir))
+    if not pathlib.Path(single_header_path).exists():
+        logger.info("Single Header: create because single header does not exist.")
+        return True
+
+    if "ALPAKA_SINGLE_HEADER" in os.environ:
+        env_value = os.environ["ALPAKA_SINGLE_HEADER"]
+        if env_value == "ON":
```
psychocoderHPC: Only a note: `ON`, `OFF` is not typically used for environment variables. The old behaviour setting to `1` is more natural, I liked it more, but I am ok with the change.
It would be nice if we could support `ON`,`OFF`,`1` and `0`.

psychocoderHPC (changes requested):

ikbuibui (approved):

docs/source/sphinx_helper/single_header.py
SimeonEhrig: You are right. I support now `0` or `OFF` and `1` or `ON`.

psychocoderHPC (approved):

PR #501, opened 2026-04-22
document terms: API, device kind and device

SimeonEhrig:
Actual I had a good idea how to define the term `frame` yesterday. But therefore I needs some other terms :sweat_smile:  So I started with this terms :-)

### DIFF
--- a/docs/source/basic/terms.rst
+++ b/docs/source/basic/terms.rst
@@ -56,9 +56,9 @@ Properties:
 API
 ---
 
-The ``API`` represents the runtime environment used to execute a kernel on a specific processor.
-Depending on the runtime environment, different processor types are available.
-The processor type is selected via :ref:`device_kind`, see the next section.
+The ``API`` represents the underlying runtime environment that alpaka code is mapped to work on. Specifically, this ``API`` is the environment used to execute a kernel on a particular processor.
+Depending on the ``API``, one or more different processor types may be available.
+The processor type is selected via :ref:`device_kind`.
 
 .. figure:: ../tutorial/images/api_deviceKind.svg
 
@@ -75,15 +75,15 @@ Device Kind
 ```````````
 
 The ``Device Kind`` determines which type of processor we want to use.
-The combination of :ref:`api` and ``Device Kind`` results in a specific processor type.
+The combination of :ref:`api` and ``Device Kind`` defines a specific processor type, and is called a ``DeviceSpec`` in alpaka.
 Depending on the system, zero, one, or many processors may be available (e.g., in multi-GPU systems).
-Each of these processors is a own :ref:`Device`.
+Each of these processors is an own :ref:`Device`.
 
 The following device types are available:
 
 - ``cpu``: The host system's CPU. It uses the same processor as a standard C++ application. On a single-socket system, it uses the entire CPU. On a multi-socket system, the system can be configured as a UMA system [#f3]_, which means that two or more physical CPUs appear as a single large CPU. In this case, alpaka displays a single device.
 - ``numaCpu``: The NUMA configuration [#f4]_ of the host CPU is taken into account. A single CPU or multiple CPUs can be divided into smaller logical CPUs, which respects inhomogeneous memory topology. For example, in a dual-socket system, the operating system uses both CPUs as a single large virtual CPU. The NUMA configuration can then subdivide the two CPUs to account for the different latencies to the respective memory modules.
-- ``xGpu``: A GPU from vendor ``x``.
+- ``intelGpu``, ``nvidiaGpu`` and ``amdGpu``: A GPU from a specific vendor.
 
 
 .. [#f3] https://en.wikipedia.org/wiki/Uniform_memory_access
@@ -102,7 +102,7 @@ In alpaka, we use a combination of :ref:`api` and :ref:`device_kind` to select d
 The programmatic approach is described in the :ref:`Device Selection <device-selection>` section of the ``Getting Started`` tutorial.
 
 Each device is controlled separately by the :ref:`host`.
-This means that if a :ref:`kernel` is to be run on two GPUs in a system, GPU 0 (device 0) is selected first, the kernel is started there, then GPU 1 (device 1) is selected, and the same kernel is started there again.
+This means that if a :ref:`kernel` is to be run on two GPUs in a system, then one possible way to do it would be to select GPU 0 (device 0) first and start the kernel there, and then select GPU 1 (device 1)  and then start the same kernel there again.
 
 .. [#f5] Depending on the :ref:`device_kind`, a system can provide a different number of CPUs. On a system with two sockets, there may be one CPU device if the :ref:`device_kind` is ``CPU``, or two CPU devices if the :ref:`device_kind` is ``numaCPU``.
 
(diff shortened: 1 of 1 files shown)


docs/source/basic/terms.rst
```diff
@@ -54,14 +56,69 @@ Properties:
...
+The ``Device Kind`` determines which type of processor we want to use.
+The combination of :ref:`api` and ``Device Kind`` results in a specific processor type.
+Depending on the system, zero, one, or many processors may be available (e.g., in multi-GPU systems).
+Each of these processors is a own :ref:`Device`.
+
+The following device types are available:
+
+- ``cpu``: The host system's CPU. It uses the same processor as a standard C++ application. On a single-socket system, it uses the entire CPU. On a multi-socket system, the system can be configured as a UMA system [#f3]_, which means that two or more physical CPUs appear as a single large CPU. In this case, alpaka displays a single device.
+- ``numaCpu``: The NUMA configuration [#f4]_ of the host CPU is taken into account. A single CPU or multiple CPUs can be divided into smaller logical CPUs, which respects inhomogeneous memory topology. For example, in a dual-socket system, the operating system uses both CPUs as a single large virtual CPU. The NUMA configuration can then subdivide the two CPUs to account for the different latencies to the respective memory modules.
+- ``xGpu``: A GPU from vendor ``x``.
```
psychocoderHPC: I suggest to be explicit here, we have only 3 vendors, `intelGpu`, `nvidiaGpu` and `amdGpu`.

docs/source/basic/terms.rst
```diff
@@ -54,14 +56,69 @@ Properties:
 API
 ---
 
+The ``API`` represents the runtime environment used to execute a kernel on a specific processor.
+Depending on the runtime environment, different processor types are available.
```
ikbuibui:
```suggestion
The ``API`` represents the underlying runtime environment that alpaka code is mapped to work on. Specifically, this ``API`` is the environment used to execute a kernel on a particular processor.
Depending on the ``API``, one or more different processor types may be available.
```
I think this is slightly more accurate

docs/source/basic/terms.rst
```diff
@@ -54,14 +56,69 @@ Properties:
...
+The processor type is selected via :ref:`device_kind`, see the next section.
```
ikbuibui:
```suggestion
The processor type is selected via :ref:`device_kind`.
```
You already added a reference, so the text next section isn't needed, as it might be moved around later.

docs/source/basic/terms.rst
```diff
@@ -54,14 +56,69 @@ Properties:
...
+
+.. _device_kind:
+
+Device Kind
+```````````
+
+The ``Device Kind`` determines which type of processor we want to use.
+The combination of :ref:`api` and ``Device Kind`` results in a specific processor type.
+Depending on the system, zero, one, or many processors may be available (e.g., in multi-GPU systems).
+Each of these processors is a own :ref:`Device`.
```
ikbuibui:
```suggestion
Each of these processors is an own :ref:`Device`.
```

docs/source/basic/terms.rst
ikbuibui:
```suggestion
- ``xGpu``: A GPU from vendor ``x``, for example ``amdGpu`` for ``AMD``.
```

docs/source/basic/terms.rst
```diff
@@ -54,14 +56,69 @@ Properties:
...
+A device is a specific processor in the system, such as a CPU like an ``Intel Xeon`` or ``AMD EPYC``, or a GPU like an ``Nvidia H100`` or ``AMD Instinct``.
+Depending on the system, there may be more than one processor of the same type.
+For example, a single HPC GPU node may contain eight GPUs of the same type and two NUMA CPUs [#f5]_ of the same type.
 
-Kernel
+In alpaka, we use a combination of :ref:`api` and :ref:`device_kind` to select devices of the same type.
+The programmatic approach is described in the :ref:`Device Selection <device-selection>` section of the ``Getting Started`` tutorial.
+
+Each device is controlled separately by the :ref:`host`.
+This means that if a :ref:`kernel` is to be run on two GPUs in a system, GPU 0 (device 0) is selected first, the kernel is started there, then GPU 1 (device 1) is selected, and the same kernel is started there again.
```
psychocoderHPC: We do not enqueue the kernel to the device, I suggest refracing this.

```
This means that if a :ref:kernel is to run on two GPUs in a system, the host first selects GPU 0 (device 0) and enqueues the kernel to one of that device’s queues. It then selects GPU 1 (device 1) and enqueues the same kernel to one of that device’s queues as well.
```

psychocoderHPC (approved):
Non-mandatory suggestions

docs/source/basic/terms.rst
ikbuibui:
```suggestion
This means that if a :ref:`kernel` is to be run on two GPUs in a system, then one possible way to do it would be to select GPU 0 (device 0) first and start the kernel there, and then select GPU 1 (device 1)  and then start the same kernel there again.
```

ikbuibui (changes requested):
Some small changes which I think improve the clarity.

docs/source/basic/terms.rst
```diff
@@ -54,14 +56,69 @@ Properties:
...
+- ``hip``: The `AMD ROCm/HIP SDK <https://rocmdocs.amd.com/en/latest/index.html>`__ is used to run kernels on AMD GPUs.
+- ``oneAPI``: The `Intel OneAPI Toolkit <https://rocmdocs.amd.com/en/latest/index.html>`__ is used to run kernels on CPUs, Intel GPUs and Nvidia and AMD GPUs (plugins required)
+
+.. _device_kind:
+
+Device Kind
+```````````
+
+The ``Device Kind`` determines which type of processor we want to use.
+The combination of :ref:`api` and ``Device Kind`` results in a specific processor type.
```
ikbuibui:
```suggestion
The combination of :ref:`api` and ``Device Kind`` defines a specific processor type, and is called a ``DeviceSpec`` in alpaka.
```

docs/source/basic/terms.rst
SimeonEhrig: Actual I would avoid to use queues already here because it was not introduce yet and queues are not trivial. @ikbuibui provided an better version of the sentence. Please check this.

ikbuibui (approved):

docs/source/basic/terms.rst
SimeonEhrig: I toke @psychocoderHPC suggestion.

docs/source/basic/terms.rst
```diff
@@ -54,14 +56,69 @@ Properties:
...
+
+.. _device_kind:
+
+Device Kind
+```````````
+
+The ``Device Kind`` determines which type of processor we want to use.
+The combination of :ref:`api` and ``Device Kind`` defines a specific processor type, and is called a ``DeviceSpec`` in alpaka.
+Depending on the system, zero, one, or many processors may be available (e.g., in multi-GPU systems).
+Each of these processors is an own :ref:`Device`.
```
SimeonEhrig, line 80:
```suggestion
The ``Device Kind`` determines which type of processor we want to use. Here I want to have a change, which you can directly merge.
The combination of :ref:`api` and ``Device Kind`` defines a specific processor type, and is called a ``DeviceSpec`` in alpaka.
Depending on the system, zero, one, or many processors may be available (e.g., in multi-GPU systems).
Each of these processors is an own :ref:`Device`.
```

psychocoderHPC (approved):

PR #502, opened 2026-04-22
sphinx doc: on a local system, build doxygen only of a source file is changed

SimeonEhrig:
Same functionality like in PR #499 but for the Doxygen build instead the single header build

- behavior can be overwritten with the environment variable  ALPAKA_DOC_DOXYGEN
- changed environment variable `ALPAKA_SINGLE_HEADER` to `ALPAKA_DOC_SINGLE_HEADER`
- disable some Doxygen warnings because we agreed that not every trivial function needs to be documented and also some functions can be documented with a single line instead whole parameter documentation.

### DIFF
--- a/docs/source/sphinx_helper/doxygen.py
+++ b/docs/source/sphinx_helper/doxygen.py
@@ -128,7 +128,7 @@ def build_doxygen(app):
             logger.warning(doxygen_process.stderr.strip())
         if doxygen_process.returncode != 0:
             logger.error(f"{cmd} failed")
-            sys.exit(0)
+            sys.exit(doxygen_process.returncode)
 
     # copy user and developer to sphinx doc html output
     for src, dest in get_src_dest_paths(app):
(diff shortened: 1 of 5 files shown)


docs/source/sphinx_helper/doxygen.py
```diff
@@ -0,0 +1,142 @@
...
+    for cmd in (["doxygen"], ["doxygen", "Doxyfile_dev"]):
+        logger.info(f"Run {' '.join(cmd)}")
+        doxygen_process = subprocess.run(
+            cmd, cwd=docs_dir, stdout=subprocess.PIPE, text=True, check=True
+        )
+        if doxygen_process.stderr:
+            logger.warning(doxygen_process.stderr.strip())
+        if doxygen_process.returncode != 0:
+            logger.error(f"{cmd} failed")
+            sys.exit(0)
```
ikbuibui:
```suggestion
            sys.exit(doxygen_process.returncode)
```
or maybe simply return 1

ikbuibui (changes requested):
Looks okay, small change required

docs/source/sphinx_helper/doxygen.py
SimeonEhrig: Your are right. This is a bug and a code suggestion.

ikbuibui (approved):

PR #505, opened 2026-04-23
fixed doc string for alpaka::Vec named components

SimeonEhrig:
- refactor macro `ALPAKA_NAMED_ARRAY_ACCESS` to avoid misunderstandings

Before the components was created with:

```c++
        ALPAKA_NAMED_ARRAY_ACCESS(x, 0u)
        ALPAKA_NAMED_ARRAY_ACCESS(y, 1u)
        ALPAKA_NAMED_ARRAY_ACCESS(z, 2u)
        ALPAKA_NAMED_ARRAY_ACCESS(w, 3u)
```

This was misleading because it look likes x would point to index position 0. 

- fixed doc string also for assignment and binary operators

### DIFF
--- a/include/alpaka/Vec.hpp
+++ b/include/alpaka/Vec.hpp
@@ -308,9 +308,6 @@ namespace alpaka
             return Vec([this](uint32_t const i) constexpr { return -(*this)[i]; });
         }
 
-/** assign operator
- * @{
- */
 #define ALPAKA_VECTOR_ASSIGN_OP(op)                                                                                   \
     template<typename T_OtherStorage>                                                                                 \
     constexpr Vec& operator op(Vec<T_Type, T_dim, T_OtherStorage> const& rhs)                                         \
@@ -344,16 +341,17 @@ namespace alpaka
         return *this;                                                                                                 \
     }
 
-
+        /** assign operator
+         * @{
+         */
         ALPAKA_VECTOR_ASSIGN_OP(+=)
         ALPAKA_VECTOR_ASSIGN_OP(-=)
         ALPAKA_VECTOR_ASSIGN_OP(/=)
         ALPAKA_VECTOR_ASSIGN_OP(*=)
         ALPAKA_VECTOR_ASSIGN_OP(=)
+        /** @} */
 
 #undef ALPAKA_VECTOR_ASSIGN_OP
-
-        /** @} */
 
         constexpr decltype(auto) operator[](std::integral auto const idx)
         {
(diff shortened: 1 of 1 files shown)


include/alpaka/Vec.hpp
```diff
@@ -365,30 +363,29 @@ namespace alpaka
...
 
-        /** named member access
-         *
-         * index -> name [0->x,1->y,2->z,3->w]
-         * @{
-         */
-#define ALPAKA_NAMED_ARRAY_ACCESS(functionName, dimValue)                                                             \
-    constexpr decltype(auto) functionName() requires(T_dim >= dimValue + 1)                                           \
+#define ALPAKA_NAMED_ARRAY_ACCESS(functionName, index_pos)                                                            \
+    constexpr decltype(auto) functionName() requires(index_pos >= 0)                                                  \
```
psychocoderHPC: This introduces a bug. The index position is unsigned and using for the call `DIM - 3u` for `w` will be a large number if the vector is one dimensional and it will be allowed to call `w()` because the condition is true.
The requires must be `indexPos < T_dim`.

include/alpaka/Vec.hpp
```diff
@@ -365,30 +363,29 @@ namespace alpaka
...
         }
 
-        /** named member access
-         *
-         * index -> name [0->x,1->y,2->z,3->w]
-         * @{
-         */
-#define ALPAKA_NAMED_ARRAY_ACCESS(functionName, dimValue)                                                             \
-    constexpr decltype(auto) functionName() requires(T_dim >= dimValue + 1)                                           \
+#define ALPAKA_NAMED_ARRAY_ACCESS(functionName, index_pos)                                                            \
```
psychocoderHPC: `index_pos` should be `indexPos` because we use camelaCase not snake_case namings

include/alpaka/Vec.hpp
psychocoderHPC: or `IndexPos >=0 && indexPos < T_dim` because the other could throw a compiler error if dim is one.

psychocoderHPC (changes requested):
a bug is introduced

include/alpaka/Vec.hpp
SimeonEhrig: I use only `indexPos < T_dim` and not also `IndexPos >=0`. Technically `IndexPos >=0` is true all the time, because we unsigned numbers. But the check give the feeling, that we work with signed numbers which is miss leading. Instead I wrote a comment, that a overflow happen which is fine and we care about.

include/alpaka/Vec.hpp
```diff
@@ -365,30 +363,32 @@ namespace alpaka
...
 
-        /** named member access
-         *
-         * index -> name [0->x,1->y,2->z,3->w]
-         * @{
-         */
-#define ALPAKA_NAMED_ARRAY_ACCESS(functionName, dimValue)                                                             \
-    constexpr decltype(auto) functionName() requires(T_dim >= dimValue + 1)                                           \
+#define ALPAKA_NAMED_ARRAY_ACCESS(functionName, indexPos)                                                             \
+    /* indexPos can overflow for example if T_dim is 1 and y() (T_dim - 2u) should be declared */                     \
```
psychocoderHPC: underflow

SimeonEhrig: In a discussion we clarified, that `underflow` is not correct. A underflow is, when a floating point number near 0 become 0 because of the floating number precision.

https://en.wikipedia.org/wiki/Integer_overflow

include/alpaka/Vec.hpp
```diff
@@ -365,30 +363,32 @@ namespace alpaka
...
+    /* Therefore the `requires` is fine, because everything which should be theoretical negative become much bigger   \
```
psychocoderHPC: why do you have two comments instead of having all information within a single comment?

SimeonEhrig: Because it is a macro. I already had problems to write comment. Therefore I didn't thought about merging the comments. Comments are merged now.

psychocoderHPC (approved):

PR #506, opened 2026-04-23
document terms frame, frame extents and frame spec

SimeonEhrig:
It is a first draft and I'm pretty sure I did a bigger mistake. But I think it is a good starting to discuss the definition of a frame.


docs/source/basic/terms.rst
```diff
@@ -215,3 +215,54 @@ Kernel
 
 Frame, Frame Extents and Frame Spec
 -----------------------------------
+
```
psychocoderHPC, line 266: I would describe it top down.


draft:
```
A frame specification ``FrameSpec`` is the description of the logical parallism exposed to a kernel.
It describes how a multi dimensional index range is decomposed into fixed-size chunks called ``frames``.
The number of indices represented by a ``frame`` is described by the ``frame extent``.

The finite index domain created by a frame specification is not required to match the problem domain size solved within a kernel.
Based on the ``api`` and ``deviceKind`` *alpaka* is deriving a thread block hierarchy specification ``ThreadSpec`` from ``FrameSpec``.
The ``FrameSpec`` is the upper bound of paralism usable in the kernel.

The real mapping of indices of the full problem to worker threads is performed within the kernel with ``onAcc::makeIdxMap``.
```

docs/snippets/terms/thread_spec_kernel.cpp
```diff
@@ -0,0 +1,76 @@
...
+    onHost::Queue queue = device.makeQueue(queueKind::nonBlocking);
+
+    auto hostMemory = onHost::allocHost<int>(alpaka::Vec{size});
+    auto deviceMemory = onHost::allocLike(device, hostMemory);
+    onHost::fill(queue, deviceMemory, 42);
+
+    // BEGIN-TERMS-threadspec
+    // thread size 1 is valid on all accelerators
+    size_t numberOfThreads = 1;
+    // for the kernel, it is required that the of product of the number of threads and blocks is bigger than the size
```
psychocoderHPC:
```suggestion
    // for the kernel, it is required that the product of the number of threads and blocks is greater than the buffer size
```

docs/snippets/terms/thread_spec_kernel.cpp
```diff
@@ -0,0 +1,76 @@
...
+    size_t numberOfBlocks = std::ceil(static_cast<double>(size) / static_cast<double>(numberOfThreads));
```
psychocoderHPC:
```suggestion
    size_t numberOfBlocks = std::max(size_t{1u}, alpaka::divCeil(size, numberOfThreads));
```

docs/snippets/terms/thread_spec_kernel.cpp
```diff
@@ -0,0 +1,76 @@
...
+    onHost::concepts::Device auto device = selector.makeDevice(0);
+    onHost::Queue queue = device.makeQueue(queueKind::nonBlocking);
+
+    auto hostMemory = onHost::allocHost<int>(alpaka::Vec{size});
+    auto deviceMemory = onHost::allocLike(device, hostMemory);
+    onHost::fill(queue, deviceMemory, 42);
+
+    // BEGIN-TERMS-threadspec
+    // thread size 1 is valid on all accelerators
+    size_t numberOfThreads = 1;
```
psychocoderHPC: maybe `numThreadsPerBlock` to avoid thinking it is the total number of threads used to solve the problem

docs/snippets/terms/thread_spec_kernel.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+    if(selector.getDeviceCount() == 0)
+    {
+        return;
+    }
+
+    onHost::concepts::Device auto device = selector.makeDevice(0);
+    onHost::Queue queue = device.makeQueue(queueKind::nonBlocking);
+
+    auto hostMemory = onHost::allocHost<int>(alpaka::Vec{size});
```
psychocoderHPC:
```suggestion
    auto hostMemory = onHost::allocHost<int>(size);
```

docs/snippets/terms/thread_spec_kernel.cpp
```diff
@@ -0,0 +1,76 @@
...
+struct VectorAddKernel1D
+{
+    template<typename TAcc>
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, alpaka::concepts::IMdSpan auto inout) const
+    {
+        // get the global thread ID depending on the Thread Spec
+        std::integral auto globalThreadId
+            = acc.getIdxWithin(alpaka::onAcc::origin::grid, alpaka::onAcc::unit::threads).product();
+
+        // if the global thread ID is bigger than the data size, do nothing
```
psychocoderHPC:
```suggestion
        // if the global thread ID is larger than the data size, do nothing
```

docs/snippets/terms/thread_spec_kernel.cpp
```diff
@@ -0,0 +1,76 @@
...
+        // if the global thread ID is smaller, not all data is processed
```
psychocoderHPC: What do you like to say here. Why is not all data processed if the global thread id is smaller?
Do you like to say: `If the global number of threads is smaller, not all data will be processed `?

psychocoderHPC (changes requested):

docs/snippets/terms/thread_spec_kernel.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+// BEGIN-TERMS-kernel-threadspec
+struct VectorAddKernel1D
+{
+    template<typename TAcc>
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, alpaka::concepts::IMdSpan auto inout) const
+    {
+        // get the global thread ID depending on the Thread Spec
+        std::integral auto globalThreadId
+            = acc.getIdxWithin(alpaka::onAcc::origin::grid, alpaka::onAcc::unit::threads).product();
```
psychocoderHPC: product should not be used here.  Technical it is correct because the product of a one dimension verctor is the value itself but it would be better to use structured binding here or access .x().

docs/snippets/terms/thread_spec_kernel.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+        // if the global thread ID is bigger than the data size, do nothing
+        // if the global thread ID is smaller, not all data is processed
+        if(globalThreadId < inout.getExtents().product())
```
psychocoderHPC: For extents the product is fine even if the inout would be multi dimensional, never the less to ne constistent with the thread index querry I suggest using .x() or a temporary variable and structured bindings ro get the value.

docs/snippets/terms/frame_spec_kernel.cpp
```diff
@@ -0,0 +1,74 @@
...
+    onHost::fill(queue, deviceMemory, 42);
+
+    // BEGIN-TERMS-framespec
+    Vec frameExtents{8};
+    Vec numberOfFrames{6};
+    auto frameSpec = onHost::FrameSpec{numberOfFrames, frameExtents};
+
+    // The Thread Spec  is calculated automatically before the kernel starts.
+    // It uses the device to retrieve the accelerator from the queue
+    // in order to calculate the Thread Spec.
```
psychocoderHPC, line 63:
```suggestion
    /* The @see ThreadSpec is derived automatically before the kernel starts.
     * The properties required to derive the ThreadSpec are obtained from the device on which the queue is created.
     */
```
Please try to use multi-line comments; only old code taken over from alpaka mainline is using single line comments.
I removed the accelerator on purpose because the accelerator is based on the executor.
The accelerator is mapping hardware threads and functionality, e.g. shared memory, atomics, ... to the device.
In this example, the executor is not visible because the best-fitting executor for the device is used, this is described in the enqueue interface.

docs/snippets/terms/frame_spec_kernel.cpp
```diff
@@ -0,0 +1,74 @@
...
+// BEGIN-TERMS-kernel-framespec
+struct VectorAddKernel1D
+{
+    template<typename TAcc>
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, alpaka::concepts::IMdSpan auto inout) const
+    {
+        // Take the extent of problem/data domain `inout`.
+        // With `onAcc::worker::threadsInGrid` map all frames on the available execution units.
+        // `i` is the id of the hardware thread. `onAcc::makeIdxMap` returns an interator with
+        // the resolved mapping of data positions to hardware threads.
```
psychocoderHPC, line 25:
```suggestion
      /* Use the extent of `inout` as the problem/data index range.
       * `onAcc::makeIdxMap` creates an iterator that traverses all valid indices assigned to each hardware thread.
       * `onAcc::worker::threadsInGrid` specifies that all threads in the grid are used, ignoring thread block hierarchies.
       * `i` is the n-dimensional index each thread uses to access elements in `inout`, with no additional remapping or permutation of data access.
       */    
```

I corrected the comment. There are no frames involved in this loop. `i` is not the id of the hardware thread; it is a valid index within the index range a thread is operating on.

note: While writing this, I realized that I use the words `index domain` for the problem we would like to solve but *alpaka* is providing an object `IdxRange`; therefore, we should try to use the wording `index range` in the future instead of `index domain`.
`index range` is IMO also better since the range allows slicing and strides and for users `range` is more common for such operations.

psychocoderHPC (changes requested):
Please take the already existing comments into account too.

docs/snippets/terms/thread_spec_kernel.cpp
SimeonEhrig: Please check the new comments.

docs/snippets/terms/thread_spec_kernel.cpp
```diff
@@ -0,0 +1,83 @@
...
+            = acc.getIdxWithin(alpaka::onAcc::origin::grid, alpaka::onAcc::unit::threads);
+
+        static_assert(globalThreadId.dim() == inout.dim());
+
+        /* `globalThreadId < inout.getExtents()` prevents out of range access.
+         * All global threads, which are greater than `inout.getExtents()` do not process data and are idle.
+         * If the total number of threads is less than `inout.getExtents()`, not all data is processed, and the result
+         * is incorrect.
+         */
+        if(globalThreadId < inout.getExtents())
```
psychocoderHPC, line 33: This is not working because same as in alpaka mainline operations on a vector working element-wise except `==` and `!=`

https://github.com/alpaka-group/alpaka3/blob/864a0cd4c89c75ab59db2aea8b40fd3d376ae14b/docs/source/tutorial/vector.rst?plain=1#L79-L80

psychocoderHPC, line 33:
```suggestion
        if((globalThreadId < inout.getExtents()).reduce(std::locical_and())
```

psychocoderHPC (changes requested):
I am not done with the full review, I need to review the rst again

psychocoderHPC:
The image is cropped, parts of the image are not visisble.

SimeonEhrig:
Close because nobody can explain frame, frame extents and frame spec in a reasonable time.

PR #509, opened 2026-04-28
improved ``accelerator`` term

SimeonEhrig:


docs/source/basic/terms.rst
```diff
@@ -35,10 +35,19 @@ Properties:
...
 
+.. _accelerator:
+
 Accelerator
 ```````````
 
-The *onAcc* namespace contains the actual algorithms that are executed on the accelerator device, like a CPU or GPU.
+The ``accelerator`` is a virtual processor that is mapped to a hardware processor.
+Therefore, the term ``accelerator`` is used as an abstraction layer (virtual processor) and serves to define which code is offloaded to a physical processor.
+The *onAcc* namespace contains the actual algorithms that are executed on the ``accelerator``, like a CPU or GPU.
```
psychocoderHPC:
```suggestion
The **onAcc** namespace contains the actual algorithms that are executed within a kernel on the ``accelerator``, such as a CPU or GPU.
```

docs/source/basic/terms.rst
```diff
@@ -35,10 +35,19 @@ Properties:
...
+
+The ``accelerator`` does not represent a specific processor in the system. It represents a category of processors.
+For example, if we have a system with two Intel Xeon CPUs and four AMD Instinct GPUs, an ``accelerator`` can be an Intel Xeon CPU or an AMD Instinct GPU. A specific processor is a :ref:`Device`.
+The example system has 2 CPU :ref:`Devices <Device>` and 4 GPU :ref:`Devices <Device>`.
```
psychocoderHPC:
```suggestion

```

It is repeating the last sentence but without saying anything about the device names.

docs/source/basic/terms.rst
```diff
@@ -35,10 +35,19 @@ Properties:
...
+Therefore, the host selects a :ref:`Device`, but the actual algorithm is written for an ``accelerator``, since it can run on different :ref:`Devices <Device>`.
```
psychocoderHPC:
```suggestion
The host selects a :ref:`Device`, but the kernel is written for an abstract ``accelerator``, since it can be executed on different :ref:`Devices <Device>`.
```

psychocoderHPC (changes requested):
Some changes. I would not talk from pure algorithms and make clear that the kernel is describing the algorithm. In reality, an algorithm is not limited to a single kernel, a full workflow with many kernels is an algorithm too.

docs/source/basic/terms.rst
psychocoderHPC: Maybe if you want to repeat what alpaka is seeing:

```
Therefore *alpaka* has access to the ``deviceKind::cpu`` and four devices of ``deviceKind::amdGpu``.
```

I would not go into detail on why you do not see two CPUs, and that you could use `hwloc` to access both CPUs. This would make everything too complicated at this point because this is too detailed information.
The default case is that you would use the device kind `cpu` where you see the host CPUs as a single device.

SimeonEhrig: I would not talk about Numa here. I overworked the example. I understand why it looks redundant.

psychocoderHPC: Exactly, I would also avoid NUMA at this point.

SimeonEhrig: Is the example now useful? I would avoid `deviceKind` for the explanation here, because it is introduced later.

SimeonEhrig:
Close because the overwork takes to much time.

PR #511, opened 2026-04-28
fix SIMD math

psychocoderHPC:
Fix hard coded math call.


SimeonEhrig (approved):

SimeonEhrig:
Wait a moment, for each unary SIMD math function, where a native implement exist, we simply executed a square root and didn't recognized it?

PR #515, opened 2026-05-04
add test helper, which retruns a device if available and an optinal executor

SimeonEhrig:
Create and device and executor. Check if a device is available and skip test if not. Prints a lot of information about the test setup.

### DIFF
--- a/test/include/alpakaTest/deviceHelper.hpp
+++ b/test/include/alpakaTest/deviceHelper.hpp
@@ -1,0 +1,61 @@
+/* Copyright 2026 Simeon Ehrig
+ * SPDX-License-Identifier: ISC
+ */
+
+#pragma once
+
+#include <alpaka/api/trait.hpp>
+#include <alpaka/onHost/Device.hpp>
+#include <alpaka/onHost/DeviceSelector.hpp>
+
+#include <catch2/catch_test_macros.hpp>
+
+#include <tuple>
+
+namespace alpaka::test
+{
+    /** Takes a test configuration and create the device 0.
+     *
+     * If no device is available, skip the test. Prints information about the device Spec, API and device.
+     *
+     * @param cfg Test configuration. An entry of the list returned from alpaka::onHost::allBackends().
+     * @return The device 0.
+     */
+    auto getDevice(auto cfg)
+    {
+        auto deviceSpec = cfg[object::deviceSpec];
+        auto devSelector = onHost::makeDeviceSelector(deviceSpec);
+        UNSCOPED_INFO("DeviceSpec: " << getName(deviceSpec));
+        UNSCOPED_INFO("API: " << deviceSpec.getApi().getName());
+
+        if(!devSelector.isAvailable())
+        {
+            /* The REQUIRE_FALSE is required because otherwise all test cases in a file can be skipped because of a
+            missing device and now check was performed. In the case the Catch2 will fail because it assumes that at
+            least one check is done in test file. */
+            REQUIRE_FALSE(devSelector.isAvailable());
+            SKIP("No device available for " << deviceSpec.getName());
+        }
+
+        onHost::Device device = devSelector.makeDevice(0);
+        UNSCOPED_INFO("Device: " << device.getName());
+        return device;
+    }
+
+    /** Takes a test configuration and create the device 0 and an executor.
+     *
+     * If no device is available, skip the test. Prints information about the device Spec, API, device and
+     * executor.
+     *
+     * @param cfg Test configuration. An entry of the list returned from alpaka::onHost::allBackends().
+     * @return A std::tuple device and executor. Can be assigned to a structure binding (auto[device, exec]).
+     */
+    auto getDeviceExecutor(auto cfg)
+    {
+        onHost::Device device = alpaka::test::getDevice(cfg);
+        concepts::Executor auto executor = cfg[object::exec];
+        UNSCOPED_INFO("Executor: " << executor.getName());
+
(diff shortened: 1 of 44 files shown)


test/include/alpakaTest/deviceHelper.hpp
```diff
@@ -0,0 +1,57 @@
...
+namespace alpaka::test
+{
+    /** Takes a test configuration and create the device 0.
+     *
+     * If no device is available, skip the test. Prints information about the device Spec, API and device.
+     *
+     * @param cfg Test configuration. An entry of the list returned from alpaka::onHost::allBackends().
+     * @return The device 0.
+     */
+    auto get_device(auto cfg)
```
psychocoderHPC: Please change to camelCase naming, same for the other function.

psychocoderHPC (changes requested):

test/include/alpakaTest/deviceHelper.hpp
SimeonEhrig: done

test/include/alpakaTest/deviceHelper.hpp
```diff
@@ -0,0 +1,65 @@
...
+        auto device = alpaka::test::getDevice(cfg);
+        if(!device)
+        {
+            return {};
+        }
+        concepts::Executor auto executor = cfg[object::exec];
+        UNSCOPED_INFO("Executor: " << executor.getName());
+
+        return std::make_tuple(*device, executor);
+    }
```
psychocoderHPC, line 64: we could define a enum class here to use `std::get<0>()` with names

```C++
enum class Foo : std::size_t
{
    device = 0,
    executor = 1
};

onHost::Device computeDev = std::get<foo::device>(*optionalDeviceExec);
concepts::Executor auto exec = std::get<foo::executor>(*optionalDeviceExec);
```

test/unit/algorithm/iota.cpp
```diff
@@ -92,22 +93,13 @@ template<typename T_DataType>
...
 {
     using DataType = T_DataType;
-
-    auto deviceSpec = cfg[object::deviceSpec];
-    alpaka::concepts::Executor auto exec = cfg[object::exec];
-
-    auto computeDevSelector = onHost::makeDeviceSelector(deviceSpec);
-    if(!computeDevSelector.isAvailable())
+    auto optionalDeviceExec = test::getDeviceExecutor(cfg);
+    if(!optionalDeviceExec)
```
psychocoderHPC, line 97: discussed offline, should be 

```C++
 if(!optionalDeviceExec)
        return;
```

psychocoderHPC (changes requested):

test/include/alpakaTest/deviceHelper.hpp
SimeonEhrig: I implement helper functions. I renamed `test::getDevice` to `test::getAvailableDevice` and `test::getDeviceExecutor` to `test::getAvailableDeviceExecutor` and implemented the helper functions `test::getDevice` and `test::getExecutor`. `test::getDevice` is overloaded for an `std::optinal<Device>` and `std::optinal<std::tuple<Device,Execetuor>>`.

 https://github.com/SimeonEhrig/alpaka2/blob/f632b7617504bab57ddb56139de3cf10c505a3ce/test/unit/algorithm/transformReduce.cpp#L160-L164

test/include/alpakaTest/deviceHelper.hpp
```diff
@@ -0,0 +1,98 @@
...
+    /** Takes std::optional with an alpaka::onHost::Device and an executor and returns the executor.
+     *
+     * @attention Does not check, if the optional is valid.
+     */
+    template<alpaka::concepts::Api T_Api, alpaka::concepts::DeviceKind T_DeviceKind, alpaka::concepts::Executor T_Exec>
+    [[nodiscard]] T_Exec getExecutor(
+        std::optional<std::tuple<alpaka::onHost::Device<T_Api, T_DeviceKind>, T_Exec>> const& optionalDeviceExec)
+    {
+        return std::get<1>(*optionalDeviceExec);
+    }
```
psychocoderHPC, line 97:
```suggestion
    template<alpaka::concepts::Api T_Api, alpaka::concepts::DeviceKind T_DeviceKind>
    [[nodiscard]] alpaka::concepts::Executor auto getExecutor(
        std::optional<std::tuple<alpaka::onHost::Device<T_Api, T_DeviceKind>, T_Exec>> const& optionalDeviceExec)
    {
        return std::get<1>(*optionalDeviceExec);
    }
```

I am wondering that the return type can be in the template signature and does not need to be explicit set at the caller side. I would typically write it like I suggested. Thats also how most code is in alpaka is written.

psychocoderHPC (approved):
not mandatory but I would not use a template argument only to say the return is following a concepts. We try to keep the number of named template arguments low if possible.

test/include/alpakaTest/deviceHelper.hpp
psychocoderHPC: ahh ok, I missed that `T_Exec` is part of the tuple and thats why you have the template parameter.
So ignore my suggestion.

SimeonEhrig:
> not mandatory but I would not use a template argument only to say the return is following a concepts. We try to keep the number of named template arguments low if possible.

I'm not sure, what you mean. If you mean the template parameter of `getDevice` and `getExecutor`, I need it to well define the overload and because they are there, I also use it for the return type, because it well defines the relationship between input and output and it helps compiler and tools to better understand the function.

PR #519, opened 2026-05-05
Add SIMD feature `simdize` and benchmark HACCmk

psychocoderHPC:
- introduce simdize motivated by https://ieeexplore.ieee.org/document/11207437
- add support for simdized data types to `SimdAlgo::transformReduce`
- add HACCmk benchmark basd on the ORNL HACCmk version

### DIFF
--- a/benchmark/HACCmk/src/helper.hpp
+++ b/benchmark/HACCmk/src/helper.hpp
@@ -67,26 +67,26 @@ namespace hacc
 
     int verify(int n2, auto vx2, auto vy2, auto vz2, auto vx2_hw, auto vy2_hw, auto vz2_hw)
     {
-        int error = 0;
+        int error = EXIT_SUCCESS;
         float const eps = 1.0f;
         for(int i = 0; i < n2; i++)
         {
             if(fabsf(vx2[i] - vx2_hw[i]) > eps)
             {
                 printf("error at vx2[%d] %f %f\n", i, vx2[i], vx2_hw[i]);
-                error = 1;
+                error = EXIT_FAILURE;
                 break;
             }
             if(fabsf(vy2[i] - vy2_hw[i]) > eps)
             {
                 printf("error at vy2[%d]: %f %f\n", i, vy2[i], vy2_hw[i]);
-                error = 1;
+                error = EXIT_FAILURE;
                 break;
             }
             if(fabsf(vz2[i] - vz2_hw[i]) > eps)
             {
                 printf("error at vz2[%d]: %f %f\n", i, vz2[i], vz2_hw[i]);
-                error = 1;
+                error = EXIT_FAILURE;
                 break;
             }
         }
--- a/benchmark/HACCmk/src/helper.hpp
+++ b/benchmark/HACCmk/src/helper.hpp
@@ -1,0 +1,97 @@
+/* Copyright (c) 2020 , Argonne National Laboratory   (Zheming Jin)
+ * Copyright (c) 2020-, Oak Ridge National Laboratory   (Zheming Jin)
+ * Copyright 2026 René Widera
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+
+
+/** @file The file is providing some helpers for the HACCmk implementation shared between the original and the alpaka
+ * implementation.
+ */
+
+#include <math.h>
+#include <stdio.h>
+
+namespace hacc
+{
+    void haccmk_gold(
+        int count1,
+        float xxi,
+        float yyi,
+        float zzi,
+        float fsrrmax2,
+        float mp_rsm2,
+        auto xx1,
+        auto yy1,
+        auto zz1,
+        auto mass1,
+        auto dxi,
+        auto dyi,
+        auto dzi)
+    {
+        float const ma0 = 0.269327f, ma1 = -0.0750978f, ma2 = 0.0114808f, ma3 = -0.00109313f, ma4 = 0.0000605491f,
+                    ma5 = -0.00000147177f;
+
+
+        float dxc, dyc, dzc, m, r2, f, xi, yi, zi;
+
+        xi = 0.f;
+        yi = 0.f;
+        zi = 0.f;
+
+        for(int j = 0; j < count1; j++)
+        {
+            dxc = xx1[j] - xxi;
+            dyc = yy1[j] - yyi;
+            dzc = zz1[j] - zzi;
+
+            r2 = dxc * dxc + dyc * dyc + dzc * dzc;
+
+            if(r2 < fsrrmax2)
+                m = mass1[j];
+            else
+                m = 0.f;
+
+            f = r2 + mp_rsm2;
+            f = m * (1.f / (f * sqrtf(f)) - (ma0 + r2 * (ma1 + r2 * (ma2 + r2 * (ma3 + r2 * (ma4 + r2 * ma5))))));
+
+            xi = xi + f * dxc;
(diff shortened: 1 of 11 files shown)


benchmark/HACCmk/src/main_alpaka.cpp
```diff
@@ -0,0 +1,345 @@
...
+        int i = 0;
+
+
+        /* Initial data preparation */
+        fcoeff = 0.23f;
+        fsrrmax2 = 0.5f;
+        mp_rsm2 = 0.03f;
+        dx1 = 1.0f / (float) n2;
+        dy1 = 2.0f / (float) n2;
+        dz1 = 3.0f / (float) n2;
```
SimeonEhrig:
```suggestion
        dx1 = 1.0f / static_cast<float>(n2);
        dy1 = 2.0f / static_cast<float>(n2);
        dz1 = 3.0f / static_cast<float>(n2);
```

benchmark/HACCmk/src/main_alpaka.cpp
```diff
@@ -0,0 +1,345 @@
...
+        yy[0] = 0.f;
+        zz[0] = 0.f;
+        mass[0] = 2.f;
+
+        for(i = 1; i < n2; i++)
+        {
+            xx[i] = xx[i - 1] + dx1;
+            yy[i] = yy[i - 1] + dy1;
+            zz[i] = zz[i - 1] + dz1;
+            mass[i] = (float) i * 0.01f + xx[i];
```
SimeonEhrig:
```suggestion
            mass[i] = static_cast<float>(i * 0.01f + xx[i]);
```

benchmark/HACCmk/src/helper.hpp
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        *dxi = xi;
+        *dyi = yi;
+        *dzi = zi;
+    }
+
+    int verify(int n2, auto vx2, auto vy2, auto vz2, auto vx2_hw, auto vy2_hw, auto vz2_hw)
+    {
+        int error = 0;
```
SimeonEhrig:
```suggestion
        bool error = 0;
```

I suggest to use bool, also in the rest of the function.

benchmark/HACCmk/src/misc.hpp
```diff
@@ -0,0 +1,106 @@
...
+
+        return EXIT_SUCCESS;
+    }
+
+    /** allocated aligned memory
+     *
+     * @return a pointer to the memory section of the given type T. The pointer is aligned to 256 byte.
+     */
+    template<typename T>
+    inline auto* allocAligned(int numParticles)
```
SimeonEhrig: I this provide by the original benchmark? I thought `alpaka::onHost::alloc()` does already aligned allocation.

include/alpaka/onAcc/internal/SimdTransformReduce.hpp
```diff
@@ -70,24 +71,26 @@ namespace alpaka::onAcc::internal
                 IdxRange{numElements},
                 asParent().getTraversePolicy(),
                 asParent().getIdxLayoutPolicy());
-            using ReturnType = decltype(func(
-                acc,
-                SimdPtr{data0, *(traverse.begin()), T_MemAlignment{}, CVec<uint32_t, 1u>{}},
-                SimdPtr{dataN, *(traverse.begin()), T_MemAlignment{}, CVec<uint32_t, 1u>{}}...));
 
-            auto retValue = ReturnType::fill(neutralElement);
+            using SimdOneReturnType = ALPAKA_TYPEOF(makeSimdized<1u>(neutralElement));
```
SimeonEhrig, line 75: Is `makeSimdized<1u>(neutralElement)` the number of elements or the simd width? If last one is correct, can we assume that the width is 1 all the time?

include/alpaka/onAcc/SimdAlgo.hpp
```diff
@@ -261,7 +250,16 @@ namespace alpaka::onAcc
         }
 
         /**
+         * @copydoc transformReduce()
+         *
          * @param extents number of elements to process in each dimension
+         * @param T_maxConcurrencyInByte
```
SimeonEhrig:
```suggestion
         * @tparam T_maxConcurrencyInByte
```

include/alpaka/onAcc/SimdAlgo.hpp
```diff
@@ -261,7 +250,16 @@ namespace alpaka::onAcc
...
+         *    Maximum number of bytes to be used for concurrency.
+         *    Concurrency bytes describe a virtual simd pack size which is not exceeded.
+         *    Internally a best fitting SIMD width is calculated and instruction parallelism is exposed based on
+         *    T_maxConcurrencyInByte.
+         * @param T_MemAlignment alignment of the memory, if no alignments is given the alignment will be derived from
```
SimeonEhrig:
```suggestion
         * @tparam T_MemAlignment alignment of the memory, if no alignments is given the alignment will be derived from
```

benchmark/HACCmk/src/helper.hpp
psychocoderHPC: should be int but I should use `EXIT_SUCCESS` and `EXIT_FAILURE` insteat of magic integers.
This code is one-to-one copied from the original code.

benchmark/HACCmk/src/misc.hpp
psychocoderHPC: No, therefore it is in the `misc.hpp` which has a different license. Forgot to write it in the PR description.
In the original code `malloc` was sued.

Alignment is required for a fair comparison because if you use SIMD you need to guarantee that your pointer is aligned. This means if the auto vectorizer is vectorizing the loops if will not execute a `preamble loop, simd loop, and post loop`. It will only run `simd loop, and post loop`

include/alpaka/onAcc/internal/SimdTransformReduce.hpp
psychocoderHPC: It is the SIMD width (aka lanes) which should come out of the `simdize` type transformation

```C++
   /** ...
     * @tparam T_width the width of the used SIMD type
     * @return A simdized data type where each lane replicates the given value. If `makeSimdized` is not specialized
     * for the given type a SIMD pack wrapping the input value will be returned.
     */
    template<uint32_t T_width>
    constexpr auto makeSimdized(auto&& value);
```

benchmark/HACCmk/src/main_alpaka.cpp
psychocoderHPC: Yes I thnk I can do this code change, the C type casts is from the original code, I took it 1:1 without changes.

The correct change should be, else `i` will be implicit casted due to the multiplication with float.
```
 mass[i] = static_cast<float>(i) * 0.01f + xx[i];
```

include/alpaka/onAcc/internal/SimdTransformReduce.hpp
SimeonEhrig: Last question for general understanding. What is the unit of `T_width`? Is it something like bytes or number of elements per pack or something different?

psychocoderHPC: it is number of elements/scalars. In the language domain of SIMD e.g. std::simd it is called width and sometimes you find the word lane.

psychocoderHPC:
I applied all review suggestions

SimeonEhrig (approved):

PR #520, opened 2026-05-06
permit scalar extent for `getFrameSpec` 

TimHanel00:
The current `getFrameSpec` implementation should accept scalar extents, since the manual construction of scalar `FrameSpec`s is already permitted. 

The PR introduces thin wrapper functions for dispatching vector and scalar extents - minimizing code duplication at the cost of two copy operation in the case a scalar extent was used. 

Since both the `alpaka::Vector`  and `std::integral` concepts do not permit `lvalues` aka references -  I opted to use this pattern: 

`typename T_Extent, std::integral  = std::decay_t<T_Extent>` to apply constraints to the bare type - as I did not want to use `requires`.

This way both lvalue and rvalue extents inputs remain possible. 

Please flag this, if you have a cleaner alternative in mind or if you think I should rather use a different Implementation for scalar types - and scrap this naive copy-wrapper/`Vector`-cast approach.

### DIFF
--- a/include/alpaka/onHost/internal/interface.hpp
+++ b/include/alpaka/onHost/internal/interface.hpp
@@ -1,11 +1,11 @@
-/* Copyright 2024 René Widera
+/* Copyright 2024 René Widera, Tim Hanel
  * SPDX-License-Identifier: MPL-2.0
  */
 
 #pragma once
-
 #include "alpaka/KernelBundle.hpp"
 #include "alpaka/api/trait.hpp"
+#include "alpaka/core/Assert.hpp"
 #include "alpaka/core/common.hpp"
 #include "alpaka/onHost/DeviceProperties.hpp"
 #include "alpaka/onHost/FrameSpec.hpp"
@@ -495,23 +495,28 @@ namespace alpaka::onHost
             return GetPitches::Op<ALPAKA_TYPEOF(*any.get())>{}(*any.get());
         }
 
-        /** get a SIMD optimized frame spec
+        /** implementation to get a SIMD optimized frame spec
          *
          * @param internalDevice must be a alpaka internal device implementation
          */
         template<typename T_DataType>
-        inline constexpr auto getFrameSpec(auto const& internalDevice, auto&& extents)
+        inline constexpr auto getFrameSpec(
+            auto const& internalDevice,
+            alpaka::concepts::VectorOrScalar auto const& extents)
         {
+            Vec extentMd = extents;
             auto deviceKind = alpaka::internal::getDeviceKind(internalDevice);
             auto deviceApi = alpaka::internal::getApi(internalDevice);
-            using ExtentVecType = ALPAKA_TYPEOF(extents);
+            using ExtentVecType = ALPAKA_TYPEOF(extentMd);
+            // check that all extent dimensions are greater than zero
+            ALPAKA_ASSERT((extentMd > ExtentVecType::fill(0u)).reduce(std::logical_and{}));
             using IndexType = alpaka::trait::GetValueType_t<ExtentVecType>;
             auto props = internal::GetDeviceProperties::Op<ALPAKA_TYPEOF(internalDevice)>{}(internalDevice);
             IndexType warpSize = static_cast<IndexType>(props.warpSize);
             // try to create a specification with a frame size of 512 elements
             IndexType numFrameElements = 512;
             // avoid non-power of two values
-            auto fastDimensionValue = roundDownToPowerOfTwo(std::min(warpSize, extents.x()));
+            auto fastDimensionValue = roundDownToPowerOfTwo(std::min(warpSize, extentMd.x()));
             auto frameExtents = ExtentVecType::fill(1).rAssign(fastDimensionValue);
             numFrameElements /= frameExtents.x();
             // distribute remainder frame elements
@@ -521,7 +526,7 @@ namespace alpaka::onHost
                 IndexType maxValue = 0;
                 for(auto i = 0u; i < ExtentVecType::dim(); ++i)
                 {
-                    auto v = extents[i] / frameExtents[i] / IndexType{2};
+                    auto v = extentMd[i] / frameExtents[i] / IndexType{2};
                     if(maxValue < v)
                     {
                         maxIdx = i;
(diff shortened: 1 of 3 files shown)


SimeonEhrig:
@TimHanel00 Can you please provide a small unit test. I've an idea how to solve the problem. 
I think we implement the `std::decay` in `alpaka::concepts::Vector` because I see no reason why we need to distinguish between a lvalue, a reference or something else. For `std::integral`, we can simply create a wrapper concept.

If this is working, we can write `template<typename T_DataType, alpaka::concepts::Vector T_Extent>` and `template<typename T_DataType, alpaka::concepts::integral T_Extent>`.

include/alpaka/onHost/internal/interface.hpp
```diff
@@ -541,8 +541,29 @@ namespace alpaka::onHost
...
+            auto frameSpec = FrameSpec{numFrames, frameExtents};
             return frameSpec;
         }
+
+        /** get a SIMD optimized frame spec using a vector extent
+         *
+         * @param internalDevice must be a alpaka internal device implementation
+         */
+        template<typename T_DataType, typename T_Extent, alpaka::concepts::Vector = std::decay_t<T_Extent>>
+        inline constexpr auto getFrameSpec(auto const& internalDevice, T_Extent&& extents)
```
psychocoderHPC: Please use a single interface function:

https://github.com/alpaka-group/alpaka3/blob/4601bbe9b07146293ecefd544ba3a757ae3dfe42/include/alpaka/mem/View.hpp#L179-L182

```C++
alpaka::concepts::VectorOrScalar auto const& extents
```
This support now a scalar or a vector. If a scaar will be passed it will be transformed in the next line to an zero dimensional vector.

```C++
Vec extentMd = extents; 
```

Within the calculations, you use `extentMd` now instead of the user-provided `extents`.

psychocoderHPC (changes requested):
Simplification of the interface to match other *alpaka* interfaces.

include/alpaka/onHost/internal/interface.hpp
```diff
@@ -541,8 +541,29 @@ namespace alpaka::onHost
...
+        {
+            return getFrameSpecImpl<T_DataType>(internalDevice, extents);
+        }
+
+        /** get a SIMD optimized frame spec using a scalar extent
+         *
+         * @param internalDevice must be a alpaka internal device implementation
+         */
+        template<typename T_DataType, typename T_Extent, std::integral = std::decay_t<T_Extent>>
```
psychocoderHPC: btw way the more modern version of your code would be 

```C++
template<typename T_DataType, typename T_Extent>
inline constexpr auto getFrameSpec(auto const& internalDevice, T_Extent&& extents) requires ( std::integral<std::decay_t<T_Extent>>)
{}
```

include/alpaka/onHost/internal/interface.hpp
```diff
@@ -495,23 +495,38 @@ namespace alpaka::onHost
...
+                [&]
+                {
+                    for(auto i = 0u; i < ExtentVecType::dim(); ++i)
+                    {
+                        if(extentMd[i] <= 0)
+                            return false;
+                    }
+
+                    return true;
+                })());
```
psychocoderHPC:
```suggestion
            ALPAKA_ASSERT((extentMd > ExtentVecType::fill(0u)).reduce(std::logical_and{}));
```

psychocoderHPC (changes requested):
I will check the test next week

test/unit/frame/getFrameSpec.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include <type_traits>
+#include <utility>
+
+
+using namespace alpaka;
+
+using TestBackends = std::decay_t<decltype(onHost::allBackends(onHost::enabledDeviceSpecs, exec::enabledExecutors))>;
+template<typename T>
+using ID = std::type_identity<T>;
+using TestDataTypes = std::tuple<ID<double>, ID<float>, ID<int>, ID<uint32_t>, ID<uint64_t>, ID<std::size_t>>;
```
psychocoderHPC:
```suggestion
using TestDataTypes = std::tuple<ID<int>, ID<uint32_t>, ID<uint64_t>, ID<std::size_t>>;
```

We should not test floating point values, you need to deal with floating point errors, ...
If a user is reading the test he should not get the impression of using floating point numbers.

test/unit/frame/getFrameSpec.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include <tuple>
+#include <type_traits>
+#include <utility>
+
+
+using namespace alpaka;
+
+using TestBackends = std::decay_t<decltype(onHost::allBackends(onHost::enabledDeviceSpecs, exec::enabledExecutors))>;
+template<typename T>
+using ID = std::type_identity<T>;
```
psychocoderHPC: Why do you use `ID`? There should be no need, you could simply use the data types in the tuple.

test/unit/frame/getFrameSpec.cpp
psychocoderHPC: maybe `IdxTestType` is a better name compared to `TestDataTypes`. It is not about data thisis the index type used by frame spec.

psychocoderHPC (changes requested):
Small changes in the tests are required. All other is fine.

test/unit/frame/getFrameSpec.cpp
```diff
@@ -0,0 +1,111 @@
...
+        CHECK(numFrames[d] > ElemType{0});
+        CHECK(frameExtents[d] > ElemType{0});
+        CHECK(numFrames[d] * frameExtents[d] <= extentMd[d]);
+    }
+}
+
+template<typename T>
+void testScalar(auto const& computeDev)
+{
+    for(auto sampleElements : {1, 2, 31, 128, 129, 257, 513, 10000, 100000})
```
SimeonEhrig: Catch2 provides a feature therefore: https://github.com/catchorg/Catch2/blob/devel/docs/generators.md

If a test fails, there will be a much better and meaningful error message.

test/unit/frame/getFrameSpec.cpp
TimHanel00: If the user is using a floating point buffer, which is often the case in scientific applications - the SIMD optimized `FrameSpec` provided by `getFrameSpec`, has to account for the actual datatype of the buffer elements (and their size) -`T_DataType` in `getFrameSpec` is not an integer index type and has nothing to do with the type of the extent that is additionally provided. Therefore I think it makes sense to include foating point types in this test-dimension. I did however change its name to be more descriptive and indicate that this is about data-buffer element types rather then some Index-types.

psychocoderHPC: You are right, I thought it is the index_type, all fine you gave it now a proper name.

psychocoderHPC (approved):

PR #522, opened 2026-05-11
overworked advanced tutorial `Views and Subviews`, `Kernel - Parallelism` and `Working With Multidimensional Kernels`

SimeonEhrig:
~- [ ] overwork `Kernel - Parallelism`~

Overworking the `Kernel - Parallelism` takes to much time. I will do since I used makeIdxMap somewhere in future.

### DIFF
--- a/docs/snippets/example/080_multidim.cpp
+++ b/docs/snippets/example/080_multidim.cpp
@@ -66,6 +66,7 @@ TEMPLATE_LIST_TEST_CASE("tutorial multidimensional stencil kernel", "[docs]", do
     // END-TUTORIAL-multidimFrameSpec
 
     // BEGIN-TUTORIAL-multidimKernelLaunch
+    static_assert(frameSpec.dim() == 2);
     queue.enqueue(frameSpec, KernelBundle{FivePointAverageKernel{}, outBuffer, inBuffer});
 
     onHost::memcpy(queue, hostOutput, outBuffer);
--- a/docs/snippets/example/080_multidim.cpp
+++ b/docs/snippets/example/080_multidim.cpp
@@ -26,6 +26,9 @@ struct FivePointAverageKernel
 
         for(auto idx : onAcc::makeIdxMap(acc, onAcc::worker::threadsInGrid, IdxRange{extents}))
         {
+            /* Copy only values in the halo region
+            idx.y() == 0u -> top halo, idx.x() == 0u -> left halo
+            idx.y() + 1u == extents.y() -> bottom halo, idx.x() + 1u == extents.x() -> right halo */
             if(idx.y() == 0u || idx.x() == 0u || idx.y() + 1u == extents.y() || idx.x() + 1u == extents.x())
             {
                 out[idx] = in[idx];
--- a/docs/source/tutorial/multidim.rst
+++ b/docs/source/tutorial/multidim.rst
@@ -24,7 +24,7 @@ Typical use cases for different dimensions are:
 - 1D frames for flat vectors and simple reductions.
 - 2D frames for images, matrices, and most stencil codes.
 - 3D frames for volumetric problems, such as the position of particles in a space.
-- ND frames for algorithms that have more than three natural dimensions, such as position coordinate plus a time stamp.
+- ND frames for algorithms that have more than three natural dimensions, such as position coordinates and discretized time, e.g., to build a layered detector within our particle simulation.
 
 Keep in mind that the rightmost index, usually ``x``, is the fastest varying dimension in *alpaka* buffers.
 This is important if you are using hand written ``for``-loops to iterate over the data instead of ``makeIdxMap``.
@@ -32,7 +32,7 @@ This is important if you are using hand written ``for``-loops to iterate over th
 A Small 2D Stencil Example
 --------------------------
 
-The following kernel performs one five-point average step on a small 2D grid.
+The following kernel performs one five-point average step on a small 2D grid [#f1]_.
 This introduces three important ideas at once:
 
 - Iterating over multidimensional buffers.
@@ -51,13 +51,20 @@ The structure is still the same as in the one-dimensional tutorial:
 - Build ``IdxRange{extents}`` to describe the full valid multidimensional index domain.
 - Map worker threads via ``makeIdxMap`` to the range.
 - Guard the boundary cells, to avoid out-of-bounds accesses.
+- The own element is updated depending on the neighbor data and maybe the own element. Indices are relative calculated compare to the index of the own output coordinate.
 - Update neighbor locations by adding or subtracting direction vectors from the current ``Vec`` index.
 
 This is the natural alpaka style for stencil code.
 The project examples, such as the heat-equation stencil, operate on the multidimensional index directly and move to
 neighbors with vector offsets instead of splitting ``x`` and ``y`` into separate scalars and rebuilding indices.
-Handling the boundaries within the kernel which performs computation on the inner region of the domain is not good for performance, but it is the easiest and best way to start writing a correct kernel.
+
+At the beginning, the range check of data access should be done within the kernel.
+While this approach does not offer the best performance, it is easier to implement correctly.
+Subsequently, the range check should be moved from the kernel to before the kernel launch, and the kernel should operate on the assumption that the range is correct in order to improve performance.
+
 If you want to optimize the boundary handling later, you can write a separate kernel for the border and launch it with a different frame shape that only covers the halo/guard region.
+
+.. [#f1] https://en.wikipedia.org/wiki/Iterative_Stencil_Loops#Stencils
 
 Launching the 2D Kernel
 -----------------------
--- a/docs/source/tutorial/multidim.rst
+++ b/docs/source/tutorial/multidim.rst
@@ -1,9 +1,10 @@
 Working With Multidimensional Kernels
 =====================================
 
-Many important beginner examples in parallel computing are naturally multidimensional:
+Many important examples in parallel computing are naturally multidimensional:
 images, matrices, heat diffusion, cellular automata, and finite-difference stencils.
-For these kinds of problems, it is usually clearer to keep the kernel multidimensional instead of flattening everything into one linear index.
+*alpaka* supports multi-dimensional kernels and memory.
+That's why it's usually better to implement these types of problems using multiple dimensions rather than flattening everything into a single linear index.
 
 Choose the Kernel Shape From the Data
 -------------------------------------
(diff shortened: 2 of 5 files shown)


docs/source/tutorial/views.rst
SimeonEhrig, line 13: @psychocoderHPC I'm not sure, if this is a good example. Normally we use `MdSpans` therefore.

docs/source/tutorial/multidim.rst
SimeonEhrig: @psychocoderHPC Please describe a little more, what the algorithm should do. It is hard to understand.

If I'm not wrong, you have a 2D grid and you select an element with the coordinate (y,x). You take the values (y-1, x), (y+1, x), (y, x+1), (y, x-1) and add it to the value of position (y,x). Than divide it by 5 to get the average of all values.

docs/snippets/example/080_multidim.cpp
SimeonEhrig, line 32: @psychocoderHPC Are this boundary checks?

docs/source/tutorial/multidim.rst
SimeonEhrig, line 55: @psychocoderHPC I think this sentence is wrong. Normally you update only our own position and not neighbors.

SimeonEhrig: @psychocoderHPC I don't understand the sentence. Most of the time, boundary checks are required to process real world problem.

Maybe you mean something like processing a vector with SIMD commands. Therefore divide the vector in chunks of the size of our SIMD unit plus a residual. If you process the SIMD chunks, no boundary check is required because it is done before. Only processing the residual requires boundary checks.

docs/source/tutorial/multidim.rst
```diff
@@ -75,9 +77,8 @@ What Users Usually Need To Know Early
 The following habits are worth learning from the start:
 
 - Keep boundary handling explicit. It is fine to have branching for the border in the first implementation.
-- Iterate over the full valid problem range, not over guessed thread ids. Let alpaka deal with the mapping from threads to your problem.
```
SimeonEhrig, line 78: The is part of the chunking page.

docs/snippets/example/080_multidim.cpp
psychocoderHPC: yes this is a boundary check else the stencil will access outside of the valid memory range.

In the tutorial, a disclaimer is following:

```
Handling the boundaries within the kernel that performs computation on the inner region of the domain is not good for performance, but it is the easiest and best way to start writing a correct kernel. If you want to optimize the boundary handling later, you can write a separate kernel for the border and launch it with a different frame shape that only covers the halo/guard region.
```

docs/source/tutorial/multidim.rst
psychocoderHPC: It is a 5 point stencil. I do not know what else to add; it is a well-known computing pattern.

```
If I'm not wrong, you have a 2D grid and you select an element with the coordinate (y,x). You take the values (y-1, x), (y+1, x), (y, x+1), (y, x-1) and add it to the value of position (y,x). Then divide it by 5 to get the average of all values.
```

Yes, this is a typical `five-point average` stencil. The equations is not important, you can use any stencil, even with weightings for each neighbor. This here is a very easy example; the stencil depends on the problem the user would like to solve.

psychocoderHPC: You are right but you reference an outdated version, it is

```
Update neighbor locations by adding or subtracting direction vectors from the current Vec index.
```

psychocoderHPC: No this kernel has nothing to do with SIMD.
In a stencil algorithm, you have an inner reagion/range where you can apply the stencil functor without out-of-bounce checks and you have boundaries. The stencil in the boundaries looks different, sometimes you skip simply not existing values, or you apply a completely different functor.
It is better to have our own kernel for the boundary but for simplicity, this is not done in this example, because it "only" affects the performance and not the correctness.
Another reason for having a separate kernel is that if you have different stencil functors for the boundary and inner part, you could increase the register footprint because the boundary stencil is often more complex.

docs/source/tutorial/multidim.rst
psychocoderHPC: It has nothing to do with chunking; it is the repetition that the user should not start dealing with thread indices. IMO it is very important, everyone who comes from CUDA/HIP will start doing complex thread ID operations.
The reason why it is mentioned multiple times in the tutorial is that this should be avoided.

docs/source/tutorial/views.rst
psychocoderHPC: Both is possible, if you pass a `Buffer` into a kernel it will be lowered to a view and not `MdSpan`, but the view is fulfilling the `MdSpan` interface concept.

Here we explain the user how to create subviews to be usable with `onHost::memcpy`, ... and forwardable into the kernel. That you need for a kernel only an `MdSpan` is not mentioned because it makes all more complex.

docs/source/tutorial/views.rst
```diff
@@ -48,8 +50,8 @@ That means you can, for example, allocate device memory based on a view and copy
 Typical use cases include:
 
 - Copying a subrange of a 1D vector.
-- Copying only the active interior of a ND grid.
-- Passing a tile into a helper function.
+- Only the inner area of the matrix containing the actual data, without the halo, should be copied into a stencil code.
```
psychocoderHPC: please keep `grid` and that it is multi dimensional because it points to `stencils` and `helo+grid` is common in this domain not `matrix`

The reason why `ND` is important is that most research codes using flat data structures and remap it within the algorithm. That all works multidimensionally is a selling point of *alapka*

docs/source/tutorial/views.rst
```diff
@@ -48,8 +50,8 @@ That means you can, for example, allocate device memory based on a view and copy
...
+- One step of your algorithm operates on only a subset of your data. Use a subview to create a chunk of your data and pass it to the function that implements that step of your algorithm. This simplifies the implementation of the algorithm, since it doesn't have to worry about whether it needs to process all the data or just a chunk of it.
```
psychocoderHPC: IMO this is a longer repetition of the point before. I assume that in our target audience, `HPC` users everywhere know what a stencil is.
If you attend a parallel programming course, you will nearly always deal with a heat equation or game of life.

docs/source/tutorial/multidim.rst
```diff
@@ -17,12 +18,13 @@ This avoids hand-written index decoding and makes boundary conditions easier to
...
-We recommend that the frame shape should follow the logical shape of the work:
+We recommend that the frame shape should follow the logical shape of the work.
+Typical use cases for different dimensions are:
 
 - 1D frames for flat vectors and simple reductions.
 - 2D frames for images, matrices, and most stencil codes.
-- 3D frames only when the algorithm is truly volumetric.
-- ND frames for algorithms that have more than three natural dimensions, such as high-dimensional grids.
+- 3D frames for volumetric problems, such as the position of particles in a space.
+- ND frames for algorithms that have more than three natural dimensions, such as position coordinate plus a time stamp.
```
psychocoderHPC:
```suggestion
- ND frames for algorithms that have more than three natural dimensions, such as position coordinates and discretized time, e.g., to build a layered detector within our particle simulation.
```

docs/source/tutorial/views.rst
SimeonEhrig: I know, that it is possible, but the context/heading is `View are typically used when:` and there is no place in the alpaka code base, where a view is used in the kernel, because nobody needs the information, which API was used.

We simply use MdSpan all the time, because it is enough. Therefore it not typically.

docs/source/tutorial/multidim.rst
SimeonEhrig: It does not change the actual meaning. In pseudo code, your sentence means:

```c++
auto my_idx = getGlobalIdx();
out[my_idx + 1] = in[my_idx] + in[my_idx - 1]
```

Normally a stencil code is doing something like:
```c++
auto my_idx = getGlobalIdx();
out[my_idx] = in[my_idx + 1] + in[my_idx - 1]
```

docs/source/tutorial/multidim.rst
SimeonEhrig: I mean the `Kernel - Parallelism` page. The chunking page is something different or not. I lost a little bit the overview about the documentation.

docs/source/tutorial/multidim.rst
SimeonEhrig: The SIMD was only a example to understand.

Okay, if I understand it correctly. This part of the documentation does not use stencil as example to explain how multi-dimensional kernel and memory works. It explains how to implement a stencil code in alpaka.

SimeonEhrig: Okay, I asked 3 people. Pavel could explain it. Richard and Alex had different ideas and I had no idea until I read the code. But I know alpaka. Therefore I could reconstruct the meaning from the code.

docs/source/tutorial/views.rst
SimeonEhrig: Than we remove this line, if it is repetition.

> Passing a tile into a helper function.
Original feedback, Richard: "I don't know what the sentence means."
Therefore I tried to give it a little bit more context.

docs/source/tutorial/multidim.rst
psychocoderHPC: Maybe

```
Indices relative to the index of the output coordinate are used to load the neighbor data.
```

I agree that `update` could suggest writing to the neighbor's data.

psychocoderHPC: > This part of the documentation does not use stencil as example to explain how multi-dimensional kernel and memory works

Both are going hand in hand. A multi-dimensional stencil is a typical example where you use multi-dimensional data, especially within our group.

docs/source/tutorial/multidim.rst
psychocoderHPC: `Kernel - Parallelism` -> grid, block, warp, and thread hierarchy and chunking play both on the same category.
In HPC, thread parallelism and memory locality have been around for ages. Data domain decomposition aka. chunking, is the concept to express data locality. 
Since we have GPUs more thread hierarchy levels are available within the compute device. The levels must be controlled explicitly by mapping the thread hierarchy layers to data chunks. Typically, with each level, you become more fine-grained until you have mostly a one-to-one mapping of single data elements to the threads.

docs/source/tutorial/multidim.rst
psychocoderHPC: https://en.wikipedia.org/wiki/Iterative_Stencil_Loops and a simple example https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life

docs/source/tutorial/views.rst
psychocoderHPC: Maybe we have different definitions of "used".
Nearly all examples using views but require at least the IMdSpan interface concept.
Currently an MdSpan can not be sub-sliced therefore within the document this is IMO correct.

psychocoderHPC (approved):

PR #524, opened 2026-05-12
add worker group `worker::allThreads`

psychocoderHPC:
`worker::allThreads` is a worker group with a single slot. Together with `onAcc:::makeIdxMap()`, all threads will get the same id's to process.

This is a long-planned feature to avoid complicated worker group generation e.g. `onAcc::WorkerGroup{numFrames.fill(0), numFrames.fill(1)}`
Until last week, I was not aware of how to implement it for multidimensional groups. 
Now I realized that we already have all dependencies with `WorkerGroup` which support a `unit` and `origin.`

### DIFF
--- a/include/alpaka/onAcc/WorkerGroup.hpp
+++ b/include/alpaka/onAcc/WorkerGroup.hpp
@@ -125,6 +125,13 @@ namespace alpaka::onAcc
         constexpr auto linearWarpsInGrid = WorkerGroup{origin::grid, unit::warps};
 
         constexpr auto linearBlocksInGrid = WorkerGroup{origin::grid, unit::blocks, linearized};
+
+        /** Represent the identity of the executor thread.
+         *
+         * All threads are in the same worker group.
+         * If used with onAcc::makeIdxMap(), any thread is getting all indices of the range.
+         */
+        constexpr auto allThreads = WorkerGroup{origin::thread, unit::threads};
     } // namespace worker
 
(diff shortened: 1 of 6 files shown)


include/alpaka/onAcc/WorkerGroup.hpp
```diff
@@ -125,6 +125,13 @@ namespace alpaka::onAcc
         constexpr auto linearWarpsInGrid = WorkerGroup{origin::grid, unit::warps};
 
         constexpr auto linearBlocksInGrid = WorkerGroup{origin::grid, unit::blocks, linearized};
+
+        /** Represent the identity of the executor thread.
+         *
+         * All threads are in the same worker group.
```
SimeonEhrig, line 131: Does it mean a race condition is possible, because all threads should process the same index?

So not something like:
```c++
my_id = get_id();
if (my_id == 0){
  // do the job
}
// all other threads does nothing and wait until thread 0 is finished.
onAcc::syncBlockThreads(all);

psychocoderHPC: If you let write all threads the same value than yes, you have a classical race. The same happens if you use `blocksInGrid` and let all threads in the group slot write the same value.

Take care if your `get_id` is the index of the slot in the group, than with `allThreads` the id will always be zero and therefore every thread which is evaluating the expression is executing it, this means the `if` will always be true.

SimeonEhrig (approved):

PR #526, opened 2026-05-18
CI: install GCC via bash script

SimeonEhrig:
Start to modularise the CI steps. Use bash scripts instead write everything in the GitHub Actions yaml. 

This PR contains a lot of infrastructure bash scripts for the modularization and ports only one CI step for testing purpose. Ported the GCC install step. This PR enables the following features:

- support GitHub Actions and GitLab CI
- enables setting environment variables to configure the job
- Add the `var_storage` functionality. It allows to write and read bash variables to file. Therefore scripts can be separated executed instead everything needs to be source stating from a top-level script.

# Notes

- The `configure.sh` has no functionality. It only verifies if the several install scripts stores the for CMake configure  required variables are stored.
- The `gitlab-ci.yml` contains only really basic jobs to test the scripts. Later the jobs will be generated with a generator.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+# for shfmt
+[*.sh]
+# like -i=4
+indent_style = space
+indent_size = 4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -155,6 +155,8 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v6
 
+      - name: job configuration environment variables
+        run: $APCI_ALPAKA_ROOT/script/ci/print_vars.sh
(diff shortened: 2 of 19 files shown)


.editorconfig
```diff
@@ -0,0 +1,5 @@
+# for shfmt
```
psychocoderHPC, line 1: out of interest, which IDE is reading this file?

psychocoderHPC (approved):

.editorconfig
SimeonEhrig: In our case, the CLI tool `shfmt`. It has no own config file. It uses this open standard.

Here is a list of supported IDE's: https://editorconfig.org/

PR #529, opened 2026-05-18
fix pre-commit for `shellcheck`

### LINKED ISSUE #528: pre-commit shell check issue
WIth https://github.com/alpaka-group/alpaka3/pull/525 shellcheck was introduced into the pre-commit hooks.
On a system without Docker, the following error is shown.

```
ShellCheck v0.11.0.......................................................An unexpected error has occurred: CalledProcessError: command: ('docker', 'system', 'info', '--format', '{{ json . }}')
return code: 1
stdout:
    Executable `docker` not found
stderr: (none)
Check the log at /home/widera/.cache/pre-commit/pre-commit.log
```

psychocoderHPC:
fix #528

The original version of the pre-commit for shellcheck required docker. Switch the version used to support a non docker dependent version.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,8 +43,8 @@ repos:
   rev: v3.13.1-1
   hooks:
     - id: shfmt
-- repo: https://github.com/koalaman/shellcheck-precommit
-  rev: v0.11.0
+- repo: https://github.com/shellcheck-py/shellcheck-py
+  rev: v0.11.0.1
   hooks:
     - id: shellcheck
       # required that shellcheck can follow sourced scripts


.pre-commit-config.yaml
```diff
@@ -43,8 +43,8 @@ repos:
   rev: v3.13.1-1
   hooks:
     - id: shfmt
-- repo: https://github.com/koalaman/shellcheck-precommit
-  rev: v0.11.0
+- repo: https://github.com/shellcheck-py/shellcheck-py
+  rev: v0.11.0.1
```
SimeonEhrig, line 47: We should not fix the patch level.
```suggestion
  rev: v0.11.0
```

https://github.com/shellcheck-py/shellcheck-py/tags

SimeonEhrig: It is not possible to this. There is no real version semantic implemented, which automatically pulls the latest patch release.

SimeonEhrig (approved):

PR #532, opened 2026-05-20
Print instructions to run CI job locally

SimeonEhrig:
At the beginning of a CI job, display instructions to run job locally via docker container. This feature is restricted on Linux because I have no Windows and MacOS test system and I assume it is not possible because if I'm not wrong there are no MacOS and Windows container. On this operations systems Linux VMs are started and inside the VM the Linux container are executed.

# GitLab CI

image

 On GitLab CI, empty lines are removed. Therefore the instructions are a little bit compressed. I also don't want to indent the bash commands for better readability because you should be able to copy and paste the instructions and an instruction with indent will not appear in the bash history.

# GitHub Actions

image


psychocoderHPC:
Please rebase against the dev. I updated the required jobs in github which defines if a PR can be merged.

psychocoderHPC (approved):

PR #533, opened 2026-05-20
refactor `makeIdxMap()` contigious mapping

psychocoderHPC:
Change the way how id's are distributed to worker group slots are distributes.
The new distribution is keeping the workload per slot more even and avoid slots without work.
Add a unit test for `makeIdxMap` to test all combinations of layout and traverse policy.

before PR:
image

after PR:
image


psychocoderHPC:
@SimeonEhrig I fixed the compile issues

SimeonEhrig (approved):

SimeonEhrig (approved):

PR #527, opened 2026-05-18
refactor `FrameSpec`, `ThreadSpec` and enqueue interface

psychocoderHPC:
- `FrameSpec`
  - The user specified frame specification is not forwared into the
   kernel.
  - A kernel can only query `acc.launchedWithFrameSpec()` to have information that the kernel start parameters and in kernel thread block configuration differ.
  - Is not holding a thread specification anymore.
  - Contains the executor and is therefore a launch configuration.
- ThreadSpec`: provides the used executor and is therefore a launch
  configuration.
- The executor is removed from the interface `queue.enqueue()`.

This PR is a **breaking** interface change, the number of frames and frame extents cann ot be used anymore within the kernel. If needed, the user is responsible for using explicit chunking and chunk sizes must be forwarded as arguments or kernel template parameters.

```C++
struct FooKernel
{
   ALPAKA_FN_ACC void operator(auto acc) const
  {
    for(auto elem :
            alpaka::onAcc::makeIdxMap(
                acc, 
                onAcc::worker::threadsInBlock, 
                onAcc::range::frameExtent))
            {
            }
  }
};
auto const frameSpec = FrameSpec{numChunks, chunkSize};
// ...
queue.enqueue(computeExec, frameSpec, FooKernel{});
```

became

```C++
struct FooKernel
{
   ALPAKA_FN_ACC void operator(auto acc, auto chunkExtents) const
  {
    for(auto elem :
            alpaka::onAcc::makeIdxMap(
                    acc, 
                    onAcc::worker::threadsInBlock, 
                    IdxRange{chunkExtents}))
            {
            }
  }
};

auto const frameSpec = FrameSpec{numChunks, chunkSize, computeExec};
queue.enqueue(frameSpec, FooKernel{}, chunkSize);
```

# Review

- the first commit is the changes required within alpaka
- all other commits using the new interface in different parts e.g. examples, docs, ...
- all places where frames were used within the kernel are replaced with explicit chunking and sometimes chunking was removed when it was not required.

PR #541, opened 2026-05-22
update dynamic shared memory interface

### LINKED ISSUE #540: BlockDynSharedMemBytes needs an interface update
With #527 the enqueue interface was updated bit we missed to update the operator() signature of BlockDynSharedMemBytes. It is not required anymore to pass the executor into the operator because the object is part of the thread specification.

psychocoderHPC:
fix #540

With PR #527 we forgot to update the dynamic shared memory traits. The executor is now part of the thread specification and is therefore not required as seperate parameter.
Keeping the old interface could lead in mistakes if the executor passed as argument differ from the executor in the thread specification.

### DIFF
--- a/docs/snippets/cheatsheet/cheatsheet.cpp
+++ b/docs/snippets/cheatsheet/cheatsheet.cpp
@@ -78,17 +78,17 @@ struct DynSharedMemTrait
 // specialization within the host code
 namespace alpaka::onHost::trait
 {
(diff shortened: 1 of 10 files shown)


psychocoderHPC:
@costeastefan @TimHanel00 if you use dynamic shared memory in your kernel, this upcoming change could affect your code too.

SimeonEhrig (approved):

PR #542, opened 2026-05-22
add math and atomic tutorial

psychocoderHPC:
- add tutorials sections
- add helper script to list math functions in the documentation

note: I tried using breath to list the math functions. It was looking horrible, therefore I decided to keep the xml parsing, which allows linking the math functions directly to the user documentation and avoid showing the full signature of each function.

### DIFF
--- a/docs/source/sphinx_helper/math_function_families.py
+++ b/docs/source/sphinx_helper/math_function_families.py
@@ -1,0 +1,64 @@
+"""Generate the math function reference block used by the tutorials."""
+
+from __future__ import annotations
+
+import pathlib
+import xml.etree.ElementTree as ET
+
+DOXYGEN_NAMESPACE_PAGE = "../doxygen/namespacealpaka_1_1math.html"
+
+
+def _load_function_data(xml_file_path: pathlib.Path) -> dict[str, tuple[str, int]]:
+    root = ET.parse(xml_file_path).getroot()
+
+    function_data: dict[str, tuple[str, int]] = {}
+    for member in root.findall(".//memberdef[@kind='function']"):
+        function_name = member.findtext("name")
+        member_id = member.get("id")
+        if function_name and member_id:
+            anchor = member_id.rsplit("_1", maxsplit=1)[-1]
+            function_data[function_name] = (f"{DOXYGEN_NAMESPACE_PAGE}#{anchor}", len(member.findall("param")))
+
+    return function_data
+
+
+def _group_functions(
+        function_data: dict[str, tuple[str, int]]
+) -> dict[str, list[str]]:
+    grouped_functions: dict[str, list[str]] = {
+        "Unary Functions": [],
+        "Binary Functions": [],
+        "Other Functions": [],
+    }
+
+    for function_name, (_, param_count) in sorted(function_data.items()):
+        if param_count == 1:
+            grouped_functions["Unary Functions"].append(function_name)
+        elif param_count == 2:
+            grouped_functions["Binary Functions"].append(function_name)
+        else:
+            grouped_functions["Other Functions"].append(function_name)
+
+    return grouped_functions
+
+
+def generate_math_function_families(app) -> None:
+    """Create the generated RST snippet referenced by the math tutorial."""
+
+    output_path = pathlib.Path(app.srcdir, "_generated", "math_function_families.rst")
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    xml_file_path = pathlib.Path(app.confdir).parent / pathlib.Path("doxygen", "xml", "namespacealpaka_1_1math.xml")
+
+    function_data = _load_function_data(xml_file_path)
+    grouped_functions = _group_functions(function_data)
+
+    lines: list[str] = []
+
+    for title, functions in grouped_functions.items():
+        if not functions:
+            continue
+        lines.extend(["", title, "-" * len(title), ""])
+        rendered_functions = [f"`{function_name} <{function_data[function_name][0]}>`_" for function_name in functions]
+        lines.append(", ".join(rendered_functions))
+
+    output_path.write_text("\n".join(lines) + "\n", encoding="utf-8")
--- a/docs/source/tutorial/atomics.rst
+++ b/docs/source/tutorial/atomics.rst
@@ -2,18 +2,14 @@ Atomics
 =======
 
 Atomics are the tool you reach for when several workers may update the same memory location.
-Typical examples are histograms, counters, sparse assembly, work lists, and some reductions.
-
-When to Use Atomics
--------------------
-
 Use an atomic operation only when two or more workers can hit the same location at the same time.
-If every output element is written exactly once, atomics are usually unnecessary and slower than a plain store.
+Atomics have performance disadvantages, so they should be used only when necessary.
+Typical examples are histograms, counters, sparse assembly, work lists, and some reductions.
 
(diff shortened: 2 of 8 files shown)


docs/source/sphinx_helper/math_function_families.py
```diff
@@ -0,0 +1,64 @@
...
+            grouped_functions["Unary Functions"].append(function_name)
+        elif param_count == 2:
+            grouped_functions["Binary Functions"].append(function_name)
+        else:
+            grouped_functions["Other Functions"].append(function_name)
+
+    return grouped_functions
+
+
+def generate_math_function_families(app) -> None:
```
SimeonEhrig, line 45: Do we really need our own xml parser? Can this be done by [breath](https://www.breathe-doc.org/).

docs/source/tutorial/atomics.rst
```diff
@@ -0,0 +1,90 @@
...
+
+*alpaka* provides operations such as ``atomicAdd``, ``atomicMin``, ``atomicMax``, ``atomicExch``, and ``atomicCas``.
+For a first encounter, ``atomicAdd`` with the default device scope is the easiest one to reason about.
+
+Common Mistakes
+---------------
+
+- Using atomics for data that is actually written exactly once.
+- Choosing device scope when block-local scope would be enough.
+- Mixing atomics and non-atomic accesses to the same location without a clear protocol. This leads to data races.
```
SimeonEhrig: What does protocol means?

docs/source/tutorial/atomics.rst
```diff
@@ -0,0 +1,90 @@
...
+If you are accumulating into shared memory or another block-private structure, block scope expresses the real requirement more precisely.
+If all blocks may update the same global counter, histogram bin, or reduction output, you need device scope.
+
+*alpaka* provides operations such as ``atomicAdd``, ``atomicMin``, ``atomicMax``, ``atomicExch``, and ``atomicCas``.
+For a first encounter, ``atomicAdd`` with the default device scope is the easiest one to reason about.
+
+Common Mistakes
+---------------
+
+- Using atomics for data that is actually written exactly once.
```
SimeonEhrig: I'm not sure, if this is a common mistake. Sounds really uncommon.

docs/source/tutorial/atomics.rst
```diff
@@ -0,0 +1,90 @@
...
+
+- ``scope::block`` means the atomic operation only needs to be coherent for threads in the same block.
+- ``scope::device`` means the atomic operation must work across the whole device.
+
+Many algorithms only need block-local coordination.
+Even on device global memory using the block scope can be enough if the algorithm guarantees that only threads of the same block can update the same data chunk.
+If you are accumulating into shared memory or another block-private structure, block scope expresses the real requirement more precisely.
+If all blocks may update the same global counter, histogram bin, or reduction output, you need device scope.
+
+*alpaka* provides operations such as ``atomicAdd``, ``atomicMin``, ``atomicMax``, ``atomicExch``, and ``atomicCas``.
```
SimeonEhrig, line 60:
```suggestion
*alpaka* provides operations such as ``atomicAdd``, ``atomicMin``, ``atomicMax``, ``atomicExch``, and ``atomicCas``.
You can find a complete list of all available atomic operations in the `Doxygen documentation <https://alpaka3.readthedocs.io/en/latest/doxygen/namespacealpaka_1_1onAcc.html>`_.
```

docs/source/tutorial/math.rst
```diff
@@ -0,0 +1,50 @@
+Math Functions
+==============
+
+Inside kernels, prefer ``alpaka::math`` over calling backend-specific math APIs or C++ `std` math functions directly.
+That keeps the code portable across the API's ``host``, ``cuda``, ``hip``, and ``oneApi``.
```
SimeonEhrig:
```suggestion
*alpaka* provides a wide range of mathematical functions in the ``alpaka::math`` namespace.
These enable a portable codebase while delivering the best possible performance for the data types used. 
To this end, *alpaka* uses the native math functions of the :ref:`api` (e.g., CUDA built-in math functions) with which a kernel is executed, whenever available.
```

Don't name the user stupid ideas ;-)

docs/source/tutorial/atomics.rst
```diff
@@ -0,0 +1,90 @@
...
+=======
+
+Atomics are the tool you reach for when several workers may update the same memory location.
+Typical examples are histograms, counters, sparse assembly, work lists, and some reductions.
+
+When to Use Atomics
+-------------------
+
+Use an atomic operation only when two or more workers can hit the same location at the same time.
+If every output element is written exactly once, atomics are usually unnecessary and slower than a plain store.
```
SimeonEhrig:
```suggestion
Atomics
=======

Atomics are the tool you reach for when several workers may update the same memory location.
Use an atomic operation only when two or more workers can hit the same location at the same time.
Atomics have performance disadvantages, so they should be used only when necessary.
Typical examples are histograms, counters, sparse assembly, work lists, and some reductions.
```

docs/source/tutorial/atomics.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Histogram
+---------
+
+Histograms are a typical example because many input elements may contribute to the same bin.
```
SimeonEhrig:
```suggestion
`Histograms <https://en.wikipedia.org/wiki/Histogram>`_ are a typical example because many input elements may contribute to the same bin.
```

docs/source/sphinx_helper/math_function_families.py
psychocoderHPC: As written above, I was only able to pull the breath-created documentation into the document and this looks very boring. I am not aware of a way to pull the functions only and link them to the Doxygen documentation.

Thats why I kept the xml parsing.

docs/source/tutorial/atomics.rst
psychocoderHPC: A protocol means that you need to guard against data races. You could write your own handshake protocol  e.g. by using fences instead of using atomics, but if this is possible depends on your use case.

docs/source/tutorial/atomics.rst
psychocoderHPC: I know a time where NVIDIA said you can always use atomics because they are prefecthed and therefore fast. Sometimes user also tent to use atomics because they have not thought about there real access pattern and using atomics.
But if you prefare I can remove it.

docs/source/tutorial/atomics.rst
SimeonEhrig: I would not use the term protocol, because it is uncommon in our domain. If you talk about a protocol, I think about something like `HTTP`.

My suggestion:

Mixing atomics and non-atomic accesses to the same location without a clear code structure. This leads to data races. Try to reuse common and well-known software pattern and idioms.

docs/source/tutorial/atomics.rst
SimeonEhrig: I would remove it. Maybe there was a time but I never heard about this ;-)

docs/source/tutorial/math.rst
psychocoderHPC: Took yours but added `*alpaka* math functions can also be used in host code outside of kernels.`

psychocoderHPC:
I updated the PR

SimeonEhrig (approved):

PR #543, opened 2026-05-24
refactor `onAcc::SimdAlgo::transformReduce()`

psychocoderHPC:
Unify the `transformReduce()` functionality from `onHost` and the in-kernel equivalent.
This is a **breaking change** but is making typical code simpler, because the user does not need to deal with `SimdPtr`.
`SimpPtr` is interesting for stencil operations but often not required for typical transform operations.

- provide the usage of `StencilFunc` and `ScalarFunc` functionality from `onHost::transformReduce()` to the in-kernel variant.
- add test for `onAcc` interface
- update HACCmk to take new interface into account
- remove the overuse of `std::forward()`, it should not be used when a universal reference is used multiple times, e.g. in a loop

### DIFF
--- a/include/alpaka/onAcc/SimdAlgo.hpp
+++ b/include/alpaka/onAcc/SimdAlgo.hpp
@@ -165,9 +165,16 @@ namespace alpaka::onAcc
          *            All other MdSpan objects must have at least the same number of elements.
          *
          * @param neutralElement the neutral element for the reduction operation
-         * @param reduceFunc the binary reduction operation to be executed, e.g. std::plus
-         * @param transformFunc n-nary functor to be executed, values of all containers will be passed to the functor
-         * as arguments. If the result of this functor is a structured value providing an overload to simdize the type
+         * @param reduceFunc The binary reduction operation to be executed, e.g. std::plus. The functor should support
+         * Simd packages.
+         * @param transformFunc N-nary functor to be executed, values of all containers will be passed to the functor
+         * as arguments. The functor should support Simd packages. If not you can enforce the element wise execution by
+         * wrapping into
+         * @see ScalarFunc. If you would like to support stencil executions wrapp fn into @see StencilFunc. StencilFunc
+         * is getting all arguments as @see SimdPtr. If StencilFunc is used you should take care to not read outside of
+         * valid memory ranges by using sub-views to your input and output data. Optionally a transformFn can have an
+         * accelerator as first argument.
+         * If the result of this functor is a structured value providing an overload to simdize the type
          * can improve the performance see alpaka::makeSimdized.
          * @param data0 the first data to be processed
          * @param dataN the remaining data to be processed
--- a/include/alpaka/onAcc/SimdAlgo.hpp
+++ b/include/alpaka/onAcc/SimdAlgo.hpp
@@ -170,8 +170,8 @@ namespace alpaka::onAcc
          * @param transformFunc N-nary functor to be executed, values of all containers will be passed to the functor
          * as arguments. The functor should support Simd packages. If not you can enforce the element wise execution by
          * wrapping into
-         * @see ScalarFunc. If you would like to support stencil executions wrapp fn into @see StencilFunc. StencilFunc
-         * is getting all arguments as @see SimdPtr. If StencilFunc is used you should take care to not read outside of
+         * ScalarFunc. If you would like to support stencil executions wrapp fn into StencilFunc. StencilFunc
+         * is getting all arguments as SimdPtr. If StencilFunc is used you should take care to not read outside of
          * valid memory ranges by using sub-views to your input and output data. Optionally a transformFn can have an
          * accelerator as first argument.
          * If the result of this functor is a structured value providing an overload to simdize the type
--- a/include/alpaka/onAcc/internal/SimdTransformReduce.hpp
+++ b/include/alpaka/onAcc/internal/SimdTransformReduce.hpp
@@ -42,6 +42,7 @@ namespace alpaka::onAcc::internal
         {
             auto numElements = typename ALPAKA_TYPEOF(extents)::UniVec{extents};
             using ValueType = alpaka::trait::GetValueType_t<ALPAKA_TYPEOF(data0)>;
+            decltype(auto) transformFunc = wrapTransformFunc(ALPAKA_FORWARD(func));
 
             constexpr auto simdCfg = T_Parent::template calcSimdPackConfig<ValueType>(
                 ALPAKA_TYPEOF(acc.getApi()){},
@@ -58,7 +59,7 @@ namespace alpaka::onAcc::internal
                     numElements,
                     neutralElement,
                     ALPAKA_FORWARD(reduceFunc),
-                    ALPAKA_FORWARD(func),
+                    transformFunc,
                     ALPAKA_FORWARD(data0),
                     ALPAKA_FORWARD(dataN)...);
             }
@@ -79,8 +80,9 @@ namespace alpaka::onAcc::internal
             {
                 simdizedReducedValue = reduceFunc(
                     simdizedReducedValue,
-                    func(
+                    callFunctor(
                         acc,
+                        transformFunc,
                         SimdPtr{data0, idx, T_MemAlignment{}, CVec<uint32_t, 1u>{}},
                         SimdPtr{dataN, idx, T_MemAlignment{}, CVec<uint32_t, 1u>{}}...));
             }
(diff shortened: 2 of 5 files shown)


include/alpaka/onAcc/SimdAlgo.hpp
```diff
@@ -165,9 +165,16 @@ namespace alpaka::onAcc
...
-         * @param reduceFunc the binary reduction operation to be executed, e.g. std::plus
-         * @param transformFunc n-nary functor to be executed, values of all containers will be passed to the functor
-         * as arguments. If the result of this functor is a structured value providing an overload to simdize the type
+         * @param reduceFunc The binary reduction operation to be executed, e.g. std::plus. The functor should support
+         * Simd packages.
+         * @param transformFunc N-nary functor to be executed, values of all containers will be passed to the functor
+         * as arguments. The functor should support Simd packages. If not you can enforce the element wise execution by
+         * wrapping into
+         * @see ScalarFunc. If you would like to support stencil executions wrapp fn into @see StencilFunc. StencilFunc
+         * is getting all arguments as @see SimdPtr. If StencilFunc is used you should take care to not read outside of
```
SimeonEhrig:
```suggestion
         * ScalarFunc. If you would like to support stencil executions wrapp fn into StencilFunc. StencilFunc
         * is getting all arguments as SimdPtr. If StencilFunc is used you should take care to not read outside of
```

You use `@see` wrong. It does not generate a link. It creates an extra section. Doxygen should automatically link the Types.

Image

include/alpaka/onAcc/internal/SimdTransformReduce.hpp
```diff
@@ -94,14 +96,60 @@ namespace alpaka::onAcc::internal
...
+            auto&&... data)
+        {
+            return Simd{op(CVec<uint32_t, T_idx>{}, acc, ALPAKA_FORWARD(func), ALPAKA_FORWARD(data)...)...};
+        }
+
+        ALPAKA_FN_INLINE ALPAKA_FN_ACC static constexpr auto wrapTransformFunc(auto&& transformFunc)
+        {
+            if constexpr(isSpecializationOf_v<ALPAKA_TYPEOF(transformFunc), StencilFunc>)
+            {
+                return ALPAKA_FORWARD(transformFunc);
```
SimeonEhrig, line 114: Are stencil functors really working? If I'm not wrong a functor which is wrapped in a `StencilFunc` and non-wrapped kernel get the same argument types.

In the test I found the following kernels which does also no stencil operations, because they simply access the central element and no neighbor.

https://github.com/psychocoderHPC/alpaka3/blob/8f0ce8f9889bf580bd492edb9ba67e3fb7f09c5e/test/unit/algorithm/transformReduce.cpp#L25-L42

Can you check, if stencil is working?

include/alpaka/onAcc/SimdAlgo.hpp
psychocoderHPC: ok I will remove it

include/alpaka/onAcc/internal/SimdTransformReduce.hpp
psychocoderHPC: Yes should work, I have not tested it with an example because we do not have one currently. But the code you linked is a stencil that accesses the current position only. 
A Stencil functor is getting `SimdPtr` as arguments, so you can call the `operator[]()` on the SimdPtr to build your stenicl and non wrapped functors takeing the last else branch in this conditional selection and will get a `Simd` object, so `.load()` is already executed for each `SimdPtr`

psychocoderHPC: `SimdPtr` -> you can perform stencil loads.
`Simd` -> is a loaded SIMD pack, you can not change the adress from which the memory is loaded.

SimeonEhrig: I tried first the following kernel. It does not work.

```c++
struct StencilAdd
{
    constexpr auto operator()(concepts::SimdPtr auto const& a, concepts::SimdPtr auto const& b) const
    {
        auto const aIdx = a.getIdx();
        auto const bIdx = b.getIdx();
        return a[aIdx] + b[bIdx];
    }
};
```

This also does not work.

```c++
struct StencilAdd
{
    constexpr auto operator()(concepts::SimdPtr auto const& a, concepts::SimdPtr auto const& b) const
    {
        auto const aIdx = a.getIdx();
        auto const bIdx = b.getIdx();
        return a[aIdx].load() + b[bIdx].load();
    }
};
```

SimeonEhrig: Actual I would do something like `a[aIdx + Vec{-1, 0}] + b[bIdx + Vec{-1, 0}]` to access the halo. But already the code above does not work.

psychocoderHPC: I assume all is compiling and it fails at runtime. The reason is that you perform an out of memory access. You can not us ethe value from `getIdx()` for your stencil. `getIdx()` is providing you with the absolut offset in elements from the begin of the buffer used to build the `SimdPtr`. Stencils are described relative to the current position of the data element you are processing.

Assuming 1D memory the stencil access will be `a[Vec{-1}].load()`.

In algorithms, you already get an element from your buffer, so you do not need to shift the data to the corresponding point in the data space. If you do not use `operator[]()` you and only call `.load()` you will get the element at the point where the pointer is currently located.

psychocoderHPC: > Actual I would do something like `a[aIdx + Vec{-1, 0}] + b[bIdx + Vec{-1, 0}]` to access the halo. But already the code above does not work.

`a[Vec{-1, 0}].load() + b[Vec{-1, 0}].load()`  is the correct syntax. `operator[]()` is returning a `SimdPtr`. The dereferencing operation is `.load()`.

attention: `a[Vec{-1, 0}].load() + b[Vec{-1, 0}].load()`  assumes that you work on data where you can savely access the elements above and below, so you need to use a `SubView` for you algorithm to avoid processing the first row and the last row of your 2D data.

SimeonEhrig: Nope, compile time error:

```
error: no viable overloaded operator[] for type 'const alpaka::SimdPtr<alpaka::View<alpaka::api::Host, int, alpaka::Vec<int, 1>, alpaka::Alignment<16>>, alpaka::Vec<int, 1>, alpaka::Alignment<>, alpaka::Vec<unsigned int, 1, alpaka::detail::CVec<unsigned int, 4>>>'
   40 |         a[Vec{-1, 0}].load() + b[Vec{-1, 0}].load();
      |         ^ ~~~~~~~~~~
```

SimeonEhrig: I think the problem is, that the tests uses different dimensions for the input.

SimeonEhrig: I think this should compile for a 2D extents but it doesn't compile:

```c++
struct StencilAddWithAcc
{
    constexpr auto operator()(
        onAcc::concepts::Acc auto const&,
        concepts::SimdPtr auto const& a,
        concepts::SimdPtr auto const& b) const
    {
        using IdxType = typename ALPAKA_TYPEOF(a.getIdx());
        a[Vec{IdxType{0}, IdxType{0}}];
        return a.load() + b.load();
    }
};
```

psychocoderHPC: If you simply used the test you linked above you can not use hard coded 1D or 2D because the test is scanning over mutiple dimension and is executing 1d,2D,3D and 4D.

This is also not working because the second line is doing nothing. What should work is this, which selects the diagonal element in positive direction. But this only works if you start the algorith mwith a subview.


```C++
    {
        using IdxType = typename ALPAKA_TYPEOF(a.getIdx());
        return a[Vec{IdxType::fill(1)].load() + b.load();
    }
```

psychocoderHPC: I opened a PR to fix the stencil/shift operator of `SimdPtr` we broke it while introducing a concept.
A test case will be shipped in a seperate PR.

https://github.com/alpaka-group/alpaka3/pull/550

psychocoderHPC: the fix #550 is merged. A test for the stencil will be a different PR.

SimeonEhrig:
@psychocoderHPC I would merge the PR because it is a refactoring but we need a test/example which verifies the stencil functionality soon. It the moment it simply fulfills all conditions for a dead and useless feature. It is not tested, not used somewhere and not documented.

Are fine, that we solve the this issue soon? Otherwise I would remove the feature and maybe bring back later.

psychocoderHPC:
> @psychocoderHPC I would merge the PR because it is a refactoring but we need a test/example which verifies the stencil functionality soon. 

Yes please merge it as it is. The test case is on my todo list.

psychocoderHPC:
> It the moment it simply fulfills all conditions for a dead and useless feature. It is not tested, not used somewhere and not documented.

That's a little bit harsh, in that case we would need to delete alpaka mainline and also parts of alpaka3.

SimeonEhrig:
> > It the moment it simply fulfills all conditions for a dead and useless feature. It is not tested, not used somewhere and not documented.
> 
> That's a little bit harsh, in that case we would need to delete alpaka mainline and also parts of alpaka3.

Not whole alpaka mainline. It has a good test coverage and sometimes the doc is not nice but okay.

But in the end, that's what good software project does. Remove features where nobody knows how to use and/or if they are still working. Even it sounds usable.

PR #547, opened 2026-05-26
CI: implement retry_cmd

SimeonEhrig:
- retries to run a command in case of an error
- maybe it solves the llvm apt update problem in the GitHub Actions CI

### DIFF
--- a/script/ci/utils/misc.sh
+++ b/script/ci/utils/misc.sh
@@ -8,7 +8,7 @@
 : "${APCI_ALPAKA_ROOT?'APCI_ALPAKA_ROOT is not defined. Root directory of the alpaka project'}"
 
 # shellcheck source=script/ci/utils/color_echo.sh
-source "${APCI_ALPAKA_ROOT}/script/ci/utils//color_echo.sh"
+source "${APCI_ALPAKA_ROOT}/script/ci/utils/color_echo.sh"
 
 # display a install message in green
 install_msg() {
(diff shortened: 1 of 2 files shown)


script/ci/utils/misc.sh
```diff
@@ -53,11 +53,49 @@ parse_compiler_version() {
...
+retry_cmd() {
+    if [[ $# -lt 1 ]]; then
+        exit_error "retry_cmd requires at least one argument."
+    fi
+    (
+        set +euo pipefail
+        local max_tries=10
+        if [[ -n ${RETRY_CMD_MAX+x} ]]; then
+            max_tries="$RETRY_CMD_MAX"
+        fi
```
psychocoderHPC: can be written more imsple

```suggestion
        local max_tries="${RETRY_CMD_MAX:-10}"
```

script/ci/utils/misc.sh
```diff
@@ -53,11 +53,49 @@ parse_compiler_version() {
...
+
+        # time in seconds
+        local wait_time=1
+        if [[ -n ${RETRY_CMD_WAIT+x} ]]; then
+            wait_time="$RETRY_CMD_WAIT"
+        fi
```
psychocoderHPC:
```suggestion
        local wait_time="${RETRY_CMD_WAIT:-1}"
```

psychocoderHPC (changes requested):

SimeonEhrig:
It looks like the problem is solved but different than expected. I would assume the `wget https://apt.llvm.org/llvm-snapshot.gpg.key` has still network problems and the `retry_cmd` is triggered. But this is not the case. The `wget` works each time at the first try. Maybe splitting up fetching the key and move to the right location solved the problem.

SimeonEhrig:
@psychocoderHPC Please review again.

psychocoderHPC (approved):

PR #548, opened 2026-05-26
optimze compile time of alpaka

psychocoderHPC:
Optimize includes to reduce the compile time by around 12%. The biggest improve came from removing `#include <regex>` by manual parsing.
Change the implementation of `simplifyFunctionSignature()`, due to the manuel parsing some issues with the shortening are fixed and the compile time improved a lot.

### DIFF
--- a/include/alpaka/onHost/demangledName.hpp
+++ b/include/alpaka/onHost/demangledName.hpp
@@ -6,7 +6,6 @@
 
 #include "alpaka/core/config.hpp"
 
-#include <regex>
 #include <source_location>
 #include <string>
 #include <string_view>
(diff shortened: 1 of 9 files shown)


include/alpaka/onHost/demangledName.hpp
```diff
@@ -64,18 +63,46 @@ namespace alpaka::onHost
      */
     inline std::string simplifyFunctionSignature(std::string const& deName)
```
SimeonEhrig, line 64: Did you try to make the function `constexpr`?

SimeonEhrig (approved):

include/alpaka/onHost/demangledName.hpp
psychocoderHPC: I have not tried it.

psychocoderHPC: I checked if `constexpr` helps: The usage of constexpr is not possible in places where it is used, so marking it constexpr does not help.
Maybe it would be possible if there were bigger changes on other places.

SimeonEhrig: That's fine, because logging was not performance critical at the moment. Was only a check for low hanging fruit.

PR #551, opened 2026-05-28
remove `onAcc::Atomic*` functors

psychocoderHPC:
The functors `onAcc::Atomic*` was taken over from alpaka mainline, the issue with them is that they are not atomic and the name is missleading. In alpaka mainline users tried to use them to perfrom atomic operations.

This refactoring is moving them into `alpaka::operator` to marke tha they are usable on `onHost` and `onAcc`.
A documentation string of the namespace adn by removing `Atomic` from the name the missinterpretation should not happen again.

### DIFF
--- a/docs/snippets/cheatsheet/cheatsheet.cpp
+++ b/docs/snippets/cheatsheet/cheatsheet.cpp
@@ -108,10 +108,10 @@ struct KernelWait
 
         int* ptr = nullptr;
         // BEGIN-CHEATSHEET-atomicAdd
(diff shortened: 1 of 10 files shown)


SimeonEhrig:
I think we should first merge PR #542 that the documentation can be updated.

SimeonEhrig:
> I think we should first merge PR #542 that the documentation can be updated.

We discussed offline. This PR is independent of PR #542. This functionality has nothing to do with atomics. It's a different mechanism. It only shares the same function interface definition. Therefore the misleading name.

SimeonEhrig (approved):

PR #553, opened 2026-06-01
Stencil transform example Game of Life

SimeonEhrig:
I want to implement [Game of Life](https://en.wikipedia.org/wiki/Conway%27s_Game_of_Life) as example using `alpaka::onHost::transform` and `alpaka::StencilFunc`.

During the review of PR #543 I found the stencil feature. There are no tests or examples who uses the feature and there is also no documentation. During viewing the code, I got strong concerns, that the feature does not work as expected or maybe missing some features which are required to implement real world problems.

# TODO

- [ ] based on PR #543. Rebase after the PR is merged.


SimeonEhrig:
@psychocoderHPC I have problem. I'm not sure how to use a halo. The halo is not automatically applied. Maybe I need to use a subview. How did you planned to configure a halo for the `alpaka::StencilFunction`?

psychocoderHPC:
> @psychocoderHPC I have problem. I'm not sure how to use a halo. The halo is not automatically applied. Maybe I need to use a subview. How did you planned to configure a halo for the `alpaka::StencilFunction`?

Typically you do not perform the stencil on the helo. So you run it only on the inner part of the volume. For the helo you run a seperate kernel like in the heat equation and you copy the data from the left side to the right and the other way around. Same with diagonals.

not sure if it helps but here is an example how it is implemented with PMacc https://github.com/ComputationalRadiationPhysics/picongpu/blob/39e00f9745cd3eb46b11a0cd6d69c8d95b974936/share/pmacc/examples/gameOfLife2D/include/Evolution.hpp#L92-L106

example/stencilTransform/src/stencilTransform.cpp
```diff
@@ -0,0 +1,72 @@
...
+    constexpr std::size_t site_size = 10;
+    // Add 2 to each site for the halo
+    constexpr Vec<std::size_t, 2> extents{site_size + 2, site_size + 2};
+
+    auto deviceSelector = onHost::makeDeviceSelector(deviceSpec);
+    onHost::Device device = deviceSelector.makeDevice(0);
+    onHost::Queue queue = device.makeQueue();
+
+    auto bufHost = onHost::allocHost<int>(extents);
+    auto bufDevice = onHost::alloc<int>(device, extents);
```
psychocoderHPC, line 33: I suggest using the coreView to avoid out-of-memory reads. Thats very typical for stencil codes

```C++
auto coreView = bufDevice.getSubView(Vec<std::size_t, 2>::fill(1u), extents  - Vec<std::size_t, 2>::fill(1u) * 2u);
```

example/stencilTransform/src/stencilTransform.cpp
```diff
@@ -0,0 +1,72 @@
...
+    // We use a square world. Each site has the size.
+    constexpr std::size_t site_size = 10;
+    // Add 2 to each site for the halo
+    constexpr Vec<std::size_t, 2> extents{site_size + 2, site_size + 2};
+
+    auto deviceSelector = onHost::makeDeviceSelector(deviceSpec);
+    onHost::Device device = deviceSelector.makeDevice(0);
+    onHost::Queue queue = device.makeQueue();
+
+    auto bufHost = onHost::allocHost<int>(extents);
```
psychocoderHPC, line 32: You do not need the helo on the host side
```C++
auto bufHost = onHost::allocHost<int>(extents - 2u);
```

example/stencilTransform/src/stencilTransform.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto bufDevice = onHost::alloc<int>(device, extents);
+    onHost::memset(queue, bufDevice, 0);
+
+    onHost::transform(queue, exec, bufDevice, StencilFunc{GameOfLife{}}, bufDevice);
+    onHost::memcpy(queue, bufHost, bufDevice);
+    onHost::wait(queue);
+
+    std::cout << "Only user data: \n";
+    for(std::size_t y = 1; y < site_size + 1; ++y)
```
psychocoderHPC, line 41: If the host buffer does not contain the helo you can simply run 

```C++
onHost::memcpy(queue, bufHost, coreView);
```

psychocoderHPC (changes requested):

example/stencilTransform/src/stencilTransform.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    auto deviceSelector = onHost::makeDeviceSelector(deviceSpec);
+    onHost::Device device = deviceSelector.makeDevice(0);
+    onHost::Queue queue = device.makeQueue();
+
+    auto bufHost = onHost::allocHost<int>(extents);
+    auto bufDevice = onHost::alloc<int>(device, extents);
+    onHost::memset(queue, bufDevice, 0);
+
+    onHost::transform(queue, exec, bufDevice, StencilFunc{GameOfLife{}}, bufDevice);
```
psychocoderHPC, line 36: Note, for a game of life you need a double buffer and a loop around the computation because it is a step-based cellular state machine.

```
W = world at step
S = stencil at any point

for(n=0; n < maxN; ++n)
   Wn+1 = S(Wn))
```

example/stencilTransform/src/stencilTransform.cpp
```diff
@@ -0,0 +1,72 @@
...
+    onHost::memcpy(queue, bufHost, bufDevice);
+    onHost::wait(queue);
+
+    std::cout << "Only user data: \n";
+    for(std::size_t y = 1; y < site_size + 1; ++y)
+    {
+        for(std::size_t x = 1; x < site_size + 1; ++x)
+        {
+            std::cout << bufHost[Vec{y, x}] << " ";
```
psychocoderHPC, line 45: I suggest to use the PNGWriter equal to the heat equation, it is easy to use and better for verification.
What I do not know is how to verify via algorithm if  the game of life iterations did the correct calculation. 
I always used the images to verify the correctness but thats bad in the CI.
Maybe we need to start with a regular pattern instead of random numbers and store a reference csv file after N steps.

psychocoderHPC:
Note: If you use `onHost::transform`, there is no need to depend on #543, because the PR adds the ability to define the type of the transform functor to the `onAcc::SimdAlgo::transformReduce` class. Before #543 `onAcc::SimdAlgo` always used `SimdPtr` aka `StencilFunc` for the transform reduce algorithm.

psychocoderHPC:
@SimeonEhrig Could you please name the example `gameOfLife`, we name our examples based on what the example is not based on the technique used

SimeonEhrig:
Closed because it is to time consuming.

PR #560, opened 2026-06-03
add event tutorial

psychocoderHPC:
Show basic usage of events.


SimeonEhrig:
@psychocoderHPC  Pull request with the feedback: https://github.com/psychocoderHPC/alpaka3/pull/37

I did two major points:
- Write some examples where multi queues are used to get an idea why this is important
- Summarize the code examples in one code block. In my opinion the code blocks didn't work standalone. A code block should contain all information which are important to understand the feature. In this case, you need to remember what happen in the code block before to understand the current block. For example, non of the code blocks contains both queues but this is the major idea to synchronize both queues.

SimeonEhrig (approved):

PR #564, opened 2026-06-03
fix clion ignore

psychocoderHPC:
I excluded the wrong not existing folder.


SimeonEhrig (approved):

SimeonEhrig:
Thanks for the ignore list

PR #565, opened 2026-06-03
refactor `onHost::getFrameSpec()`

psychocoderHPC:
- Introduce a new implementation of `getFrameSpec()` which is not performing SIMD optimizations and therfore does not need a data type.
- The old SIMD optimized function is now called `getSimdFrameSpec<DataType>()`
- Update the current places where the functions were used.

**breaking change:** This is a breaking change because places where `getFrameSpec<Type>()` was sued must be changed to `getFrameSpec()`. Only if you really need a SIMD optimized frame specification, you should use `getSimdFrameSpec<DataType>()`.

### DIFF
--- a/docs/snippets/cheatsheet/cheatsheet.cpp
+++ b/docs/snippets/cheatsheet/cheatsheet.cpp
@@ -379,12 +379,14 @@ auto main() -> int
 
         {
             // BEGIN-CHEATSHEET-autoFrameSpec
+            // Provides kernel start parameters sutable for the device and executor
+            onHost::concepts::FrameSpec auto frameSpec = onHost::getFrameSpec(device, exec::anyExecutor, extentMd);
             // DataType is used to optimize the kernel parameters for working on data of this type
-            onHost::concepts::FrameSpec auto frameSpec
-                = onHost::getFrameSpec<DataType>(device, exec::anyExecutor, extentMd);
+            onHost::concepts::FrameSpec auto simdFrameSpec
+                = onHost::getSimdFrameSpec<DataType>(device, exec::anyExecutor, extentMd);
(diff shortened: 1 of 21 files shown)


SimeonEhrig:
I assume `onHost::getSimdFrameSpec()` is indirectly tested via algorithm. Can we write a small test case, to verify that the calculation is correct and the alignment works?

psychocoderHPC:
> I assume `onHost::getSimdFrameSpec()` is indirectly tested via algorithm. Can we write a small test case, to verify that the calculation is correct and the alignment works?

Yes, it is tested with the algorithms. I do not know how we can test it directly, the reason is that the results differ on each architecture and based on compiler flags. We would need to mock interfaces to make this unit testable.

note: It has nothing to do with alignment; it is only creating a frame specification that supports SIMD execution on the given device and data type if `onAcc::worker::threadsPerGrid` is in the outer loop used.  If data is aligned is not affected by this function.

If something is wrong, it is mostly only visible in benchmark results.

SimeonEhrig:
> > I assume `onHost::getSimdFrameSpec()` is indirectly tested via algorithm. Can we write a small test case, to verify that the calculation is correct and the alignment works?
> 
> Yes, it is tested with the algorithms. I do not know how we can test it directly, the reason is that the results differ on each architecture and based on compiler flags. We would need to mock interfaces to make this unit testable.
> 
> note: It has nothing to do with alignment; it is only creating a frame specification that supports SIMD execution on the given device and data type if `onAcc::worker::threadsPerGrid` is in the outer loop used. If data is aligned is not affected by this function.
> 
> If something is wrong, it is mostly only visible in benchmark results.

I'm not completely fine with the answer but we can merge.

SimeonEhrig (approved):

PR #568, opened 2026-06-04
fix incorrect pitches documentation

ikbuibui:
Pitches are returned in bytes by alpaka.

### DIFF
--- a/include/alpaka/onHost/interface.hpp
+++ b/include/alpaka/onHost/interface.hpp
@@ -45,7 +45,7 @@ namespace alpaka::onHost
     /** Object pitches
      *
      * @param any can be a std::vector, std::array, ...
-     * @return the number of elements of the object
+     * @return the pitches in bytes (number of elements of the object * sizeof(object) + padding)
      */
     inline decltype(auto) getPitches(auto&& any)
     {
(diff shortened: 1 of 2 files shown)


include/alpaka/onHost/interface.hpp
```diff
@@ -45,7 +45,7 @@ namespace alpaka::onHost
     /** Object pitches
      *
      * @param any can be a std::vector, std::array, ...
-     * @return the number of elements of the object
+     * @return the pitches in bytes (number of elements of the object * sizeof(object) + padding)
```
psychocoderHPC: Could you change it to

```
@return Multidimensional value with number of bytes to jump to the next value within the corresponding dimension.
```

The reason is that it is in x dimension of sizeof(type), in y direction it is `number of elements in x * sizeof(type) + padding`, ...

With the current text I would assume it is the line row length in bytes only but it is multidimensional.
Not sure if we could add into the detailed text of doxygen. That if we multiply an ND index with the piches and take the sum (aka dor product of index and pitches) we get the byte address where the data point starts. But I do not know how to phrase it.

psychocoderHPC (changes requested):

include/alpaka/onHost/interface.hpp
ikbuibui: good point. I used your suggestion and rephrased it

psychocoderHPC (approved):

PR #569, opened 2026-06-05
add stencil transform test

psychocoderHPC:
In #543 the reviewer would like to have a test for the stencil feature usable within transform. To avoid breaking the feature see #550.

The test is using a simple asymetric cross stencil and test it for up to 4 dimensions.

The second commit is adding a workaround for nvcc. Nvcc segfaults during compile wenn calling `SimdPtr::operator[]()` due to the concept in the function signature.

### DIFF
--- a/test/unit/algorithm/transform.cpp
+++ b/test/unit/algorithm/transform.cpp
@@ -293,3 +293,136 @@ TEMPLATE_LIST_TEST_CASE("transform generator", "", TestBackends)
 
     std::apply([&](auto... extents) { (prepareTest<DataType>(cfg, extents, functorList), ...); }, extentMdList);
 }
+
+/** Cross stencil
+ *
+ * N-dimensional asymmetric cross stencil. The positive direction is weighted with 5 and the negative with 3.
+ * The middle value is not weighted.
+ */
+struct CrossStencil
+{
+    constexpr auto operator()(concepts::SimdPtr auto const& in) const
+    {
+        using SimdPtrType = ALPAKA_TYPEOF(in);
+        using VecIdxType = typename SimdPtrType::IdxType;
+
+        // Load the value the simd pointer is pointing to.
+        auto result = in.load();
+
+        for(uint32_t d = 0u; d < VecIdxType::dim(); ++d)
+        {
+            /* Shift relative to the current element pointed to into the negative or positive direction within a
+             * dimension.
+             */
+            concepts::Vector auto negative = VecIdxType::fill(0);
+            concepts::Vector auto positive = VecIdxType::fill(0);
+            negative[d] = -1;
+            positive[d] = 1;
+
+            result += in[negative].load() * 3;
+            result += in[positive].load() * 5;
+        }
+
+        return result;
+    }
+};
+
+template<typename T_DataType, typename T_StencilFn>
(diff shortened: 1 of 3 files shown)


SimeonEhrig (approved):

SimeonEhrig:
@psychocoderHPC The test is good. Can we extent it to add a stack variable? Something like

```c++
struct CrossStencil
{
    constexpr auto operator()(concepts::SimdPtr auto const& in) const
    {
        using SimdPtrType = ALPAKA_TYPEOF(in);
        using VecIdxType = typename SimdPtrType::IdxType;

        // Load the value the simd pointer is pointing to.
        auto result = in.load();

        using SimdStackType = /* ? */;
        SimdStackType const stackValue = 2;

        for(uint32_t d = 0u; d < VecIdxType::dim(); ++d)
        {
            /* Shift relative to the current element pointed to into the negative or positive direction within a
             * dimension.
             */
            concepts::Vector auto negative = VecIdxType::fill(0);
            concepts::Vector auto positive = VecIdxType::fill(0);
            negative[d] = -1;
            positive[d] = 1;

            result += in[negative].load() * 3 + stackValue;
            result += in[positive].load() * 5 + stackValue;
        }

        return result;
    }
};
```

It is actual motivated by game of life, where you count number of living neighbors and set the cell state depending on the value.

psychocoderHPC:
Yes , if you like I can add it tomorrow.

SimeonEhrig:
> Yes , if you like I can add it tomorrow.

Yes please. I think this is real use case which we can easy test.

psychocoderHPC:
I added the magic constant you like to add to the result.

SimeonEhrig (approved):

test/unit/algorithm/transform.cpp
```diff
@@ -293,3 +293,141 @@ TEMPLATE_LIST_TEST_CASE("transform generator", "", TestBackends)
...
+    {
+        using SimdPtrType = ALPAKA_TYPEOF(in);
+        using VecIdxType = typename SimdPtrType::IdxType;
+
+        // Load the value the simd pointer is pointing to.
+        concepts::Simd auto result = in.load();
+
+        // Arbitrary constant added to the result.
+        using SimdType = ALPAKA_TYPEOF(result);
+        concepts::Simd auto const value = SimdType::fill(42);
```
SimeonEhrig, line 314: It is not require to use it this PR!

```suggestion
        concepts::Simd auto const value = in.load().fill(42);
```

This can be simplified. Maybe an idea for an upcoming PR:

```c++
concepts::Simd auto const value = in.create_value(42);
```

psychocoderHPC: Not sure of this triggers a load and than overwrite the value. The syntax is valid and our Vec class is supporting it too.
I would avoid it because of the possible side effect of the load. 
In principle the reader could also thing that the data of the loaded value is overwritten.

SimeonEhrig: Actual than only `concepts::Simd auto const value = in.create_value(42);` make sense where the implementation is:

```c++
static auto create_value(auto val){
   using SimdType = ALPAKA_TYPEOF(this->load());
   return SimdType::fill(val);
}
```

PR #575, opened 2026-06-11
mini tutorial for bit intrinsics

psychocoderHPC:
Tiny tutorial which is mentioning bit intrinsics.
I had no good ideas for useful small examples; therefore the section is currently very small.
For popcount, something together with warps could be used but I shifted it to something we could improve later.


psychocoderHPC:
pr is now rebased

SimeonEhrig (approved):

PR #578, opened 2026-06-15
test SKIP() in test::getAvailableDevice()

SimeonEhrig:
Refactor testing. The basic idea is, that tests should fail if they does not check anything but tests should not fail if all tests are skipped because of a missing runtime device.

Split up both steps in two PR's to make it easier to review.
- This PR introduces Catch2 `SKIP()` to get easier an alpaka device if available
- PR #581 enables the check, that each test does a assertation

### DIFF
--- a/test/include/alpakaTest/deviceHelper.hpp
+++ b/test/include/alpakaTest/deviceHelper.hpp
@@ -32,7 +32,7 @@ namespace alpaka::test
 
         if(!devSelector.isAvailable())
         {
-            WARN("No device available for " << deviceSpec.getName());
+            SKIP("No device available for " << deviceSpec.getName());
             return {};
         }
 
--- a/test/include/alpakaTest/deviceHelper.hpp
+++ b/test/include/alpakaTest/deviceHelper.hpp
@@ -19,10 +19,24 @@ namespace alpaka::test
      *
      * Prints information about the device Spec, API and device.
      *
+     * @attention This function should only be used with the Catch2 macro `TEMPLATE_LIST_TEST_CASE`.
+     * See following code snippet:
+     *
+     * @code
+     * using TestApis = std::decay_t<decltype(onHost::allBackends(onHost::enabledDeviceSpecs,
+     * exec::enabledExecutors))>;
+     *
+     * TEMPLATE_LIST_TEST_CASE("device analysis", "", TestApis)
+     * {
+     *    onHost::Device device = test::getDeviceOrSkipTest(TestType::makeDict());
+     *    // ...
+     * }
+     * @endcode
+     *
      * @param cfg Test configuration. An entry of the list returned from alpaka::onHost::allBackends().
      * @return The device 0 if available. Otherwise, SKIP() the test.
      */
-    [[nodiscard]] auto getAvailableDevice(auto const& cfg)
+    [[nodiscard]] auto getDeviceOrSkipTest(auto const& cfg)
         -> decltype(onHost::makeDeviceSelector(cfg[object::deviceSpec]).makeDevice(0))
     {
         auto deviceSpec = cfg[object::deviceSpec];
@@ -44,13 +58,29 @@ namespace alpaka::test
      *
      * Prints information about the device Spec, API, device and executor.
      *
+     * @attention This function should only be used with the Catch2 macro `TEMPLATE_LIST_TEST_CASE`.
+     * See following code snippet:
+     *
+     * @code
+     * using TestApis = std::decay_t<decltype(onHost::allBackends(onHost::enabledDeviceSpecs,
+     * exec::enabledExecutors))>;
+     *
+     * TEMPLATE_LIST_TEST_CASE("device analysis", "", TestApis)
+     * {
+     *    auto deviceExec = test::getDeviceExecutorOrSkipTest(TestType::makeDict());
+     *    onHost::Device device = test::getDevice(deviceExec);
+     *    concepts::Executor auto exec = test::getExecutor(deviceExec);
+     *    // ...
+     * }
+     * @endcode
+     *
      * @param cfg Test configuration. An entry of the list returned from alpaka::onHost::allBackends().
      * @return A std::tuple with the device 0 and an executor. If no device is available, SKIP() the test.
      */
-    [[nodiscard]] auto getAvailableDeviceExecutor(auto const& cfg) -> std::
+    [[nodiscard]] auto getDeviceExecutorOrSkipTest(auto const& cfg) -> std::
         tuple<decltype(onHost::makeDeviceSelector(cfg[object::deviceSpec]).makeDevice(0)), decltype(cfg[object::exec])>
     {
-        auto device = alpaka::test::getAvailableDevice(cfg);
+        auto device = alpaka::test::getDeviceOrSkipTest(cfg);
         concepts::Executor auto executor = cfg[object::exec];
         UNSCOPED_INFO("Executor: " << executor.getName());
 
--- a/test/include/alpakaTest/deviceHelper.hpp
+++ b/test/include/alpakaTest/deviceHelper.hpp
@@ -20,10 +20,10 @@ namespace alpaka::test
      * Prints information about the device Spec, API and device.
      *
      * @param cfg Test configuration. An entry of the list returned from alpaka::onHost::allBackends().
-     * @return The device 0 if available. Otherwise, the std::optional is invalid.
+     * @return The device 0 if available. Otherwise, SKIP() the test.
      */
     [[nodiscard]] auto getAvailableDevice(auto const& cfg)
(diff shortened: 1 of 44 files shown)


psychocoderHPC:
IMO using `SKIP()` in the device helper `getAvailableDevice()` should not be the preferred way. 
I added this pseudo-code example. Skip is throwing an exception and is therefore stopping the full test case.
If the first used combination from `allBackends` is failing, all other combinations will silently be ignored and never be executed, even if they are never tested.

Very often, test cases are written that way:

```C++
using DeviceSpecs = std::decay_t<decltype(onHost::getDeviceSpecsFor(onHost::enabledApis))>;

TEMPLATE_LIST_TEST_CASE("example", "", DeviceSpecs)
{
}
```

Where the skip is not a problem. But we have test which using folding expressions within the test.
IMO that's also ok, I would like to avoid enforcing that `getAvailableDevice()` can only be used in a very specific pattern, especially since we are thinking about exposing this helper to external projects too.

Here is an example that shows where a skip can provide hidden skipped tests.

```C++
void testCase(auto cfg)
{
    auto optionalDevice = test::getAvailableDevice(cfg);
    if(!optionalDevice)
        return; 
    
    // run test
}
TEST_CASE("example", "")
{
    auto allBackends = onHost::allBackends(onHost::enabledDeviceSpecs, exec::enabledExecutors);
    
    std::apply([&](auto ... backends)
    {
        ((testCase(backends)), ...);
    },allBackends);
}
```

SimeonEhrig:
> IMO using `SKIP()` in the device helper `getAvailableDevice()` should not be the preferred way. I added this pseudo-code example. Skip is throwing an exception and is therefore stopping the full test case. If the first used combination from `allBackends` is failing, all other combinations will silently be ignored and never be executed, even if they are never tested.
> 
> Very often, test cases are written that way:
> 
> ```c++
> using DeviceSpecs = std::decay_t<decltype(onHost::getDeviceSpecsFor(onHost::enabledApis))>;
> 
> TEMPLATE_LIST_TEST_CASE("example", "", DeviceSpecs)
> {
> }
> ```
> 
> Where the skip is not a problem. But we have test which using folding expressions within the test. IMO that's also ok, I would like to avoid enforcing that `getAvailableDevice()` can only be used in a very specific pattern, especially since we are thinking about exposing this helper to external projects too.
> 
> Here is an example that shows where a skip can provide hidden skipped tests.
> 
> ```c++
> void testCase(auto cfg)
> {
>     auto optionalDevice = test::getAvailableDevice(cfg);
>     if(!optionalDevice)
>         return; 
>     
>     // run test
> }
> TEST_CASE("example", "")
> {
>     auto allBackends = onHost::allBackends(onHost::enabledDeviceSpecs, exec::enabledExecutors);
>     
>     std::apply([&](auto ... backends)
>     {
>         ((testCase(backends)), ...);
>     },allBackends);
> }
> ```

Your assumption is wrong. The macro `TEMPLATE_LIST_TEST_CASE` works different than you assumes. It generates a own function for each entry in the `std::stuple`, which you passed as arguments. For example, I used a CMake configuration, which enables the host CPU, OpenMP and CUDA on a system without CUDA GPU.

```bash
$ ./build/ninja-rel-cuda-nvcc-gcc-cpu/test/unit_test_algorithm_concurrent --list-tests
All available test cases:
  concurrent generator - TestBackends - 1
  concurrent generator - TestBackends - 2
  concurrent generator - TestBackends - 0
  concurrent - TestBackends - 1
  concurrent - TestBackends - 0
  concurrent - TestBackends - 2
6 test cases

$ ./build/ninja-rel-cuda-nvcc-gcc-cpu/test/unit_test_algorithm_concurrent "concurrent - TestBackends - 0" -s | grep -e "DeviceSpec:" -e "API:" -e "Executor:" -e "SKIP"
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuOmpBlocks
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuOmpBlocks
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuOmpBlocks
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuOmpBlocks
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuOmpBlocks
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuOmpBlocks
$ ./build/ninja-rel-cuda-nvcc-gcc-cpu/test/unit_test_algorithm_concurrent "concurrent - TestBackends - 1" -s | grep -e "DeviceSpec:" -e "API:" -e "Executor:" -e "SKIP"
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuSerial
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuSerial
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuSerial
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuSerial
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuSerial
  DeviceSpec: Host Cpu
  API: Host
  Executor: CpuSerial
$ ./build/ninja-rel-cuda-nvcc-gcc-cpu/test/unit_test_algorithm_concurrent "concurrent - TestBackends - 2" -s | grep -e "DeviceSpec:" -e "API:" -e "Executor:" -e "SKIP"
/home/user/projects/alpaka2/test/include/alpakaTest/deviceHelper.hpp:35: SKIPPED:
  DeviceSpec: Cuda NvidiaGpu
  API: Cuda
./build/ninja-rel-cuda-nvcc-gcc-cpu/test/unit_test_algorithm_concurrent "concurrent - TestBackends - 1"
Filters: "concurrent - TestBackends - 1"
Randomness seeded to: 3765705274
===============================================================================
All tests passed (450484 assertions in 1 test case)
$ ./build/ninja-rel-cuda-nvcc-gcc-cpu/test/unit_test_algorithm_concurrent "concurrent - TestBackends - 2"
Filters: "concurrent - TestBackends - 2"
Randomness seeded to: 4197496696

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
unit_test_algorithm_concurrent is a Catch2 v3.15.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
concurrent - TestBackends - 2
-------------------------------------------------------------------------------
/home/user/projects/alpaka2/build/ninja-rel-cuda-nvcc-gcc-cpu/alpaka_build_files/cuda/alpaka/unit/algorithm/concurrent.cpp:134
...............................................................................

/home/user/projects/alpaka2/test/include/alpakaTest/deviceHelper.hpp:35: SKIPPED:
explicitly with messages:
  No device available for Cuda NvidiaGpu

================================================================================
test cases: 1 | 1 skipped
assertions: - none -
```

Therefore you can exit a test function for a specific device via `return` in the top namespace of the function, like we do now or via `SKIP()` which throws an exception, which is cached and also exit the function.

By the way, both approaches requires that we iterate over the devices via `TEMPLATE_LIST_TEST_CASE`. If we would iterate of the devices via `std::apply`, it would not work.

psychocoderHPC:
> By the way, both approaches requires that we iterate over the devices via TEMPLATE_LIST_TEST_CASE. If we would iterate of the devices via std::apply, it would not work.

Yes for `TEMPLATE_LIST_TEST_CASE` it is working fine. Exactly what I wrote `... Where the skip is not a problem. ` but when used with `TEST_CASE` it will fail.
That's surprising for a function with a name `getAvailableDevice()`.

IMO we can simply extend the doxygen of the function and mention that it must be used with `TEMPLATE_LIST_TEST_CASE` or we could rename the function as suggested [here](https://github.com/alpaka-group/alpaka3/pull/582#issuecomment-4729088996) to `test::getDeviceOrSkipTest()` but maybe still warn in the doxygen about the early exit of the full test.

SimeonEhrig:
> > By the way, both approaches requires that we iterate over the devices via TEMPLATE_LIST_TEST_CASE. If we would iterate of the devices via std::apply, it would not work.
> 
> Yes for `TEMPLATE_LIST_TEST_CASE` it is working fine. Exactly what I wrote `... Where the skip is not a problem. ` but when used with `TEST_CASE` it will fail. That's surprising for a function with a name `getAvailableDevice()`.
> 
> IMO we can simply extend the doxygen of the function and mention that it must be used with `TEMPLATE_LIST_TEST_CASE` or we could rename the function as suggested [here](https://github.com/alpaka-group/alpaka3/pull/582#issuecomment-4729088996) to `test::getDeviceOrSkipTest()` but maybe still warn in the doxygen about the early exit of the full test.

done in commit [a3d6ea6](https://github.com/alpaka-group/alpaka3/commit/a3d6ea632a862ac693aa407214f913d2b47d8c8d)

psychocoderHPC (approved):

PR #580, opened 2026-06-16
Critical: fix n-dimensional memcpy

psychocoderHPC:
The memcopy operation with n-dimensional data was broken if a sub-view was used.

Cuda: copies with more than 3 dimensions are broken
OneApi: copies with more than 2 dimensions are broken
Host: all works as expected

- add test to check the sub-view copy method
- fix backends
- use `size_t` for all calculations to avoid index range overflows

**important**:  If you **not** use sub-views but full views or buffers directly, all works as expected. This has already been tested.

`alpaka::onHost::memset()` has the same problem. A fix for `memset` will be provided in a separate PR on top of this fix.

### DIFF
--- a/include/alpaka/api/host/Queue.hpp
+++ b/include/alpaka/api/host/Queue.hpp
@@ -385,6 +385,11 @@ namespace alpaka::onHost
                 ALPAKA_LOG_FUNCTION(onHost::logger::memory + onHost::logger::queue);
                 constexpr auto dim = alpaka::trait::getDim_v<T_Extents>;
 
+                // use always 64bit precision to avoid overflows in the pitch calculations
+                auto extentMd = pCast<size_t>(extents);
+                if(extentMd.product() == size_t{0u})
+                    return;
+
                 /* Get all required properties outside the lambda function to not extend the life-time of the data.
                  * The life-time is not extended to have some life-time behaviours with all backends.
                  */
@@ -394,36 +399,38 @@ namespace alpaka::onHost
                 if constexpr(dim == 1u)
                 {
                     queue.submit(
-                        [extents, destPtr, srcPtr]()
+                        [numElementsInX = extentMd.x(), destPtr, srcPtr]()
                         {
-                            std::memcpy(destPtr, srcPtr, extents.x() * sizeof(alpaka::trait::GetValueType_t<T_Dest>));
+                            std::memcpy(
+                                destPtr,
+                                srcPtr,
+                                numElementsInX * sizeof(alpaka::trait::GetValueType_t<T_Dest>));
                         });
                 }
                 else
                 {
                     // memcpy is implemented as row wise copy therefore the last dimension is not required
-                    auto destPitchBytesWithoutColumn = dest.getPitches().eraseBack();
-                    auto sourcePitchBytesWithoutColumn = source.getPitches().eraseBack();
+                    auto destPitchBytesWithoutColumn = pCast<size_t>(onHost::getPitches(dest).eraseBack());
+                    auto sourcePitchBytesWithoutColumn = pCast<size_t>(onHost::getPitches(source).eraseBack());
 
                     queue.submit(
-                        [extents, destPtr, srcPtr, destPitchBytesWithoutColumn, sourcePitchBytesWithoutColumn]()
+                        [extentMd, destPtr, srcPtr, destPitchBytesWithoutColumn, sourcePitchBytesWithoutColumn]()
                         {
-                            auto const dstExtentWithoutColumn = extents.eraseBack();
-                            if(static_cast<std::size_t>(extents.product()) != 0u)
-                            {
-                                meta::ndLoopIncIdx(
-                                    dstExtentWithoutColumn,
-                                    [&](auto const& idx)
-                                    {
-                                        std::memcpy(
-                                            reinterpret_cast<std::uint8_t*>(destPtr)
-                                                + (idx * destPitchBytesWithoutColumn).sum(),
-                                            reinterpret_cast<std::uint8_t const*>(srcPtr)
-                                                + (idx * sourcePitchBytesWithoutColumn).sum(),
-                                            static_cast<size_t>(extents.back())
-                                                * sizeof(alpaka::trait::GetValueType_t<T_Dest>));
-                                    });
-                            }
+                            alpaka::concepts::Vector<size_t> auto const dstExtentWithoutColumn
+                                = pCast<size_t>(extentMd.eraseBack());
+
+                            meta::ndLoopIncIdx(
+                                dstExtentWithoutColumn,
+                                [&](auto const& idx)
+                                {
+                                    std::memcpy(
+                                        reinterpret_cast<std::uint8_t*>(destPtr)
+                                            + (idx * destPitchBytesWithoutColumn).sum(),
+                                        reinterpret_cast<std::uint8_t const*>(srcPtr)
+                                            + (idx * sourcePitchBytesWithoutColumn).sum(),
+                                        static_cast<size_t>(extentMd.back())
+                                            * sizeof(alpaka::trait::GetValueType_t<T_Dest>));
+                                });
                         });
                 }
             }
--- a/include/alpaka/api/host/Queue.hpp
+++ b/include/alpaka/api/host/Queue.hpp
@@ -379,6 +379,11 @@ namespace alpaka::onHost
         template<typename T_Device, typename T_Dest, typename T_Source, typename T_Extents>
         struct Memcpy::Op<cpu::Queue<T_Device>, T_Dest, T_Source, T_Extents>
         {
+            /** Perform data copy.
+             *
+             * To understand the usage of pitches to shift pointers within the implementation see
+             * https://alpaka3.readthedocs.io/en/latest/advanced/datastorage.html#pitches
+             */
             void operator()(cpu::Queue<T_Device>& queue, auto&& dest, T_Source const& source, T_Extents const& extents)
                 const requires std::same_as<ALPAKA_TYPEOF(dest), T_Dest>
             {
--- a/include/alpaka/api/oneApi/Queue.hpp
+++ b/include/alpaka/api/oneApi/Queue.hpp
@@ -73,6 +73,11 @@ namespace alpaka::onHost::internal
     requires(alpaka::trait::getDim_v<T_Extents> > 1u)
     struct internal::Memcpy::Op<syclGeneric::Queue<T_Device>, T_Dest, T_Source, T_Extents>
     {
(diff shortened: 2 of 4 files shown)


include/alpaka/api/oneApi/Queue.hpp
```diff
@@ -95,28 +99,142 @@ namespace alpaka::onHost::internal
...
+                     * This is more efficient than calling the sycl memcopy function multiple times.
+                     */
+                    alpaka::concepts::Vector<size_t, 2u> auto mappedExtentMd = extentMd.template rshrink<2u>();
+                    // remove x dimension, fuse all other dimensions into the y component
+                    mappedExtentMd.y() = extentMd.eraseBack().product();
+                    using VecIdxType = ALPAKA_TYPEOF(mappedExtentMd);
+
+                    ev = memcopy2D<alpaka::trait::GetValueType_t<T_Dest>>(
+                        sycl_queue,
+                        // 3D is nativ supported therefore we can handle the memcpy with a single call
```
SimeonEhrig: Do you mean 2D is native supported?

test/unit/mem/memcpy.cpp
```diff
@@ -171,3 +177,114 @@ TEMPLATE_LIST_TEST_CASE("memcopy test", "", DeviceSpecs)
...
+                    REQUIRE(refIotaCounter == resultFull[idx]);
+                    ++refIotaCounter;
+                }
+            });
+        // check that we touched at least once an inner value
+        REQUIRE(refIotaCounter != 0);
+    }
+}
+
+TEMPLATE_LIST_TEST_CASE("memcopy subview test", "", DeviceSpecs)
```
SimeonEhrig, line 239: I strongly suggest a to add an extra test with hand crafted, fixed input data and expected result data.
My personal experience is, if you have logical miss take in your mind, you do it in the implementation and the generator logic of your tests.

My suggestion. Create a 6 dimension array where each element contains a 6 vec with it's coordinate. Than think about one or more subviews. For each subview, create it and copy the data to two different output buffer.
1. The output buffer is allocated the with the size of the sub view.
2. The output buffer is allocated with the size of the input view, all values are set to `Vec::file(0)` and a sub view is created.

I both cases you can easily verify with 6 nested loops, if the values of the vec is correct. In the second case you can also check if all values outside the view are still 0.

We can do this test also for a 6D array without pitches and a 6D array with manually specified pitches.

SimeonEhrig (commented):
I didn't finish my review yet, because each api has a different solution.

But I have already an idea for an additional test cases because the existing tests have some generator logic and may have the same possible logical miss takes, like the implementation.

include/alpaka/api/oneApi/Queue.hpp
psychocoderHPC: it is a copy past bug, yes it should be 2D

include/alpaka/api/host/Queue.hpp
```diff
@@ -394,36 +399,38 @@ namespace alpaka::onHost
...
-                            }
+                            alpaka::concepts::Vector<size_t> auto const dstExtentWithoutColumn
+                                = pCast<size_t>(extentMd.eraseBack());
+
+                            meta::ndLoopIncIdx(
+                                dstExtentWithoutColumn,
+                                [&](auto const& idx)
+                                {
+                                    std::memcpy(
+                                        reinterpret_cast<std::uint8_t*>(destPtr)
```
SimeonEhrig, line 432: It took me some time to understand, why this address calculation works. If you understand how our pitches are defined and does some math by yourself, it make sense but I think a comment can shortcut this process.


Maybe you should add a comment (and also on other places).

```suggestion
/* Multidimensional indices without pitches are typically calculated using the formula `idx(n) * size(n+1) * size(n+2) * … * size(dim-1) + idx(n+1) * size(n+2) * size(n+3) * … *  size(dim-1) + … + idx(dim-1)`, where dim is the dimension and n is the index starting at 0 and ending at dim-1. In alpaka, the pitch for each dimension is `size(n+1) * size(n+2) * … *  size(dim-1)`, where `size()` not only returns the number of elements multiplied by the byte size per element but also adds the pitch bytes. Therefore a single multiplication per dimension is required. */
                                        reinterpret_cast<std::uint8_t*>(destPtr)
```

test/unit/mem/memcpy.cpp
psychocoderHPC: If I understand you correctly, this is what the test is doing.
The generators are used only to define the size of the buffer and than the real test in `memcpySubViewTest()` is called for all memory copy directions, `host->device`,`device->host`,`host->host` and `device->device`.

IMO, writing a single handcrafted test would not solve any issue. A single 6D test is also saying nothing, it is equal to 3D for sycl or 4D for cuda. That why this test is testing at least one example extent up to 5D.
The critical thing is not more dimensions, it is the next higher dimension based on the native copy method in the API. In OneApi it is 3D for CUDA/HIP it is 4D.

psychocoderHPC: note: I tested if the old implemenation passed the test without fixes. It failed as expected.

SimeonEhrig: The single test is for proving if you test implementation is correct. This is something which I learned with bashi.

If you can no longer imagine how a test verifies the test data, then you need two levels of testing. The first level involves achieving as complete coverage as possible of the entire problem domain. With the given scale of the problem, this can only be achieved using generated test data and verification. And a second level that performs spot checks to verify whether the first level of testing might be correct.

I’ve learned from experience that if I have a logical error in the production code, I’ll carry that error over to the test data generator and the expected results. In this case the entire test chain is incorrect but the test itself appears to be correct.

And I can’t imagine how 5D is stored in linear memory with padding.

I choose simply 6D because CUDA/HIP supports 3D memcpy. Therefore we need to handle the other 3 dimensions by ourself. Index calculation of 1D and 2D is more or less trivial. The problems starts with 3D. Therefore 6D.

SimeonEhrig: If you don't want to implement and additional test, we can also use an additional human instead a computer. @ikbuibui told me, that he found the bug.

In my case, without additional help of the computer, I can only assume that the problem is solved but not approve.

psychocoderHPC: The line 232 is a sanity check that the test is not only go into the guard region and allow passing the test without checking the inner part.

```
// check that we touched at least once an inner value
        REQUIRE(refIotaCounter != 0)
```

I agree that there is always the question if the test is correct but I would not become paranoid and start testing tests. 
The test is in line 193 - 213 and I would say it is something understandable. Writing the test again is IMO not helping.

> If you can no longer imagine how a test verifies the test data, then you need two levels of testing.

As I said the test is less than 20 lines and not hard to understand.


>  @ikbuibui told me, that he found the bug.

This was in another project where the bug was shown first because we never used mutidimensional copy of subviews in test cases or examples. We had no test before that, so we have not found the bug. The test case is fully written by me.

> If you don't want to implement an additional test, we can also use an additional human instead a computer.

Yes, another pair of eyes is never bad. 


note: Please do not overcomplicate things. 
- We found a bug because of an missing test.
- We implemented the test, and reproduces the issue.
- We fixed the route of the issue.
- The new tests are now passing.

If we become paranoid any time a user finds a bug and we implement a reproducer as test case we would never go on within the project.

SimeonEhrig: I asked for an additional test, because of the estimation for this bug:

1. The first implementation was wrong, therefore it is more complicated as expected.
2. The problem is so complex, that I cannot solve it anymore by myself. I need help from the computer to prove the solution.
3. I'm not sure, if it is easy discoverable in an real world application if this implementation is still wrong. Maybe the end result is slightly different but wrong and you have no pattern in you data. You will not see, if there is a shift or transportation in the data copy.

Don't assume I will do this for each bug. That's only my experience in this such kind of bug.

include/alpaka/api/host/Queue.hpp
psychocoderHPC: Your description uses `size()` once as an element and once as bytes. 
Nevertheless, I think we should not repeat everywhere in the code base the basics of alpaka, this would create many duplicates everywhere. And this is the backend of alpaka the user will never touch.
If you want I can add the [link to the documentation of pitches](https://alpaka3.readthedocs.io/en/latest/advanced/datastorage.html#pitches) where we describe this. 
The doc got also 2 weeks ago an update by @ikbuibui in PR #568.

psychocoderHPC: I linked the pitch documentation in the doxygen of the function. Adding it directly to the line would mean I need to add it in the other implementation also to many places.

test/unit/mem/memcpy.cpp
ikbuibui: Sorry if I gave the wrong impression, I did not find the bug.

test/unit/mem/memcpy.cpp
```diff
@@ -205,11 +205,14 @@ void memcpySubViewTest(auto& copyQueue, auto& destDevice, auto& srcDevice, alpak
         alpaka::concepts::IView auto output = outputFull.getSubView(negGuard, extents - negGuard - posGuard);
 
         alpaka::concepts::IBuffer auto resultFull = onHost::allocHost<T_DataType>(extents);
-        onHost::fill(srcQueue, inputFull, T_DataType{0});
```
psychocoderHPC, line 208: Unnecessary operation, next line is overwriting it.

test/unit/mem/memcpy.cpp
```diff
@@ -205,11 +205,14 @@ void memcpySubViewTest(auto& copyQueue, auto& destDevice, auto& srcDevice, alpak
...
+
         onHost::iota(srcQueue, T_DataType{0}, input);
         onHost::fill(destQueue, outputFull, T_DataType{42});
-        // Overwrite the inner part
+        /* Overwrite the inner part, this is the operation we would like to validate in this test setup.
+         * It is operating on subviews.
+         */
         onHost::memcpy(copyQueue, output, input);
+        // Use copy of the full buffer, this is already validated in a separate test.
```
psychocoderHPC, line 215: Add some comments to help the reader to understand why the full buffer is copied and not subviews only.

SimeonEhrig:
@psychocoderHPC  After all, the review is out of my scope. Ether you find somebody else or simply force merge it. I cannot do more.

psychocoderHPC:
I will merge it to fix the bug, we should have a look on in with a whiteboard next time in person.

SimeonEhrig:
> I will merge it to fix the bug, we should have a look on in with a whiteboard next time in person.

A whiteboard cannot solve the problem. That's the actual problem.

PR #583, opened 2026-06-17
CI: fix GitHub workflows

psychocoderHPC:
For unknown reasons the jobnames selectable as required jobs before merging is allowed it not matching the job names created by the github workflow engine.
This PR is adding a dependent job that validates all matrix jobs.

This as as positive side effect that we can not forget to add a new job as required and workarounds the anming issue in GitHub which we see since a week.


psychocoderHPC:
I will merge this PR, that other pull requests can be rebased to have a clean state of the CI.

PR #584, opened 2026-06-17
Critical: fix n-dimensional memset

psychocoderHPC:
The memset operation with n-dimensional data was broken if a sub-view was used.

Cuda: memset with more than 3 dimensions is broken
OneApi: memset with more than 2 dimensions is broken
Host: all works as expected

- add test to check the full and sub-view memset method
- fix backends
- use size_t for all calculations to avoid index range overflows
- second commit contains a workaround for a GCC15 false positive

The fix uses the same structure for the fix and tests we use in https://github.com/alpaka-group/alpaka3/pull/580.

- [x] rebase against #580  (first commit)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a GCC 15 compiler false-positive warning in vector stream insertion.
  * Improved memset precision for multi-dimensional memory operations with consistent 64-bit calculations.
  * Enhanced memset robustness for complex memory layouts and non-contiguous buffers.

* **Tests**
  * Added comprehensive unit tests for memset operations on full and partial buffer regions.

### DIFF
--- a/include/alpaka/Vec.hpp
+++ b/include/alpaka/Vec.hpp
@@ -754,7 +754,13 @@ namespace alpaka
     template<typename Type, uint32_t T_dim, typename T_Storage>
     std::ostream& operator<<(std::ostream& s, Vec<Type, T_dim, T_Storage> const& vec)
     {
-        return s << vec.toString();
+        /* Usage of a temporary universal vector variable is required as workaround to silent an error with gcc 15
+         * error: 'frameSpec' may be used uninitialized [-Werror=maybe-uninitialized] when compiling with 'ubsan'
+         * enabled. It is a false positive in some cases where a CVec is used. Since this method is not performance
+         * critical this workaround should be acceptable.
+         */
+        typename Vec<Type, T_dim, T_Storage>::UniVec v(vec);
+        return s << v.toString();
     }
 
 #define ALPAKA_VECTOR_BINARY_OP(typenameOrConcept, resultScalarType, op)                                              \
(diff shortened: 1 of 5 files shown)


SimeonEhrig:
Looks like the same logic in #580. Therefore it can be force merge if no reviewer is found.

psychocoderHPC:
ok, I need to figure out at first why the CI is failing on a file that is not touched, and no PR before triggered the issue

```
In function 'std::ostream& alpaka::onHost::operator<<(std::ostream&, const auto:99&) [with auto:99 = FrameSpec<alpaka::Vec<unsigned int, 3, alpaka::ArrayStorage<unsigned int, 3> >, alpaka::Vec<unsigned int, 3, alpaka::detail::CVec<unsigned int, 4, 4, 2> >, alpaka::exec::CpuOmpBlocks>]',
    inlined from 'void testVectorAddKernel3D(auto:896, auto:897) [with auto:896 = alpaka::onHost::Device<alpaka::api::Host, alpaka::deviceKind::Cpu>; auto:897 = alpaka::exec::CpuOmpBlocks]' at /__w/alpaka3/alpaka3/example/tutorial/src/07_chunkedData.cpp:270:82:
/__w/alpaka3/alpaka3/include/alpaka/onHost/FrameSpec.hpp:145:82: error: 'frameSpec' may be used uninitialized [-Werror=maybe-uninitialized]
  145 |         return s << "FrameSpec{ frames=" << d.getNumFrames() << ", frameExtent=" << d.getFrameExtents() << " }";
```

PR #585, opened 2026-06-18
warning for uncompatible nvcc host compiler

TimHanel00:
I ran into an issue where alpaka deactivated the CUDA backend without making the actual cause obvious. The problem only became clear after compiling a small CUDA “Hello World” device program: CUDA 12.8 was not compatible with the GCC 15.x host compiler.

From my understanding, the compatibility checks happen inside check_language(CUDA). If CUDA cannot be enabled, CMAKE_CUDA_COMPILER is set to NOTFOUND, even though a local CUDA installation and nvcc may still be present.

This PR adds an else() path that checks whether nvcc is available while CMAKE_CUDA_COMPILER is unset or NOTFOUND. If such a mismatch is detected, CMake now emits a warning. This should make CUDA host compiler incompatibilities and similar configuration problems much easier to identify.

### DIFF
--- a/cmake/alpakaCuda.cmake
+++ b/cmake/alpakaCuda.cmake
@@ -26,6 +26,7 @@ check_language(CUDA)
 
 if(CMAKE_CUDA_COMPILER)
     enable_language(CUDA)
+    message(STATUS "CUDA language is available")
     checkcompilercxxsupport(CUDA ${alpaka_CXX_STANDARD})
 
     if(NOT TARGET alpaka::cuda)
@@ -135,5 +136,27 @@ if(CMAKE_CUDA_COMPILER)
     ## GCC compiler flag to show a longer stack for concept diagnostics
     if(${_alpaka_CUDA_HOST_COMPILER} STREQUAL "GNU")
         alpaka_set_compiler_options(HOST target alpaka_target_cuda "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-fconcepts-diagnostics-depth=${alpaka_GCC_CONCEPT_DEPTH}>")
+    endif()
+else()
+    find_package(CUDAToolkit)
+    if(NOT CUDAToolkit_FOUND)
+        message(
+            WARNING
+            "CUDA support was requested, but no Cuda Compiler/Toolchain has been found. "
+            "CUDA backends will be disabled. "
+        )
+    else()
+        if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.26)
+            set(_alpaka_CMAKE_CONFIGURE_LOG "${CMAKE_BINARY_DIR}/CMakeFiles/CMakeConfigureLog.yaml")
+        else()
+            set(_alpaka_CMAKE_CONFIGURE_LOG "${CMAKE_BINARY_DIR}/CMakeFiles/CMakeError.log")
+        endif()
+        message(
+            FATAL_ERROR
+            "CUDA support was requested, but CMake did not accept "
+            "CUDA as a language. This usually means that there is a compatibility problem, "
+            "for example an unsupported host compiler. Check "
+            "'${_alpaka_CMAKE_CONFIGURE_LOG}' for further information."
+        )
     endif()
 endif()
--- a/cmake/alpakaHip.cmake
+++ b/cmake/alpakaHip.cmake
@@ -16,6 +16,7 @@ check_language(HIP)
 
 if(CMAKE_HIP_COMPILER)
     enable_language(HIP)
+    message(STATUS "HIP language is available")
     find_package(hip REQUIRED)
 
     set(_alpaka_HIP_MIN_VER 6.0)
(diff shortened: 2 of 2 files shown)


psychocoderHPC:
@TimHanel00 can you please rebase your PR, I assume you branched out before #583 was merged.

cmake/alpakaCuda.cmake
```diff
@@ -136,4 +136,24 @@ if(CMAKE_CUDA_COMPILER)
     if(${_alpaka_CUDA_HOST_COMPILER} STREQUAL "GNU")
         alpaka_set_compiler_options(HOST target alpaka_target_cuda "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-fconcepts-diagnostics-depth=${alpaka_GCC_CONCEPT_DEPTH}>")
     endif()
+else()
+    find_program(
```
psychocoderHPC: I am not sure but if you compile with `clang` as CUDA compiler, you do not need nvcc on your system, the cuda library is all you need, this would trigger you warning.

Can you test if we can maybe get all information already via this

```
check_language(CUDA)

if(CMAKE_CUDA_COMPILER)
    enable_language(CUDA)
    message(STATUS "CUDA enabled: ${CMAKE_CUDA_COMPILER}")
else()
    message(STATUS "CUDA language not available")

    set(_cuda_log "${CMAKE_BINARY_DIR}/CMakeFiles/CMakeError.log")
    if(EXISTS "${_cuda_log}")
        file(READ "${_cuda_log}" _cuda_error_log)

        string(REGEX MATCH
            "The CUDA compiler[^\n]*(\n|.)*"
            _cuda_error "${_cuda_error_log}"
        )

        if(_cuda_error)
            message(STATUS "CUDA check failed with:\n${_cuda_error}")
        else()
            message(STATUS "See ${_cuda_log} for details.")
        endif()
    else()
        message(STATUS "No CMakeError.log found. See ${CMAKE_BINARY_DIR}/CMakeFiles/")
    endif()
```

TimHanel00: Yes, this sounds like a reasonable change. I also tested it with the incompatible CUDA/GCC 15.x configuration.
However, I think the diagnostic message is still easy to miss in a real build log, and the provided regex did not catch the actual error in my case. One reason is that, starting with CMake 3.26, configure diagnostics are written to CMakeConfigureLog.yaml instead of CMakeError.log.
I updated the commit to check both diagnostic log locations and broadened the regex pattern.

I also noticed that this configuration still causes build problems afterwards, even though CUDA is not available and the CUDA backend should be disabled. For example, the scan example still reaches CUDA/CUB headers and then fails with:

```
cub/util_ptx.cuh:199:3: error: '__syncthreads' was not declared in this scope
  199 |   __syncthreads();
```
  
  Should I create a seperate PR/Issue for that?

psychocoderHPC: We could simply route the user to the possible files with more information instead of parsing them.
What I dislike in my suggestion is that we parse the log.
In principle, we try to solve an issue within CMake. I am not sure if there is another way. 
We could also simply compile a `.cu` file with a kernel to maybe a better error message.

https://github.com/alpaka-group/alpaka3/blob/72fa3b8e72ac51bdd103e9a6ef1e59900ac456b0/cmake/alpakaCommon.cmake#L281-L308

That's a common CMake way to test e.g. for CXX features.

The user is responsible for using a valid compiler combination; we can only try to detect it, but if it becomes to complicated and we do not get enough information from CMake for a clean error message, we can not solve it.

At the moment, I would say let us try the test compile, if this works. If not, have a short chat next week about our findings and possible solutions.

btw: Welcome to the hell of CMake^^.

TimHanel00: Ok since `try_compile` doesnt work, because it requires the CUDA language to be enabled. I opted back to refer to the error logs  - rather than parsing them using REGEX. I saw no way to make the error surface directly by using enable_language(CUDA) without corrupting the cmake cache in subsequent cmake passes. However your previous solution omitted the case where the user wants to activate CUDA by setting alpaka_DEP_CUDA yet there is no CUDA toolchain found on that system. But in order to distinguish between "not present" and "wrong configuration" looking for the `CUDAToolkit` is the better option than trying to find nvcc - as already deployed in the `scan` example -  because this will also surface in the clang compiler setup. 

I also escalated the cuda/hip compiler not present case to a `WARNING` and the incompatibility case to a `FATAL_ERROR`.

cmake/alpakaHip.cmake
```diff
@@ -75,7 +76,28 @@ if(CMAKE_HIP_COMPILER)
...
+        if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.26)
+            set(_alpaka_CMAKE_CONFIGURE_LOG "${CMAKE_BINARY_DIR}/CMakeFiles/CMakeConfigureLog.yaml")
+        else()
+            set(_alpaka_CMAKE_CONFIGURE_LOG "${CMAKE_BINARY_DIR}/CMakeFiles/CMakeError.log")
+        endif()
+        message(
+            FATAL_ERROR
+            "HIP support was requested, but CMake did not accept "
+            "HIP as a language. This usually means that there is a compatibility problem, "
+            "for example an unsupported host compiler. Check "
```
psychocoderHPC:
```suggestion
            "HIP as a language. This usually means that there is a compatibility problem. Check "
```

HIP does not have something like a separate host compiler. So we can mention that there is an issue but could not hint at the reason.
If you accept this change in the web GUI you need to rerun pre-commit locally to fix the from code style.

cmake/alpakaHip.cmake
```diff
@@ -75,7 +76,28 @@ if(CMAKE_HIP_COMPILER)
...
+            "'${_alpaka_CMAKE_CONFIGURE_LOG}' for further information."
+        )
+    endif()
 endif()
 
 target_compile_definitions(alpaka_target_hip INTERFACE ALPAKA_CMAKE_TARGET_HIP)
```
psychocoderHPC: Can you please move all lines behind and including this line into the if branch of `if(CMAKE_HIP_COMPILER)`. The code should only be executed if `HIP` is found and usable.

This was already wrong in the past because we had not handled failures correctly.
In the cuda cmake we did it correctly.

psychocoderHPC (changes requested):
One documentation change is required and some old code is not where it should be.

psychocoderHPC (approved):

PR #587, opened 2026-06-18
add correct test condition to build `scan` and `vendorAPI` examples

TimHanel00:
When the CUDA compiler is misconfigured, check_language(CUDA) may silently leave `CMAKE_CUDA_COMPILER` unset, as described in #585. In that case, relying on `alpaka_DEP_CUDA` alone is not sufficient: both examples use this flag to decide whether they can compile native CUDA code, but they may still produce compilation errors if no valid CUDA compiler/toolchain was detected.
Therefore, `CMAKE_CUDA_COMPILER` is the appropriate condition to check here, since it reflects whether CMake actually found a compatible CUDA - toolchain/Compiler.

### DIFF
--- a/example/scan/CMakeLists.txt
+++ b/example/scan/CMakeLists.txt
@@ -31,7 +31,9 @@ if(NOT TARGET alpaka::alpaka)
 endif()
 
 set(HAS_CUDA_CUB 0)
-if(alpaka_DEP_CUDA)
+# alpaka probes CUDA with check_language(CUDA),
+# which only sets CMAKE_CUDA_COMPILER if a valid CUDA compiler/toolchain is provided.
+if(CMAKE_CUDA_COMPILER)
     find_package(CUDAToolkit)
     if(NOT CUDAToolkit_FOUND)
         message(STATUS "Optional CUDAToolkit not found, cub scan implementation will be disabled.")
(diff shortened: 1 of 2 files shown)


example/scan/CMakeLists.txt
```diff
@@ -31,7 +31,8 @@ if(NOT TARGET alpaka::alpaka)
 endif()
 
 set(HAS_CUDA_CUB 0)
-if(alpaka_DEP_CUDA)
+# this flag is set by alpaka in case valid CUDA compiler/toolchain is present
```
psychocoderHPC: can you changebit to ,`this flag is set by CMake during alpaka validate the CUDA clmoiler...`

This is a variable is that by CMake not alpaka.

psychocoderHPC: please copy the text to other copies too

psychocoderHPC (changes requested):

example/scan/CMakeLists.txt
TimHanel00: I updated the comment in the attempt to describe the intended cmake flags usage here more precisely - based on your recommendation.

psychocoderHPC (approved):

PR #590, opened 2026-06-20
use unsigned index math in `Simd` class

psychocoderHPC:
The index type of `Simd` is unsigned, so try to keep it where possible.


psychocoderHPC:
The change produces 

```
/__w/alpaka3/alpaka3/include/alpaka/Simd.hpp(394): error #186-D: pointless comparison of unsigned integer with zero
                  uint32_t const sourceIdx = i >= static_cast<uint32_t>(laneIdxToRemove) ? i + 1 : i;
```

It can be suppressed, but this looks totally boring in the code. So I like to stay with using `int`. We can not get an overflow anyway because the number of lanes is small.

PR #597, opened 2026-06-22
add onHost::fill() test 

psychocoderHPC:
Add a fill test equal to the structure we use for `memcopy` and `memset` test case.

- [x] rebase against #584

Tested by hand on our development server for:
 - oneApi 2025.2 CPU, AMD-GPU, NVIDIA-GPU, CPU, Intel-GPU
 - Cuda 13.0
 - CPU

The test is nearly equal to #584 (`onHost::memset()` test)

### DIFF
--- a/test/unit/mem/fill.cpp
+++ b/test/unit/mem/fill.cpp
@@ -1,0 +1,141 @@
+/* Copyright 2026 René Widera
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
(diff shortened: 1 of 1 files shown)


SimeonEhrig:
For me it looks wrong that we have 3 times the same test for the same functionality (memcpy, fill, memset). It looks like, we can unify code.

But better a lacking test than no test.

SimeonEhrig (approved):

PR #604, opened 2026-06-25
validate pCast and lpCast

psychocoderHPC:
Add a static compile test to check that `pCast` and `lpCast` is callable. 
It is not validating the results, this should be added to the tests for the corresponding types because both cast methods change the type signature.
The casting methods are also allowed to return an absolutely different type; that's not the case at the moment, but it's not forbidden.

Fix their broken `isConvertible_v` implementation found by the test.

note: `SimdMask` is currently not tested because the value type of `SimdMask` is always `bool` even if a mask is stored as a value bit mask. How to test this is out of scope for this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the compile-time type-conversion rules used by vector operations by refining the “convertible” concept.
  * Corrected the default storage policy selection for `Vec` and `Simd` so the intended storage type is consistently used.

* **Tests**
  * Added device-side compile-time checks for `lpCast` and `pCast`, validating which casts are accepted or rejected across common `Vec`/`Simd` sizes and types (including negative/non-castable cases).

### DIFF
--- a/test/unit/utility/cast.cpp
+++ b/test/unit/utility/cast.cpp
@@ -18,53 +18,59 @@ using TestBackends = std::decay_t<decltype(onHost::allBackends(onHost::enabledDe
 template<typename T_To, typename T_AnyInput>
 concept IsLpCastCallable = requires(T_AnyInput in) { lpCast<T_To>(in); };
 
+// Test type for user type conversion.
+struct Foo
+{
+};
+
 struct TestKernelLpCast
 {
-    ALPAKA_FN_ACC void operator()(auto const& acc) const
+    /* Vec and Simd have the same required template signature, therefore we can combine the tests.
+     *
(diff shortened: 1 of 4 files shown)


test/unit/utility/cast.cpp
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    ALPAKA_FN_ACC void operator()(auto const&) const
+    {
+        check<1u, Vec>();
+        check<2u, Vec>();
+        check<3u, Vec>();
+        check<4u, Vec>();
+
+        check<1u, Simd>();
```
SimeonEhrig, line 70: Only a comment. Sometimes it looks odd because of the missing `alpaka::` namespace. Therefore I think it is a temporary class, created in the test file for the test case.

SimeonEhrig (approved):

PR #611, opened 2026-06-30
CI: add helper function quiet_run()

SimeonEhrig:
Run a command suppress the output if the error code is 0. If the error code is non 0, display output and run exit_error.

## Summary by CodeRabbit

* **New Features**
  * Added quieter, consistent command execution for installation steps, reducing noisy logs while preserving clear error output.
* **Bug Fixes**
  * Improved colored/status console messages to display all provided text (not just the first word).
  * Standardized Linux package installation to be non-interactive and avoid recommended extras across toolchain, ROCm, and sudo fallback flows.

### DIFF
--- a/script/ci/utils/misc.sh
+++ b/script/ci/utils/misc.sh
@@ -12,17 +12,17 @@ source "${APCI_ALPAKA_ROOT}/script/ci/utils/color_echo.sh"
 
 # display a install message in green
 install_msg() {
-    echo_green "[INSTALL]: $1"
+    echo_green "[INSTALL]: $*"
 }
 
 # display a install message in green
 script_msg() {
-    echo_green "[SCRIPT]: $1"
+    echo_green "[SCRIPT]: $*"
 }
 
 # display a error message in red
 error_msg() {
-    echo_red "[ERROR]: $1"
+    echo_red "[ERROR]: $*"
 }
 
 exit_error() {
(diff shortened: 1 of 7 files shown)


script/ci/utils/misc.sh
```diff
@@ -85,6 +85,47 @@ retry_cmd() {
...
 }
 
+# Run a command suppress the output if the error code is 0.
+# If the error code is non 0, display output and run exit_error.
+quiet_run() {
+    if [[ $# -lt 1 ]]; then
+        exit_error "quiet_run requires at least one argument."
+    fi
+
+    echo_green "$*"
```
psychocoderHPC, line 95:
```suggestion
    echo_green "$@"
```

IMO this should avoid showing wrong escapes arguments

script/ci/utils/misc.sh
```diff
@@ -85,6 +85,47 @@ retry_cmd() {
...
+    fi
+
+    if [[ "${exit_code}" != 0 ]]; then
+        cat "${log_file}"
+    fi
+
+    rm "${log_file}"
+
+    if [[ ${exit_code} -ne 0 ]]; then
+        exit_error "failed cmd: $*"
```
psychocoderHPC:
```suggestion
        exit_error "failed cmd: $@"
```

script/ci/utils/misc.sh
```diff
@@ -12,17 +12,17 @@ source "${APCI_ALPAKA_ROOT}/script/ci/utils/color_echo.sh"
 
 # display a install message in green
 install_msg() {
-    echo_green "[INSTALL]: $1"
+    echo_green "[INSTALL]: $*"
```
psychocoderHPC: I would use $@ everywhere.

psychocoderHPC (changes requested):

script/ci/utils/misc.sh
SimeonEhrig: shellcheck suggest to use `$*` instead of `$@`.

psychocoderHPC (approved):

PR #617, opened 2026-07-06
fix warp in grid calculation

psychocoderHPC:
## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected grid-level warp indexing to account for the configured number of warps per block, ensuring global warp positions are computed consistently.
  * Resolved cases where warp indices could be misaligned when multiple warps exist within a single block.
* **Tests**
  * Expanded warp size and indexing validation to cover different warps-per-block configurations, improving confidence in runtime sizing behavior.


SimeonEhrig (approved):

psychocoderHPC:
@SimeonEhrig I improved the test to detect the overlooked issue.

SimeonEhrig (approved):

PR #618, opened 2026-07-06
restrict the number of parallel AMD GPU runtime tests

SimeonEhrig:
- Looks like the GPU of the GPU runner gets problems, if to many CI jobs are executed in paralllel.
- Temporary solution until the job generator is implemented.

## Summary by CodeRabbit

* **Chores**
  * Updated the CI pipeline to enforce clearer job ordering for several ROCm-related builds.
  * This should improve build consistency and reduce issues caused by jobs running before their prerequisites are ready.


SimeonEhrig:
Does not work, because if there many PR's running in parallel, the Runner is still fully utilized.

PR #623, opened 2026-07-07
refactor `concepts::Backend`

psychocoderHPC:
- rename to `concepts::BackendSpec`
- update documentation
- add `onHost::makeSpec()` factory
- remove `alpaka::object::deviceSpec`
- refactor `onHost::createBackendsFor()`
- use the refactored interface that supports BackendSpec

## Summary by CodeRabbit

* **New Features**
  * Streamlined backend/device setup across examples and tests using higher-level helpers to derive device specs and executors.
* **Bug Fixes**
  * Added/standardized explicit device availability checks before executing, preventing unsupported-backend runs.
* **Documentation**
  * Updated tutorials and backend terminology/snippets to match the new backend-spec and device-selector workflow.
* **Breaking Changes**
  * Updated core concepts and selection APIs (e.g., shift toward backend-spec-based dispatch) and removed older conveniences that relied on direct backend dictionary access. 
* **Tests**
  * Refreshed unit tests, benchmarks, and test helpers to use the new selection pattern.

### DIFF
--- a/docs/source/basic/terms.rst
+++ b/docs/source/basic/terms.rst
@@ -223,11 +223,16 @@ Executor
 
 .. _backend:
 
-Backend
--------
+Backend Spec
+------------
 
-A ``Backend`` is the combination of a ``DeviceSpec`` and an ``Executor``. It is typically returned by ``onHost::allBackends(...)``.
+A ``BackendSpec`` (backend specification) provides three properties:
 
+The ``api`` and ``deviceKind``, which describe a ``DeviceSpec``.
+The ``executor``, which describes how thread parallelism is mapped onto the device according to the device specification.
+
+It is typically returned by ``onHost::allBackends(...)``.
+Use ``alpaka::getExecutor(backend)`` to query its executor, and ``alpaka::onHost::DeviceSpec{backend}`` to recover its device specification.
 .. _thread_spec:
 
 Thread Spec
--- a/docs/source/basic/terms.rst
+++ b/docs/source/basic/terms.rst
@@ -221,18 +221,16 @@ Kernel
 Executor
 --------
 
+The ``executor``, which describes how thread parallelism is mapped onto the device according to the device specification.
+
 .. _backend:
 
 Backend Spec
 ------------
 
-A ``BackendSpec`` (backend specification) provides three properties:
+A ``BackendSpec`` (backend specification) provides two properties:
+The :ref:`DeviceSpec <device_kind>` and the :ref:`executor`, it is typically returned by ``onHost::allBackends(...)``.
 
-The ``api`` and ``deviceKind``, which describe a ``DeviceSpec``.
-The ``executor``, which describes how thread parallelism is mapped onto the device according to the device specification.
-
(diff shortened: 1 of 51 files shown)


docs/source/basic/terms.rst
```diff
@@ -223,11 +223,16 @@ Executor
...
+
+A ``BackendSpec`` (backend specification) provides three properties:
 
-A ``Backend`` is the combination of a ``DeviceSpec`` and an ``Executor``. It is typically returned by ``onHost::allBackends(...)``.
+The ``api`` and ``deviceKind``, which describe a ``DeviceSpec``.
+The ``executor``, which describes how thread parallelism is mapped onto the device according to the device specification.
 
+It is typically returned by ``onHost::allBackends(...)``.
+Use ``alpaka::getExecutor(backend)`` to query its executor, and ``alpaka::onHost::DeviceSpec{backend}`` to recover its device specification.
 .. _thread_spec:
```
SimeonEhrig:
```suggestion
A ``BackendSpec`` (backend specification) provides two properties:

The :ref:`DeviceSpec <device_kind>` and the :ref:`executor` It is typically returned by ``onHost::allBackends(...)``.

.. _thread_spec:
```

Some points about the change:
 
- We should not start to count sub components. A backend stores a DeviceSpec. If you want to know, what a DeviceSpec contains, check the DeviceSpec. Otherwise we get trouble at other positions, where we use more complicated objects, like mdspan (has 4 properties and some of the properties has also many properties)
- Use links instead repeating information. Make it easier to update and extent the documentation.
- In the terms section, I try to avoid source code, because the idea is first to explain with the terms and then work with it in the code. I see a strong benefit to name `onHost::allBackends(...)`. But the rest looks like more small details which are not beneficial here.
- A empty line is required between the text and `.. _thread_spec:` that the anchor works.

docs/source/basic/terms.rst
```diff
@@ -223,11 +223,16 @@ Executor
 
 .. _backend:
```
SimeonEhrig, line 226:
```suggestion

The ``executor``, which describes how thread parallelism is mapped onto the device according to the device specification.

.. _backend:
```

include/alpaka/core/DictTraits.hpp
```diff
@@ -0,0 +1,41 @@
+/* Copyright 2024 René Widera, Tim Hanel
```
SimeonEhrig:
```suggestion
/* Copyright 2026 René Widera, Tim Hanel
```

SimeonEhrig (changes requested):

docs/source/basic/terms.rst
psychocoderHPC: the source code snippets was already in, you merged it yesterday.

SimeonEhrig: But only the `onHost::allBackends(...)` and I was already thinking about if this is necessary. But without the function, we actual never work with a backend.

SimeonEhrig (approved):

PR #625, opened 2026-07-08
CI: report more device details

psychocoderHPC:
In case a job is failing this PR is reporting a detailed overview about the device.

## Summary by CodeRabbit

* **Tests**
  * Test output now includes more detailed device information, such as device properties and available global memory, when a device is selected during testing.

### DIFF
--- a/test/include/alpakaTest/deviceHelper.hpp
+++ b/test/include/alpakaTest/deviceHelper.hpp
@@ -50,7 +50,8 @@ namespace alpaka::test
         }
 
         onHost::Device device = devSelector.makeDevice(0);
-        UNSCOPED_INFO("Device: " << device.getName());
+        UNSCOPED_INFO("Device: \n" << device.getDeviceProperties());
+        UNSCOPED_INFO("Free memory (bytes): " << device.getFreeGlobalMemBytes());
         return device;
     }
 


SimeonEhrig:
@psychocoderHPC I found a problem. There is no output in the logs, if a test fails. I played a little bit around. We have three major problems.
- In a case of an error, ctest only displays logging messages from the sterr. All logging macros execept `FAIL` and `FAIL_CHECK` prints to stdout.
- If we use `FAIL` or `FAIL_CHECK`, the test is marked as failed. Therefore we cannot use it for logging.
- In a case of a segmentation fault (maybe also other cases of "unexpected" exits) there is no log from Catch2.

psychocoderHPC:
> @psychocoderHPC I found a problem. There is no output in the logs, if a test fails. I played a little bit around. We have three major problems.
> 
> * In a case of an error, ctest only displays logging messages from the sterr. All logging macros execept `FAIL` and `FAIL_CHECK` prints to stdout.
> * If we use `FAIL` or `FAIL_CHECK`, the test is marked as failed. Therefore we cannot use it for logging.
> * In a case of a segmentation fault (maybe also other cases of "unexpected" exits) there is no log from Catch2.

Yes you are right; only `std::cerr` is guaranteed to be printed because it is unbuffered.
Nevertheless, even in normal issues e.g. where `CHECK()` or `REQUIRES()` is failing the additional information added in this PR can help.

SimeonEhrig:
This CI test has useful output: https://gitlab.com/hzdr/crp/alpaka3/-/jobs/15384977418

SimeonEhrig:
I opened a issue at Catch2 to address the issue: https://github.com/catchorg/Catch2/issues/3183

PR #631, opened 2026-07-10
add `Queue::enqueueNativeFn()`

psychocoderHPC:
Add a function so that native API operations on a queue can be exposed to alpaka to keep the alpaka queue behaviour consistent for the user. It ensures that an operation on a blocking queue blocks the execution.

Without this PR Sycl, HIP, Host and CUDA function calls with the function API of alpaka would result in different behaviours. Now anything enqueued in the alpaka queue is known by alpaka and is blocking in cases the queue is blocking.

## Summary by CodeRabbit

* **New Features**
  * Added a unified way to enqueue “native” vendor operations on host-side queues (CPU, SYCL, CUDA, HIP), passing the underlying native queue/stream handle to the operation.

* **Bug Fixes**
  * Improved ordering and synchronization for queued transforms by routing them through native-enqueue paths.
  * Ensured CUDA/HIP native transform work runs asynchronously on the configured stream.
  * Blocking queues now properly wait for native operations to finish.

### DIFF
--- a/include/alpaka/api/host/Queue.hpp
+++ b/include/alpaka/api/host/Queue.hpp
@@ -217,6 +217,12 @@ namespace alpaka::onHost
                 m_workerThread.submit(task);
             }
 
+            void enqueueNativeFn(auto const& fn)
+            {
+                ALPAKA_LOG_FUNCTION(onHost::logger::queue);
+                submit([queueId = getNativeHandle(), fn]() { fn(queueId); });
+            }
+
             friend struct alpaka::internal::GetDeviceType;
 
             auto getDeviceKind() const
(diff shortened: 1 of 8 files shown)


include/alpaka/api/host/Queue.hpp
```diff
@@ -217,6 +217,12 @@ namespace alpaka::onHost
                 m_workerThread.submit(task);
             }
 
+            void enqueueNativeFn(auto const& fn)
+            {
+                ALPAKA_LOG_FUNCTION(onHost::logger::queue);
+                m_workerThread.submit([queueId = getNativeHandle(), fn]() { fn(queueId); });
```
psychocoderHPC: bug it must be submit() only and not passed into the async callback thread

SimeonEhrig (approved):

psychocoderHPC:
@SimeonEhrig Thanks for the review

PR #632, opened 2026-07-13
[DO NOT MERGE] This is new nice feature.

SimeonEhrig:
This is only for demonstration for the students.


SimeonEhrig:
The students are happy with the presentation and we also discussed the CI problem. Therefore I close the PR.

PR #639, opened 2026-07-15
Fix repeated CMake target initialization in external projects

TimHanel00:
When multiple targets call `alpaka_finalize()`, global dependency and feature detection is repeated for each target:

```text
OpenMP found: 4.5
-- std::atomic_ref<T> found
-- std::simd found
-- CUDA language is available
-- cuda with host compiler: GNU
-- hwloc not found, deviceKind::numaCpu will not be available (AUTO)
-- OpenMP found: 4.5
-- std::atomic_ref<T> found
-- std::simd found
-- CUDA language is available
-- cuda with host compiler: GNU
-- hwloc not found, deviceKind::numaCpu will not be available (AUTO)
-- OpenMP found: 4.5
```

The existing guard `_alpaka_TARGETS_EXTENDED` to prevent this re-occuring initialization is exported through exactly one `PARENT_SCOPE`. When using alpaka through `FetchContent`, `add_subdirectory()`, or `find_package()` - the marker can remain in a transient scope and therefore be invisible to the next alpaka_finalize() call, resulting in additional hwloc, language and std::atomic checks.
  
  Using a global property makes the guard independent of the directory or function from which `alpaka_finalize()` is called and gives it the same configure-tree-wide lifetime as the targets it protects.

## Summary by CodeRabbit

* **Chores**
  * Improved CMake configuration to avoid repeating shared backend target setup across multiple configure passes, reducing build overhead.
  * Added global early-exit guards so CUDA, HIP, and oneAPI backend configuration is applied only once per build configuration.
  * Simplified backend interface/alias creation so it initializes consistently within the relevant language-enabled paths.

### DIFF
--- a/cmake/alpakaCuda.cmake
+++ b/cmake/alpakaCuda.cmake
@@ -26,15 +26,22 @@ check_language(CUDA)
 
 if(CMAKE_CUDA_COMPILER)
     enable_language(CUDA)
+endif()
+
+get_property(_alpaka_TARGETS_EXTENDED GLOBAL PROPERTY ALPAKA_TARGETS_EXTENDED)
+if(_alpaka_TARGETS_EXTENDED)
+    # returns control back to the calling include() scope - it does not stop the cmake configure
+    return()
+endif()
+
+if(CMAKE_CUDA_COMPILER)
     message(STATUS "CUDA language is available")
     checkcompilercxxsupport(CUDA ${alpaka_CXX_STANDARD})
 
-    if(NOT TARGET alpaka::cuda)
-        add_library(alpaka_target_cuda INTERFACE)
-        add_library(alpaka::cuda ALIAS alpaka_target_cuda)
-        target_link_libraries(alpaka_target_cuda INTERFACE alpaka::host)
-        set_property(TARGET alpaka_target_cuda PROPERTY CUDA_STANDARD ${alpaka_CXX_STANDARD})
-    endif()
+    add_library(alpaka_target_cuda INTERFACE)
+    add_library(alpaka::cuda ALIAS alpaka_target_cuda)
+    target_link_libraries(alpaka_target_cuda INTERFACE alpaka::host)
+    set_property(TARGET alpaka_target_cuda PROPERTY CUDA_STANDARD ${alpaka_CXX_STANDARD})
 
     alpaka_compiler_option(CUDA_SHOW_REGISTER "Show kernel registers and create device ASM" DEFAULT)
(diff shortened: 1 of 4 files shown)


cmake/alpakaCuda.cmake
```diff
@@ -26,15 +26,21 @@ check_language(CUDA)
 
 if(CMAKE_CUDA_COMPILER)
     enable_language(CUDA)
+endif()
+
+# The language must be enabled in each consumer scope, but the global backend target only needs to be configured once.
+if(TARGET alpaka::cuda)
```
psychocoderHPC: I am not sure about this early return. This is skipping all variable definitions; what if an alpaka_CUDA variable is updated? Is the change then ignored? E.g. `alpaka_FAST_MATH` or `alpaka_CUDA_KEEP_FILES` is changed after the first configuration.

psychocoderHPC (changes requested):
I will block the PR for now, I need to check this PR first if it behaves like expected.

cmake/alpakaCuda.cmake
TimHanel00: Apparently it works because cmake alias targets are not cached in CMakeCache.txt, they get created on each configure pass, this shortcut only prevents re-initializing the same global variables multiple times during one cmake configure pass. Although - I unfortunately could not find official cmake documentation explicitly stating that these do not persist between configure runs (but also nothing that hints that alias targets such as `alpaka::cuda` indeed do persist like cached variables) - I confirmed the mechanism locally by running cmake 3.28 using `make VERBOSE=1 -j`, where I changed `alpaka_CUDA_KEEP_FILES` after the first configure and cmake emitted the correct `--keep` compiler flag for each target.

cmake/alpakaCuda.cmake
```diff
@@ -26,15 +26,21 @@ check_language(CUDA)
...
+    return()
```
TimHanel00, line 34: In an offline discussion we also explored the possibility of the return() function suddenly closing the cmake configure without handing control over to the caller that invoked include(filename). In the official cmake documentation: https://cmake.org/cmake/help/latest/command/return.html the first line directly contradicts this possibility. As far as I am concerned each file does get processed at least once from start to finish since the alias targets are created after the shortcut return(). If you are still in favor of a enclosing scope to make the conrtol flow more clear for future maintenance I will change the included .cmake files accordingly.

TimHanel00:
Following an offline discussion, this PR now uses global properties in alpakaCommon and shortcuts in each backend specific cmake file after language activation - in order to remove redundant initializations during one cmake configure, while still running necessary configuration steps for each target individually.

psychocoderHPC (approved):

PR #641, opened 2026-07-16
CI: job generator

SimeonEhrig:
# Questions
- How to handle extra SM level for CUDA jobs? At the moment, at runtime the CI ask for the SM level of GPU running in the runner. If it is a compile only job, simply use SM Level 80.
 -> will be implemented in a upcoming PR: #652
- When enable the serial executor?
 -> for every job, where GCC or Clang is the device compiler 
- When enable emulated SIMD?
 -> add to special jobs with the latest GCC and Clang version

## Summary by CodeRabbit

* **New Features**
  * Added tools to generate and validate supported CI job combinations.
  * CI jobs are now generated dynamically and split into compile-only, CPU, GPU, and special pipelines.
  * Added support for compiler-, backend-, sanitizer-, and SIMD-specific job configurations.
* **Documentation**
  * Added usage and development guidance for the job-generation and validation tools.
* **Bug Fixes**
  * Improved compiler discovery and handling of unset GPU architecture settings.
  * Fixed missing standard-library declarations in logging support.
* **Chores**
  * Added automated mypy and pylint checks to continuous integration.


script/job_generator/src/alpaka_bashi/ci_yaml/images.py
```diff
@@ -0,0 +1,145 @@
...
+    GCC,
+    HOST_COMPILER,
+    OFF_VER,
+    UBUNTU,
+)
+from typeguard import typechecked
+
+from alpaka_bashi.utils import print_warn
+
+GITLAB_IMAGE_CACHE: list[str] | None = None
```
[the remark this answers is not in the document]
SimeonEhrig: It's really unlikely, that we support different container versions at the same time.

include/alpaka/onHost/internal/logger.hpp
```diff
@@ -14,6 +14,7 @@
 #include <iostream>
 #include <ostream>
 #include <source_location>
+#include <sstream>
```
psychocoderHPC, line 17: not critical but I am wondering why `sstream` is required, this is the only C*+ change in this PR.

script/ci/install/gcc.sh
```diff
@@ -28,9 +28,7 @@ if [[ "$compiler_name" == "gcc" ]]; then
     if agc-manager -e "gcc@${compiler_version}"; then
         echo_green "use preinstalled gcc@${compiler_version}"
 
-        # TODO: because of a bug in agc-manager the gcc/g++ base path is wrong. Search for gcc/g++ instead.
-        # `APCI_CXX_COMPILER="$(agc-manager -b gcc@${compiler_version})/gcc-${compiler_version}"`
-        APCI_CXX_COMPILER=$(which "g++-${compiler_version}")
+        APCI_CXX_COMPILER=$(agc-manager -b "gcc@${compiler_version}")"/g++-${compiler_version}"
```
psychocoderHPC, line 31: question: Do we need `"gcc@${compiler_version}"` because we do not need the C compiler in alpaka or is it maybe required for CMake somehow?

script/job_generator/README.md
```diff
@@ -0,0 +1,98 @@
...
+The `job-generator` uses `bashi`'s naming convention.
+[Here](https://github.com/alpaka-group/bashi/blob/main/docs/naming.md) you can find out what the terms mean.
+
+# Install
+
+The `job-generator` is a Python package and is very easy to install.
+
+```bash
+# create a virtual environment with your favorite tool, then
+pip3 install .
```
psychocoderHPC, line 28: I am missing a `requirements.txt` file for bashi in the alpaka job generator folder that I can install all python dependencies, is there a reason for not shipping it?

psychocoderHPC (changes requested):

include/alpaka/onHost/internal/logger.hpp
SimeonEhrig: A header check with icpx failed. Was the first bug which I detected in alpaka 3 and I thought it is fine simply fix it here and do not open an extra PR.

script/ci/install/gcc.sh
SimeonEhrig: In general I use GCC as synonym for gcc and g++. It's a left over from the past. I would need to change the entry in agc-manager that `agc-manager -b "g++@${compiler_version}")` works, which means rebuilding the container.

script/job_generator/README.md
SimeonEhrig: Because I provide a real python package with a `pyproject.toml`. The `pyproject.toml` contains the dependencies. The `requirements.txt` file is intended primarily for simple scripts to describe dependencies.

script/ci/install/gcc.sh
psychocoderHPC: ok, keep it as it is for now, I was only wondering

include/alpaka/onHost/internal/logger.hpp
psychocoderHPC: it's fine, I wondered because I do not expected a C++ change in this PR

psychocoderHPC (approved):

script/ci/install/gcc.sh
SimeonEhrig: I expected more C++ changes in the PR ;-)

PR #647, opened 2026-07-22
add missing `setDevice` for HIP/CUDA

psychocoderHPC:
Most cuda/hip functions requires that the device is selected. If the user is using multiple devices from a single process we need to select the correct device before applying the native API call.

note: Same as in alpaka mainline, we do not have multi-GPU tests at the moment and cannot test it in the CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when using multiple CUDA/HIP devices.
  * Memory allocations, transfers, deallocations, and memory queries now consistently operate on the correct device.
  * Queue, stream, event, and synchronization operations now target their associated device reliably.
  * Host tasks and deferred operations execute with the appropriate device context.
  * Device property and pointer accessibility queries now return more consistent results.

### DIFF
--- a/include/alpaka/api/unifiedCudaHip/Device.hpp
+++ b/include/alpaka/api/unifiedCudaHip/Device.hpp
@@ -149,6 +149,7 @@ namespace alpaka::onHost
             {
                 std::size_t freeGlobalMemBytes(0u);
                 std::size_t globalMemCapacityBytes(0u);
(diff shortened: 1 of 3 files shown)


SimeonEhrig (approved):
Please create the issue, that the tests are missing.

psychocoderHPC:
> Please create the issue, that the tests are missing.

done #648

PR #650, opened 2026-07-23
fix inline assembler issue with clang as CUDA host compiler

psychocoderHPC:
When using clang as host compiler it is complaining about the assembler syntax.

```
api/cuda/warp.hpp:34:33: error: invalid % escape in inline assembly string
302
19:54:25
   34 | __asm__ volatile("mov.u32 %0, %laneid;" : "=r" (lIdx) :);
303
19:54:25
      |                  ~~~~~~~~~~~~~~~^~~~~~
api/cuda/warp.hpp:34:33: error: invalid % escape in inline assembly
string
302
19:54:25
   34 | __asm__ volatile("mov.u32 %0, %laneid;" : "=r" (lIdx) :);
303
19:54:25
      |                  ~~~~~~~~~~~~~~~^~~~~~~~
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected CUDA lane identification for NVCC builds, ensuring warp-level operations retrieve the correct lane index.

### DIFF
--- a/include/alpaka/api/cuda/warp.hpp
+++ b/include/alpaka/api/cuda/warp.hpp
@@ -31,7 +31,7 @@ namespace alpaka::onAcc::warp::internal
         {
             unsigned lIdx;
 #    if ALPAKA_COMP_NVCC
-            asm volatile("mov.u32 %0, %laneid;" : "=r"(lIdx));
+            asm volatile("mov.u32 %0, %%laneid;" : "=r"(lIdx));
 #    else
             asm("mov.u32 %0, %%laneid;" : "=r"(lIdx));
 #    endif


SimeonEhrig (approved):
I found the bug during the work on the job-generator (#641). Unfortunately the generator is not working yet. Therefore a good test coverage via CI is hardly doable. I checked the code with this configuration locally and it works: https://gitlab.com/hzdr/crp/alpaka3/-/jobs/15484843918

I assume the bug is fixed for different CUDA + Clang versions.

psychocoderHPC:
> I found the bug during the work on the job-generator (#641). Unfortunately the generator is not working yet. Therefore a good test coverage via CI is hardly doable. I checked the code with this configuration locally and it works: https://gitlab.com/hzdr/crp/alpaka3/-/jobs/15484843918
> 
> I assume the bug is fixed for different CUDA + Clang versions.

Note: I reproduced the bug locally and fixed it. But I know that for inline assampler `%%` must be used from another project; most likely I forgot about it.

PR #658, opened 2026-07-27
add timed queue and events

TimHanel00:
This PR adds timing events and a timing-capable queue. Both features simply extend the existing objects with a timing policy.

Because CUDA and HIP attach the timing property to events, while SYCL requires profiling to be enabled when the queue is constructed (via `sycl::property::queue::enable_profiling{}`), the policy is retained on both abstractions (queue + events): this preserves alpaka’s existing support for standalone event creation while providing the queue-level configuration required by SYCL.

To enforce consistent behavior across all APIs, timed events can therefore only be enqueued into timing-enabled queues.
A call to:
`queue.makeEvent()`
constructs an appropriate event based on the queue's timing policy. 
This is a shorthand to avoid having to specify the same timing policy redundantly.

Default events and queues remain non-timed to preserve backward compatibility and avoid the additional launch overhead.

This PR adds the policy in the form of two discrete object tags:
`timing::enabled`
`timing::disabled`

## Summary by CodeRabbit

* **New Features**
  * Added opt-in queue/event timing using `timing::enabled`/`timing::disabled`, with timing-aware queue/event creation across backends.
  * `getElapsedTime` now measures queued work for timing-enabled events, including host-side queued tasks between markers.
  * Enforced correctness: timing-enabled events can only be enqueued on timing-enabled queues.
* **Documentation**
  * Updated the “Events and Synchronization” tutorial with a “Measuring Queued Work” section and timing example.
* **Tests**
  * Added unit tests for queued-work timing, host synchronization behavior, enqueue restrictions, and error handling.

### DIFF
--- a/include/alpaka/api/host/Device.hpp
+++ b/include/alpaka/api/host/Device.hpp
@@ -140,7 +140,7 @@ namespace alpaka::onHost
 
             friend struct internal::MakeQueue;
 
-            Handle<cpu::Queue<Device>> makeQueue(alpaka::concepts::QueueKind auto kind)
+            Handle<cpu::Queue<Device>> makeQueue(alpaka::concepts::QueueKind auto kind, alpaka::concepts::Timing auto)
             {
                 ALPAKA_LOG_FUNCTION(onHost::logger::queue);
                 static_assert(
@@ -168,12 +168,12 @@ namespace alpaka::onHost
 
             friend struct internal::MakeEvent;
 
-            Handle<cpu::Event<Device>> makeEvent()
+            Handle<cpu::Event<Device>> makeEvent(alpaka::concepts::Timing auto timingMode)
             {
                 ALPAKA_LOG_FUNCTION(alpaka::onHost::logger::event);
                 auto thisHandle = this->getSharedPtr();
                 std::lock_guard<std::mutex> lk{queuesGuard};
-                auto newEvent = std::make_shared<cpu::Event<Device>>(std::move(thisHandle), events.size());
+                auto newEvent = std::make_shared<cpu::Event<Device>>(std::move(thisHandle), events.size(), timingMode);
 
                 events.emplace_back(newEvent);
                 return newEvent;
--- a/include/alpaka/api/host/Event.hpp
+++ b/include/alpaka/api/host/Event.hpp
@@ -12,6 +12,7 @@
 #include "alpaka/onHost/internal/interface.hpp"
 #include "alpaka/onHost/logger/logger.hpp"
 
+#include <chrono>
 #include <cstdint>
 #include <cstring>
 #include <future>
@@ -25,9 +26,13 @@ namespace alpaka::onHost
         struct Event : std::enable_shared_from_this<Event<T_Device>>
         {
         public:
-            Event(internal::concepts::DeviceHandle auto device, uint32_t const idx)
+            Event(
+                internal::concepts::DeviceHandle auto device,
+                uint32_t const idx,
+                alpaka::concepts::Timing auto timingMode)
                 : m_device(std::move(device))
                 , m_idx(idx)
+                , m_timingEnabled(timingMode == timing::enabled)
             {
                 ALPAKA_LOG_FUNCTION(onHost::logger::event);
             }
@@ -69,6 +74,8 @@ namespace alpaka::onHost
             //!< (not enqueued or already completed). If m_enqueueCount ==
             //!< m_LastReadyEnqueueCount, the event is currently not enqueued
             std::size_t m_LastReadyEnqueueCount = 0u;
+            bool m_timingEnabled = false;
+            std::chrono::steady_clock::time_point m_timestamp;
 
             friend struct alpaka::internal::GetName;
 
@@ -127,6 +134,8 @@ namespace alpaka::onHost
 
             friend struct internal::WaitFor;
             friend struct internal::Wait;
+            template<typename, typename>
+            friend struct internal::GetElapsedTime::Op;
 
             void wait()
             {
@@ -145,7 +154,34 @@ namespace alpaka::onHost
 
             friend struct alpaka::internal::GetApi;
         };
+
     } // namespace cpu
+
+    template<typename T_Device>
+    struct internal::GetElapsedTime::Op<cpu::Event<T_Device>, cpu::Event<T_Device>>
+    {
+        auto operator()(cpu::Event<T_Device>& start, cpu::Event<T_Device>& end) const -> std::chrono::duration<double>
+        {
+            start.wait();
+            end.wait();
+            if(&start == &end)
+            {
+                std::lock_guard<std::mutex> lock{start.m_mutex};
+                if(start.m_enqueueCount == 0u)
+                    throw std::logic_error{"Elapsed time requires recorded events"};
+                if(!start.m_timingEnabled)
+                    throw std::logic_error{"Elapsed time requires timing-enabled events"};
+                return std::chrono::duration<double>::zero();
+            }
+
+            std::scoped_lock lock{start.m_mutex, end.m_mutex};
+            if(start.m_enqueueCount == 0u || end.m_enqueueCount == 0u)
+                throw std::logic_error{"Elapsed time requires recorded events"};
+            if(!start.m_timingEnabled || !end.m_timingEnabled)
+                throw std::logic_error{"Elapsed time requires timing-enabled events"};
+            return end.m_timestamp - start.m_timestamp;
+        }
+    };
 } // namespace alpaka::onHost
 
 namespace alpaka::internal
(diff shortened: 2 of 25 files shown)


psychocoderHPC:
@TimHanel00 could you please rebase against the latest, dev branch there are some small changes.

include/alpaka/api/host/Event.hpp
```diff
@@ -25,9 +26,13 @@ namespace alpaka::onHost
...
         {
         public:
-            Event(internal::concepts::DeviceHandle auto device, uint32_t const idx)
+            Event(
+                internal::concepts::DeviceHandle auto device,
+                uint32_t const idx,
+                alpaka::concepts::Timing auto timingMode)
                 : m_device(std::move(device))
                 , m_idx(idx)
+                , m_timingEnabled(timingMode == timing::enabled)
```
psychocoderHPC, line 35: should be `off`

include/alpaka/api/unifiedCudaHip/Event.hpp
```diff
@@ -154,10 +160,30 @@ namespace alpaka::onHost
...
+            -> std::chrono::duration<double>
+        {
+            start.wait();
+            end.wait();
+            auto milliseconds = 0.0F;
+            using ApiInterface = typename T_Device::ApiInterface;
+            ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
+                ApiInterface,
+                ApiInterface::eventElapsedTime(&milliseconds, start.getNativeHandle(), end.getNativeHandle()));
+            return std::chrono::duration<double>{static_cast<double>(milliseconds) * 1.0e-3};
```
psychocoderHPC: this is maybe wrong. should it be microsections.

psychocoderHPC (changes requested):

include/alpaka/api/unifiedCudaHip/Event.hpp
psychocoderHPC: should be `std::chrono::duration<double, std::milli>{milliseconds}`

include/alpaka/api/host/Event.hpp
TimHanel00: This only assigns a boolean runtime parameter if timingMode == timing::enabled then set true - the condition is correct. Event does not get timingMode as part of its template signature - the dispatch is controlled by a boolean runtime parameter. Are you suggesting making this constexpr controllable aka making it part of the events signature? Sorry for not pointing that out / asking earlier in the offline discussion we had.

psychocoderHPC: thanks for the explanation, I read the code wrong.

include/alpaka/api/host/Device.hpp
```diff
@@ -157,12 +157,12 @@ namespace alpaka::onHost
...
             friend struct internal::MakeEvent;
 
-            Handle<cpu::Event<Device>> makeEvent()
+            Handle<cpu::Event<Device>> makeEvent(alpaka::concepts::Timing auto timingMode)
             {
                 ALPAKA_LOG_FUNCTION(alpaka::onHost::logger::event);
                 auto thisHandle = this->getSharedPtr();
                 std::lock_guard<std::mutex> lk{queuesGuard};
-                auto newEvent = std::make_shared<cpu::Event<Device>>(std::move(thisHandle), queues.size());
+                auto newEvent = std::make_shared<cpu::Event<Device>>(std::move(thisHandle), queues.size(), timingMode);
```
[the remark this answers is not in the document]
psychocoderHPC: @TimHanel00 This is a real bug whcih was already in before it should be what the AI suggested.

```
 auto newEvent = std::make_shared<cpu::Event<Device>>(std::move(thisHandle), events.size(), timingMode);
```

include/alpaka/onHost/Device.hpp
```diff
@@ -98,21 +98,45 @@ namespace alpaka::onHost
...
+                internal::MakeQueue::
+                    Op<ALPAKA_TYPEOF(*m_device.get()), ALPAKA_TYPEOF(kind), ALPAKA_TYPEOF(timingMode)>{}(
+                        *m_device.get(),
+                        kind,
+                        timingMode),
+                kind,
+                timingMode};
+        }
+
+        auto makeQueue(alpaka::concepts::QueueKind auto kind)
```
psychocoderHPC: Please move this function up that it fits the documentation.

include/alpaka/onHost/Device.hpp
```diff
@@ -98,21 +98,45 @@ namespace alpaka::onHost
          *
          * @return A onHost::Queue that tasks and memory operations can be enqueued on.
          */
-        auto makeQueue(alpaka::concepts::QueueKind auto kind)
+        auto makeQueue(alpaka::concepts::QueueKind auto kind, alpaka::concepts::Timing auto timingMode)
```
psychocoderHPC, line 120: Move this function down and add documentation. Could you please try locally if you can pull the documentation and then extend it.

```C++
/** 
  * @copydoc alpaka::onHost::makeQueue
  * @param timingMode foooooooooo
  */
```

include/alpaka/onHost/Device.hpp
```diff
@@ -98,21 +98,45 @@ namespace alpaka::onHost
...
         }
 
         auto makeQueue()
         {
             return makeQueue(queueKind::nonBlocking);
         }
 
+        /** Create an event with an explicit timing capability.
+         *
+         * Timing-enabled events can only be enqueued on timing-enabled queues.
```
psychocoderHPC, line 134: please use doxygen `@param timingMode description`

include/alpaka/onHost/Device.hpp
```diff
@@ -98,21 +98,45 @@ namespace alpaka::onHost
...
+        auto makeEvent(alpaka::concepts::Timing auto timingMode)
+        {
+            return Event{
+                internal::MakeEvent::Op<ALPAKA_TYPEOF(*m_device.get()), ALPAKA_TYPEOF(timingMode)>{}(
+                    *m_device.get(),
+                    timingMode),
+                timingMode};
+        }
+
+        /** Create a synchronization-only event. */
```
psychocoderHPC:
```suggestion
        /** Create a synchronization-only event. 
         *  
         * Timing support is by default disabled.
         */
```

include/alpaka/api/syclGeneric/Event.hpp
```diff
@@ -142,6 +147,27 @@ namespace alpaka::onHost
...
+            start.wait();
+            end.wait();
+            if(&start == &end)
+                return std::chrono::duration<double>::zero();
+
+            auto const startEnd
+                = start.getEvent().template get_profiling_info<sycl::info::event_profiling::command_end>();
+            auto const endStart
+                = end.getEvent().template get_profiling_info<sycl::info::event_profiling::command_start>();
+            if(endStart < startEnd)
```
psychocoderHPC, line 166: For other APIs you do not switch the order. I would remove it because the user should pass it in correct order. But maybe he would like to have a negative time and is intentionally switching the order.

psychocoderHPC (changes requested):

include/alpaka/api/syclGeneric/Event.hpp
TimHanel00: This condition exists in this form only, because sycl hands back the time in an unsigned format, therefore in order to avoid underflow I changed the order, so that a positive value is inserted into the duration, then by negating it I get the original time in a signed format.  Other apis allow negative time sycl does not therefore this seemed to be the most consistent workaround. This means from the users perspective switching the order results in a negative time for all apis - or in cases where the end event simply finishes earlier, which is allowed.

TimHanel00:
I dont think the latest gitlab CI failure are related to this PR. @psychocoderHPC It would be really nice if you could restart those jobs

psychocoderHPC:
@TimHanel00 #566 is mergen, please rebase again (sry) and check the Trait implementation of the collective queue for makeEvent and makeQueue both to get the new timing feature. It is a cpu queue having the normal host queue as member, so you could use the timers stored there. If yo uhave issues to access the timer from the host queue because it is private you can simply add a new method to this specification of a visibility bypass to get access to the member https://github.com/alpaka-group/alpaka3/blob/a07cf995085a3bbba9b4f64ab4fc69f9f2a565f9/include/alpaka/api/host/OmpCollectiveQueue.hpp#L224-L229

TimHanel00:
Timestamps are owned by events, not the host queue, so I only updated the collective MakeQueue trait
specialization. For non-SYCL backends, T_Timing is a type-level queue capability used by queue.makeEvent() and
to constrain event enqueueing, the internal queue of these APIs requires no timing configuration.

psychocoderHPC (approved):

PR #659, opened 2026-07-28
CI: add clang-tidy job

SimeonEhrig:


script/ci/install/clang.sh
```diff
@@ -46,14 +46,22 @@ if [[ "$compiler_name" == "clang" && "$APCI_HIP" == 0 ]]; then
...
+
+        clang_apt_package_list=(
+            "clang-${compiler_version}"
+            "libomp-${compiler_version}-dev"
+            "clang-tools-${compiler_version}"
+            "libclang-rt-${compiler_version}-dev"
+        )
+
+        if [[ "${APCI_CLANG_TIDY}" == "ON" ]]; then
+            clang_apt_package_list+=("clang-tidy-${compiler_version}")
```
SimeonEhrig, line 58: I'm not sure, if this apt package is enough.

It installs the executable `clang-tidy-<version>` (e.g. `clang-tidy-21`). Maybe we need to create the file link `clang-tidy`, which points to `clang-tidy-<version>`, that CMake can find it.

PR #664, opened 2026-07-29
Introduce queue policies

TimHanel00:
This PR introduces onHost::QueuePolicyBundle, which supplies compile-time queue policies in any order:
```
  auto queue = device.makeQueue(
      onHost::QueuePolicyBundle{
          queueKind::nonBlocking,
          timing::enabled});
```
  Using a policy bundle instead of fixed positional arguments makes queue construction extensible. Additional
  policies can be introduced later without adding further `makeQueue` overloads or breaking the new API.

  The policy bundle is forwarded unchanged through the queue interface and unpacked only by the backend that
  requires a specific policy. The queue also retains the bundle so its configuration, including its timing
  capability, remains available.

  Key changes:

  - Add the generic `PolicyBundle` infrastructure and policy registration traits.
  - Add `QueuePolicyBundle`, `QueuePolicy`, and the corresponding concepts.
  - Allow custom queue policies through `trait::IsQueuePolicy`.
  - Preserve the existing defaults:
      - `queueKind::nonBlocking`
      - `timing::disabled`
  - Forward the complete policy bundle to the host, CUDA, HIP, and SYCL backends.
  
  Note: This PR should only be considered ready for merge once #658 is merged.

## Summary by CodeRabbit

* **New Features**
  * Added flexible queue policy configuration, including blocking/non-blocking behavior and optional timing.
  * Added support for timing-enabled events and elapsed-time measurement.
  * Added reusable policy-list support for custom queue policies.
  * Updated queue operations and algorithms to work with configured policy lists.

* **Documentation**
  * Updated queue and event tutorials with the new policy-based configuration and timing instructions.

* **Bug Fixes**
  * Improved event timestamp capture for blocking and non-blocking queue operations.

### DIFF
--- a/docs/source/tutorial/queue.rst
+++ b/docs/source/tutorial/queue.rst
@@ -27,7 +27,8 @@ Here is a small example for a *blocking* queue:
     :dedent:
 
 The use of a *non-blocking* queue requires explicit synchronization before accessing the modified data, otherwise a data race will occur.
-If you do not pass ``queueKind`` as an argument, you will get a *non-blocking* queue.
+Queue configuration is passed as ``onHost::QueuePolicyList``. If the policy list does not contain a queue-kind
+policy, or if you call ``makeQueue()`` without policies, you will get a *non-blocking* queue.
 
   .. literalinclude:: ../../snippets/example/030_queue.cpp
     :language: cpp
--- a/include/alpaka/PolicyList.hpp
+++ b/include/alpaka/PolicyList.hpp
@@ -1,0 +1,95 @@
+/* Copyright 2026 René Widera, Tim Hanel
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include "alpaka/core/common.hpp"
+#include "alpaka/meta/TypeListOps.hpp"
+#include "alpaka/meta/filter.hpp"
+#include "alpaka/unused.hpp"
+
+#include <concepts>
+#include <tuple>
+#include <type_traits>
+
+namespace alpaka
+{
+    namespace trait
+    {
+        /** Identifies compile-time policy tags.
+         *
+         * Specialize this trait with std::true_type for an application-defined policy tag. A policy tag must also be
+         * default-initializable to satisfy alpaka::concepts::Policy.
+         *
+         * @tparam T_Policy Type to inspect.
+         */
+        template<typename T_Policy>
+        struct IsPolicy : std::false_type
+        {
+        };
+    } // namespace trait
+
+    /** Whether a type is registered as a compile-time policy tag. */
+    template<typename T_Policy>
+    constexpr bool isPolicy_v = trait::IsPolicy<std::remove_cvref_t<T_Policy>>::value;
+
+    namespace concepts
+    {
+        /** A registered, default-initializable compile-time policy tag. */
+        template<typename T_Policy>
+        concept Policy = isPolicy_v<T_Policy> && std::default_initializable<std::remove_cvref_t<T_Policy>>;
+    } // namespace concepts
+
+    /** Collection of compile-time policy tags.
+     *
+     * Policy values carry no runtime state. Derived bundles can use search() with a category predicate to select one
+     * policy and provide a category-specific default.
+     *
+     * @tparam T_Policies Registered policy tag types contained in the bundle.
+     */
+    template<concepts::Policy... T_Policies>
+    struct PolicyList
+    {
+    protected:
+        using Args = std::tuple<T_Policies...>;
+
+        /** Select the policy matching a category predicate.
+         *
+         * @param predicate Callable returning true for policies in the requested category.
+         * @param defaultPolicy Policy returned when the bundle contains no match.
+         * @return The matching policy, or defaultPolicy when no policy matches.
+         */
+        template<typename T_Predicate, concepts::Policy T_DefaultPolicy>
+        static constexpr auto search(T_Predicate predicate, T_DefaultPolicy defaultPolicy)
+        {
+            auto const matches = meta::filter(predicate, Args{});
+            static_assert(std::tuple_size_v<ALPAKA_TYPEOF(matches)> <= 1u, "Duplicate policy category.");
+
+            if constexpr(std::tuple_size_v<ALPAKA_TYPEOF(matches)> == 0u)
+                return defaultPolicy;
+            else
+                return meta::Front<ALPAKA_TYPEOF(matches)>{};
+        }
+
+    public:
+        /** Construct a bundle from compile-time policy tags.
+         *
+         * @param policies Policy tags represented by this bundle's type.
+         */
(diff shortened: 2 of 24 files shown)


include/alpaka/api/host/Device.hpp
```diff
@@ -135,34 +135,30 @@ namespace alpaka::onHost
 
             friend struct internal::MakeQueue;
 
-            Handle<cpu::Queue<Device>> makeQueue(alpaka::concepts::QueueKind auto kind)
+            Handle<cpu::Queue<Device>> makeQueue(alpaka::concepts::QueuePolicyBundle auto const& policies)
```
SimeonEhrig: Is there a reason, why this needs to be a template type with concept restriction? I would assume, that we can be more restrictive here and set the argument type to `alpaka::onHost::QueuePolicyBundle`. This would have the benefit, that we can call the `QueuePolicyBundle` constructor implicit which allows to write: `device.makeQueue({queueKind::blocking});`

psychocoderHPC:
I suggest using the name `QueuePolicy` only without bunde, this makes it shorter. IMO the word bundle is not providing any usefull additional information.

include/alpaka/onHost/QueuePolicyList.hpp
```diff
@@ -0,0 +1,104 @@
...
+        /** Return the selected queue kind, or queueKind::nonBlocking if none was supplied. */
+        constexpr alpaka::concepts::QueueKind auto getQueueKind() const
+        {
+            return Base::search(queueKind::detail::QueueKind{}, queueKind::nonBlocking);
+        }
+
+        /** Return the selected timing policy, or timing::disabled if none was supplied. */
+        constexpr alpaka::concepts::Timing auto getTiming() const
+        {
+            return Base::search(timing::detail::Timing{}, timing::disabled);
```
psychocoderHPC: btw. I like this!

TimHanel00:
Bundle because QueuePolicy is already a concept -  it is the property you put into QueuePolicyBundle. Bundle was added to avoid confusion and duplicated namings.

include/alpaka/api/host/Device.hpp
TimHanel00: {A,B} is passed in as a braced initializer list that itself does not have a type. You can not parse an initializer list with arbitrary template types.C++ somehow must first deduce the function-template specialization and its concrete parameter type. There is a related discussion: https://stackoverflow.com/questions/54824308/ctad-initializer-list-non-explicit-constructor-and-call-to-a-function. But because the relevant discussion is really old - I replicated a minimal example with the behavior locally - it shows that inplace construction does not work even with deduction guide, but you can test it yourself. We can however go for a method that allows makeQueue(policyA,policB) but also makeQueue(policyB,policyA,...), which is then basically a factory that internally constructs the QueueProperties object.

psychocoderHPC:
We could provide these onHost interfaces

```C++
template<typename... T_Policies>
Queue makeQueue(T_Policies... policies)
{
    return makeQueue(QueuePolicyBundle{policies...});
}
```

and the current interface with explicit QueuePolicyBundle.
This would allow writing

```C++
auto queue = device.makeQueue(
    queueKind::nonBlocking{},
    timing::enabled{}
);
```

Where the order of arguments is not important.
So the usage would look like it is at the moment but as said without a required order or arguments.

include/alpaka/PolicyBundle.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+    /** Collection of compile-time policy tags.
+     *
+     * Policy values carry no runtime state. Derived bundles can use search() with a category predicate to select one
+     * policy and provide a category-specific default.
+     *
+     * @tparam T_Policies Registered policy tag types contained in the bundle.
+     */
+    template<concepts::Policy... T_Policies>
+    struct PolicyBundle
```
psychocoderHPC: Should be `PolicyList`

psychocoderHPC (changes requested):

docs/source/tutorial/queue.rst
```diff
@@ -27,7 +27,8 @@ Here is a small example for a *blocking* queue:
     :dedent:
 
 The use of a *non-blocking* queue requires explicit synchronization before accessing the modified data, otherwise a data race will occur.
-If you do not pass ``queueKind`` as an argument, you will get a *non-blocking* queue.
+Queue configuration is passed as ``onHost::QueuePolicyList``. If the policy bundle does not contain a queue-kind
```
psychocoderHPC:
```suggestion
Queue configuration is passed as ``onHost::QueuePolicyList``. If the policy list does not contain a queue-kind
```

include/alpaka/tag.hpp
```diff
@@ -100,30 +115,65 @@ namespace alpaka
...
+         *   - isEmpty: all threads wait to see the same state
+         *   - allocDeferred: all threads wait and get the same value
+         *
+         * Operating on some queue handle in parallel from within and outside the OpenMP parallel region results in
+         * undefined behavior.
+         *
+         * The queue can only be used if the dependency OpenMP is available, e.g. by setting the CMake option
+         * ALPAKA_DEP_OMP=ON or using the required compiler flags to activate OpenMP.
+         */
+        struct OmpCollective : category::QueueKind
```
psychocoderHPC: Please move this back to the original file. The collective queue is something user specific and should not be part of alpaka if the file OmpCollectiveQueue.hpp is not included.

psychocoderHPC (changes requested):

include/alpaka/PolicyList.hpp
```diff
@@ -0,0 +1,95 @@
...
+        }
+
+        /** Check whether the bundle contains a policy type.
+         *
+         * @param policy Policy tag whose type is queried.
+         * @return true if the exact policy type is present, otherwise false.
+         */
+        constexpr bool hasPolicy(auto policy) const
+        {
+            return meta::Contains<Args, std::remove_cvref_t<ALPAKA_TYPEOF(policy)>>::value;
```
psychocoderHPC:
```suggestion
            return meta::Contains<Args,ALPAKA_TYPEOF(policy)>::value;
```

The reference and const qualifier are already removed with `ALPAKA_TYPEOF`

include/alpaka/PolicyList.hpp
```diff
@@ -0,0 +1,95 @@
...
+        {
+            alpaka::unused(policies...);
+        }
+
+        /** Check whether the bundle contains a policy type.
+         *
+         * @param policy Policy tag whose type is queried.
+         * @return true if the exact policy type is present, otherwise false.
+         */
+        constexpr bool hasPolicy(auto policy) const
```
psychocoderHPC:
```suggestion
        constexpr bool hasPolicy(concepts::Policy auto policy) const
```

psychocoderHPC (changes requested):
last two mini changes are required

psychocoderHPC (approved):

PR #665, opened 2026-07-30
fix event contract violation

TimHanel00:
This is more a recent issue me and @SimeonEhrig encountered, rather then a ready to merge PR, therefore I suggest keeping it as a draft, while working on a solution. 

I added a small unit test the that showcases the failing behavior.
First of all the assumption is that the queue - blocking queues should support concurrent access by multiple threads. This assumption is justified, because in the documentation under https://alpaka3.readthedocs.io/en/latest/tutorial/queue.html#queue it is stated that: "The use of the same queue by different host threads is supported". This statement does not differentiate the kind of queue. 

Under that premise the test uses two threads T1 and T2. 
T1 enqueues a long running task in `producerQueue`. 
While T2 is running concurrently as the host control thread. 
Therefore if T2 enqueues an event and another queue waits on that event. 

Since `waitFor()` completes before the long-running host task finished - the event synchronization and queue ordering contract is broken.

The reason this currently breaks, is that hostTasks are executed immediately without actually deferring the task execution to a background thread for blocking queues, this paradigm does not establish a fixed syncronization order for a blocking queue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event synchronization and state updates for blocking host queues.
  * Preserved submission order for deferred host tasks.
  * Improved event propagation and ordering across supported queue types.

* **Tests**
  * Added coverage for concurrent event operations and long-running host tasks.
  * Added timing-event tests for queued work, dependencies, host tasks, and queue-created events.
  * Verified appropriate errors for unrecorded events.

### DIFF
--- a/include/alpaka/api/host/Queue.hpp
+++ b/include/alpaka/api/host/Queue.hpp
@@ -218,7 +218,16 @@ namespace alpaka::onHost
             void enqueueHostFnDeferred(auto const& task)
             {
                 ALPAKA_LOG_FUNCTION(onHost::logger::queue);
-                m_sharedCallbackThread->submit(task);
+                if(m_isBlocking)
+                {
+                    // a blocking queue must acquire this lock to ensure that all pending host tasks have finished
+                    std::lock_guard lock(m_mutex);
+                    m_sharedCallbackThread->submit(task);
+                }
+                else
+                {
+                    m_sharedCallbackThread->submit(task);
+                }
             }
 
             void enqueueNativeFn(auto const& fn)
@@ -268,7 +277,7 @@ namespace alpaka::onHost
                 }
             }
 
-            friend struct onHost::internal::GetDevice;
+            friend struct internal::GetDevice;
 
             friend struct internal::Wait;
             friend struct internal::WaitFor;
@@ -308,7 +317,8 @@ namespace alpaka::onHost
                 // open a scope to avoid logging during we hold the lock for this class
                 {
                     // Setting the event state (e.g. the future) and enqueuing it has to be atomic.
-                    std::lock_guard<std::mutex> lk(event.m_mutex);
+                    std::lock_guard<std::mutex> eventLock(event.m_mutex);
+
 
                     ++event.m_enqueueCount;
 
@@ -319,6 +329,10 @@ namespace alpaka::onHost
                      */
                     if(queue.m_isBlocking)
                     {
+                        /* a blocking queue must acquire this lock to ensure that all pending host tasks
+                         * have finished
+                         */
+                        std::lock_guard<std::mutex> queueLock(queue.m_mutex);
                         // Nothing to do if it has been re-enqueued to a later position in the queue.
                         if(enqueueCount == event.m_enqueueCount)
                         {
@@ -364,7 +378,7 @@ namespace alpaka::onHost
                 // open a scope to avoid logging during we hold the lock for this class
                 {
                     // Setting the event state and enqueuing it has to be atomic.
-                    std::unique_lock<std::mutex> lk(event.m_mutex);
+                    std::unique_lock<std::mutex> eventLock(event.m_mutex);
 
                     if(!event.isReady())
                     {
@@ -373,8 +387,12 @@ namespace alpaka::onHost
                          */
                         if(queue.m_isBlocking)
                         {
+                            /* a blocking queue must acquire this lock to ensure that all pending host tasks
+                             * have finished
+                             */
+                            std::lock_guard<std::mutex> queueLock(queue.m_mutex);
                             std::shared_future sFuture = event.m_future;
-                            lk.unlock();
+                            eventLock.unlock();
                             sFuture.get();
                         }
                         else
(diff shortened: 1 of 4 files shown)


SimeonEhrig:
@TimHanel00 Great work! I was thinking in the morning to ask you, if you can do the job but you read my minds and created it already :grin: 

But I'm a little bit confused. A lot of CI jobs already passed, therefor it looks like there is no bug. Does the test failed locally or did you write the test to be sure, that it is working?

TimHanel00:
Yes it definitely failed locally. I will have a look later today. I also played around a bit so mayby I accidently submitted the wrong version.

SimeonEhrig:
> Yes it definitely failed locally. I will have a look later today. I also played around a bit so mayby I accidently submitted the wrong version.

Okay. Maybe you need to increase also sleep timings for the CI. There is a lot of parallel work on the systems. Maybe the CPU scheduler runs the software threads sequential on a CPU thread.

Two comments:
- For testing purpose, you can filter the number of generated CI jobs via commit message. You need only to add a last line starting with `CI_FILTER: ` and regex of allowed job names. Here is an example: https://github.com/alpaka-group/alpaka3/pull/660/changes/4cc5425b18f9d26c548b05ee94df81c1492aca4c   
  - You can also test the regex locally, if you install the job-generator: https://github.com/alpaka-group/alpaka3/blob/dev/script/job_generator/README.md
- You can mark a test to run serial. This increases the chance, that the code in the test really runs in parallel. https://stackoverflow.com/questions/29983138/force-serial-execution-for-specific-targets-in-cmake

TimHanel00:
@SimeonEhrig the problem was for some reason the test I submitted used a non-blocking queue, for which the same scenario is working - since the host task actually gets enqueued and execute by an asynchronous background thread. It is corrected now and the test is failing.

psychocoderHPC:
IMO the test is a good idea, I do not think that we have a bug in the implementation but a test is a good proof.

host: https://github.com/alpaka-group/alpaka3/blob/a07cf995085a3bbba9b4f64ab4fc69f9f2a565f9/include/alpaka/api/host/Queue.hpp#L369

cuda/hip: https://github.com/alpaka-group/alpaka3/blob/a07cf995085a3bbba9b4f64ab4fc69f9f2a565f9/include/alpaka/api/unifiedCudaHip/Queue.hpp#L214

oneapi: https://github.com/alpaka-group/alpaka3/blob/a07cf995085a3bbba9b4f64ab4fc69f9f2a565f9/include/alpaka/api/syclGeneric/Queue.hpp#L418

Only as side note: alpaka mainline was not blocking the caller.

include/alpaka/api/host/Queue.hpp
```diff
@@ -308,7 +308,8 @@ namespace alpaka::onHost
                 // open a scope to avoid logging during we hold the lock for this class
                 {
                     // Setting the event state (e.g. the future) and enqueuing it has to be atomic.
-                    std::lock_guard<std::mutex> lk(event.m_mutex);
+                    std::lock_guard<std::mutex> event_lk(event.m_mutex);
```
psychocoderHPC: should be `eventLock` because we use camelCase.

psychocoderHPC (changes requested):
We need to add the queue lock to few othe rplaces where we missted it.

host/in Queue.hpp: line 218 `void enqueueHostFnDeferred(auto const& task)`
host/in Queue.hpp: line 322: ` struct Enqueue::Event<cpu::Queue<T_Device>, T_Event>`
host/in Queue.hpp: line 376: `struct WaitFor::Op<cpu::Queue<T_Device>, T_Event>`

psychocoderHPC:
> We need to add the queue lock to few othe rplaces where we missted it.
> 
> host/in Queue.hpp: line 218 `void enqueueHostFnDeferred(auto const& task)` host/in Queue.hpp: line 322: ` struct Enqueue::Event<cpu::Queue<T_Device>, T_Event>` host/in Queue.hpp: line 376: `struct WaitFor::Op<cpu::Queue<T_Device>, T_Event>`

`void enqueueHostFnDeferred(auto const& task)`

It should use the lock only if the queue is blocking.

test/unit/event/event.cpp
```diff
@@ -213,6 +216,158 @@ TEMPLATE_LIST_TEST_CASE("basic queue wait for event", "", TestApis)
     CHECK(ev.isComplete() == true);
 }
 
+TEMPLATE_LIST_TEST_CASE("wait for event enqueued concurrently on a queue with a long-running host task", "", TestApis)
+{
+#if ALPAKA_OMP
+    std::tuple queueKindsTuple = std::tuple(queueKind::blocking, queueKind::nonBlocking, queueKind::ompCollective);
+#else
+    std::tuple queueKindsTuple = std::tuple(queueKind::blocking, queueKind::nonBlocking);
+#endif
```
psychocoderHPC:
```suggestion
std::tuple optionalQueueKinds{
#if ALPAKA_OMP
  queueKind::ompCollective
#endif
};

std::tuple queueKindsTuple = std::tuple_cat(std::tuple(queueKind::blocking, queueKind::nonBlocking), optionalQueueKinds);
```

This could reduce code duplication, but I do not see it as mandatory.

psychocoderHPC (approved):
not mendentory suggestion

psychocoderHPC (approved):

PR #663, opened 2026-07-29
[API break] Add clang-tidy CTest target

TimHanel00:
This PR adds a dedicated clangTidyTest target, enabled with the alpaka_CLANG_TIDY CMake option and registered with CTest.

Similar to headerCheck, the test has its own clangTidy directory. It collects all public Alpaka headers, generates a C++ translation unit for each header, and runs clang-tidy through the CMake CXX_CLANG_TIDY integration.

This is intentionally a draft and is currently expected to fail in CI while the naming conventions and required exceptions are being clarified.

I added an initial set of naming rules to .clang-tidy. @psychocoderHPC, please review and adjust this file; some of the current rules may be unnecessary or may not reflect the intended naming conventions yet.

The currently enabled checks cover:

- Clang Static Analyzer checks
- identifier naming
- inconsistent parameter names between declarations and definitions

These checks complement the diagnostics already provided by the normal Clang compilation. Compiler diagnostic checks are not enabled again explicitly.

For testing this target in CI, the PR currently includes the cherry-picked commit 1d762d7 from #659. This commit can be removed after #659 has been merged.

Thanks to @SimeonEhrig for configuring the CI part quickly.

# Breaking changes

## Renaming

```c++
alpaka::detail::Contains::argument_type -> alpaka::detail::Contains::ArgumentType
alpaka::onAcc::order::seq_cst -> alpaka::onAcc::order::seqCst
alpaka::onAcc::order::acq_rel -> alpaka::onAcc::order::acqRel
alpaka::remove_restrict -> alpaka::RemoveRestrict
alpaka::detail::TemplateSignatureStorage_v -> alpaka::detail::templateSignatureStorage_v
alpaka::Vec::T_dim -> alpaka::Vec::dim
alpaka::trait::getVec_t -> alpaka::trait::GetVec_t
alpaka::detail::selectValue -> alpaka::detail::SelectValue
alpaka::onHost::make_sharedSingleton -> alpaka::onHost::makeSharedSingleton
alpaka::onHost::EmbedTypeIntoSignature -> alpaka::onHost::embedTypeIntoSignature
alpaka::onHost::Demangled::DummySignature -> alpaka::onHost::Demangled::dummySignature
alpaka::onHost::Demangled::EmbeddingSignature -> alpaka::onHost::Demangled::embeddingSignature
alpaka::is_decayed_v -> alpaka::isDecayed_v
alpaka::onAcc::detail::atomic_ref -> alpaka::onAcc::detail::AtomicRef
[...]

PR #674, opened 2026-09-07
Memory pining policies

AlexisLaplanche:
This pull request implements memory policies, which can be applied to all memory allocation operations.

The policies included in this pull request are, for the moment, only for host devices. It includes:

- ``alpaka::memoryProperty::bestLatency`` selects memory nodes with the best access latency
- ``alpaka::memoryProperty::bestBandwidth``selects memory nodes with the highest memory bandwidth
- ``alpaka::memoryProperty::locality``  selects the closest memory nodes
- ``alpaka::memoryProperty::defaultProperty`` equivalent to no property; uses the default alpaka3 behavior

These properties bind the memory to the selected nodes, using the MPOL_BIND policy. If the memory nodes are full, the allocation fails.

The code has been tested on Intel Sapphire Rapids with HBM; both the host and NumaCPU devices work as intended.

Other properties specific to one, several, or all devices could easily be added later.

One FIXME remains in the code: should the allocation be made non-strict (using MPOL_PREFERRED_MANY), or should a new policy be added that lets the user select non-strict allocation?

## Summary by CodeRabbit

- **New Features**
  - Added memory allocation properties for selecting latency, bandwidth, locality, or default placement.
  - Allocation APIs now support memory policies across host, mapped, unified, deferred, and “like” allocations.
  - Host allocations can be pinned according to the selected memory preference; unsupported devices safely ignore these preferences.
  - Added support for combining and reusing memory policies through policy lists.

- **Documentation**
  - Added tutorials and advanced reference documentation covering memory pinning and allocation properties.
  - Added examples demonstrating supported allocation methods and policy usage.

### DIFF
--- a/docs/snippets/example/250_memoryProperties.cpp
+++ b/docs/snippets/example/250_memoryProperties.cpp
@@ -1,0 +1,88 @@
+/* Copyright 2026 SiPearl
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#include "docsTest.hpp"
+
+#include <alpaka/alpaka.hpp>
+
+#include <catch2/catch_template_test_macros.hpp>
+#include <catch2/catch_test_macros.hpp>
+
+#include <algorithm>
+#include <iostream>
+#include <type_traits>
+#include <vector>
+
+using namespace alpaka;
+
+TEST_CASE("memory allocations", "[docs]")
+{
+    onHost::concepts::Device auto device = onHost::makeHostDevice();
+    {
+        // BEGIN-TUTORIAL-allocBufferDev
+        concepts::Vector auto extents = Vec{2u, 3u};
+        // the allocation is providing a shared buffer which will be
+        // automatically freed if the last handle runs out of a life-time
+        concepts::IBuffer auto devBuffer = onHost::alloc<int>(device, extents, memoryProperty::bestLatency);
+        // END-TUTORIAL-allocBufferDev
+        unused(devBuffer);
+    }
+    {
+        // BEGIN-TUTORIAL-allocBufferMapped
+        concepts::Vector auto extents = Vec{2u, 3u};
+        // allocate memory which lives on the host but is accessible from the compute device too
+        concepts::IBuffer auto devMappedBuffer
+            = onHost::allocMapped<int>(device, extents, memoryProperty::bestBandwidth);
+        // END-TUTORIAL-allocBufferMapped
+        unused(devMappedBuffer);
+    }
+    {
+        // BEGIN-TUTORIAL-allocBufferUnified
+        concepts::Vector auto extents = Vec{2u, 3u};
+        // allocate memory can be accessed from host and device (unified memory),
+        // the real location depends on the native backend e.g. CUDA, OneApi, ...
+        concepts::IBuffer auto devUnifiedBuffer = onHost::allocUnified<int>(device, extents, memoryProperty::locality);
+        // END-TUTORIAL-allocBufferUnified
+        unused(devUnifiedBuffer);
+    }
+}
+
+void callKernel([[maybe_unused]] auto dummyMemory)
+{
+}
+
+TEST_CASE("memory allocations deferred", "[docs]")
+{
+    onHost::concepts::Device auto device = onHost::makeHostDevice();
+    onHost::Queue queue = device.makeQueue();
+    // BEGIN-TUTORIAL-allocBufferDeferred
+    concepts::Vector auto extents = Vec{2u, 3u};
+    {
+        // The allocation is deferred.
+        // It is only allowed to access the memory after the queue processed the allocation task.
+        concepts::IBuffer auto devDeferredBuffer
+            = onHost::allocDeferred<int>(queue, extents, memoryProperty::bestBandwidth);
+        // Call the kernel with the buffer. This is only a dummy call not a real kernel call.
+        callKernel(devDeferredBuffer);
+        // At the end of the scope the buffer will be destroyed, but it could be that the kernel is not finished yet.
+        // This special allocation method take care that the buffer is waiting for the queue before the memory is
+        // freed.
+    }
+    // END-TUTORIAL-allocBufferDeferred
+}
+
+TEST_CASE("memory allocations like", "[docs]")
+{
+    onHost::concepts::Device auto computeDevice = onHost::makeHostDevice();
+    // BEGIN-TUTORIAL-allocLike
+    concepts::Vector auto extents = Vec{2u, 3u};
+    // short notation to allocate memory on the host without a host device as first argument
+    concepts::IBuffer auto hostBuffer = onHost::allocHost<int>(extents, memoryProperty::bestLatency);
+    // inherits value type and extents but does NOT copy the data
+    concepts::IBuffer auto devDoubleBuffer
+        = onHost::allocLike(computeDevice, hostBuffer, memoryProperty::bestBandwidth);
+    // END-TUTORIAL-allocLike
+
+    unused(devDoubleBuffer);
+}
--- a/docs/source/advanced/properties.rst
+++ b/docs/source/advanced/properties.rst
@@ -1,0 +1,27 @@
+Properties
+==========
+
+Memory Properties
+-----------------
+
+The properties in this subsection are available in the namespace ``alpaka::memoryProperty``. They can be applied to all allocation calls: ``alloc``, ``allocHost``, ``allocMapped``, ``allocUnified``, ``allocLike``, ``allocHostLike``, ``allocDeferred`` and ``allocLikeDeferred``.
+
+- ``memoryProperty::bestLatency`` This property selects memory nodes with the lowest access latency for memory pinning
+- ``memoryProperty::bestBandwidth`` This property selects memory nodes with the highest memory bandwidth for memory pinning
+- ``memoryProperty::locality`` This property selects the closest memory nodes for memory pinning
+- ``memoryProperty::defaultProperty`` Equivalent to no property; uses the default alpaka3 behavior
+
+Only one property of this category may be given per allocation call. If none is given, ``memoryProperty::defaultProperty`` is used.
+
+Memory Policy List
+------------------
+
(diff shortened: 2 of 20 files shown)


psychocoderHPC:
@AlexisLaplanche Thanks for your PR, I marked this PR for release 3.1 because we would like to release at the end of this month. I will have a look into you changes asap

psychocoderHPC:
Maybe some jobs in the gitlab CI will fail with github pull issues, we are working on fixing it. It is triggered by the new limits from Github. Take only real compile issues serios, additionally sometimes HIP runtime tests fail, because of data races in the AMD driver.

SimeonEhrig:
Nice work.
How we want to handle the case, if a device does not support the property. My first idea was a compiler error. At least we should log it. I strongly against to silently ignore it.

AlexisLaplanche:
> Nice work. How we want to handle the case, if a device does not support the property. My first idea was a compiler error. At least we should log it. I strongly against to silently ignore it.

Thanks! Having it raise a compiler error makes having a single code being able to run on multiple devices harder, maybe adding a warning. Or else we could raise a compiler error and create an helper like ``deviceHaspolicy( property )`` which could return "property" or DefaultProperty we then could use: alloc(... , ... , deviceHaspolicy( property )), or something like that.

SimeonEhrig:
> > Nice work. How we want to handle the case, if a device does not support the property. My first idea was a compiler error. At least we should log it. I strongly against to silently ignore it.
> 
> Thanks! Having it raise a compiler error makes having a single code being able to run on multiple devices harder, maybe adding a warning. Or else we could raise a compiler error and create an helper like `deviceHaspolicy( property )` which could return "property" or DefaultProperty we then could use: alloc(... , ... , deviceHaspolicy( property )), or something like that.

Yes, supporting different devices is the major point. I was also thinking about, if it makes unnecessary harder to write portable code. On the other side, if the user already implemented a device selection algorithm, it is not to hard to extent the algorithm to return a tuple of device and property instead only a device. And if the user ignores memory locality, the single device is still fine. We should wait for feedback from @psychocoderHPC. 

But the helper function `deviceHaspolicy( property )` sounds like a good idea.

psychocoderHPC:
I am also not 100% sure how to handle the case if a policy is not available. As @SimeonEhrig said, the goal is to make writing heterogeneous code to be easy.
The question for me is now if this should be a strong memory enforeced policies or memory hints. 
If strong policies should be enforced, we need to thow exceptions and best have our own exceptions based on std::execption in alpaka to make them catchable.
From the point of view of a user, I would like it to see it more as a hint, so it tries to keep the policy in mind but it switches to something working if not available. This would require a method that the user can check before if the policy is supported, as Simeon suggests, e.g. `deviceHaspolicy( property )`, `device.isSupport(property))` or `device.isAllocationSupporting(property)`.
I tend to the second case that is only a hint.

@AlexisLaplanche in OpenMP and AdaptivCpp, where you implemented it to, what is the behaviour there when you use a policy which can not be applied?

AlexisLaplanche:
> I am also not 100% sure how to handle the case if a policy is not available. As @SimeonEhrig said, the goal is to make writing heterogeneous code to be easy. The question for me is now if this should be a strong memory enforeced policies or memory hints. If strong policies should be enforced, we need to thow exceptions and best have our own exceptions based on std::execption in alpaka to make them catchable. From the point of view of a user, I would like it to see it more as a hint, so it tries to keep the policy in mind but it switches to something working if not available. This would require a method that the user can check before if the policy is supported, as Simeon suggests, e.g. `deviceHaspolicy( property )`, `device.isSupport(property))` or `device.isAllocationSupporting(property)`. I tend to the second case that is only a hint.
> 
> @AlexisLaplanche in OpenMP and AdaptivCpp, where you implemented it to, what is the behaviour there when you use a policy which can not be applied?

I also think the hint approach would be easier for the user. I think I also left a FIXME in the code about how we should pin memory: should we use `MPOL_BIND` (strict bind — if there isn't enough room, the allocation fails) or `MPOL_PREFERRED_MANY` (non-strict bind — if it doesn't fit, memory is allocated elsewhere)? `MPOL_PREFERRED_MANY` should be preferred over `MPOL_PREFERRED`, as the fallback behavior differs (see https://docs.kernel.org/admin-guide/mm/numa_memory_policy.html#components-of-memory-policies).

In AdaptiveCpp, since SYCL is very low level, the implementation choice was to target a NUMA node by its ID directly — the user has to manage which node they want to pin memory to:

- When the device doesn't support memory binding, the policy is ignored (see https://github.com/AdaptiveCpp/AdaptiveCpp/blob/develop/doc/extensions.md#acpp_ext_target_numa_node_property)
- If the NUMA node doesn't exist, the allocation fails and returns NULL
- If the NUMA node is full, it falls back to the standard Linux behavior, meaning the memory is placed elsewhere
- To stay compatible with other SYCL implementations (like DPC++), allocations have to be guarded with `#ifdef ACPP_EXT_TARGET_NUMA_NODE_PROPERTY`
- If libnuma isn't found at build time, a warning is printed the first time the property is encountered, to tell the user it will be ignored

Kokkos (internal, not yet upstream, subject to change) does something similar, but using hwloc and only targeting the best node according to the property rather than selecting a NUMA node directly:
- The Kokkos is tightly coupled to hwloc: there is no wrapper around the property, it is forwarded directly to the allocation
- All allocations need to be guarded with `#ifdef KOKKOS_ENABLE_HWLOC`

Making it a hint would avoid the widespread `#ifdef` usage and the multiple code paths depending on the device. A warning would still be possible.

To address both the `MPOL_BIND` vs `MPOL_PREFERRED_MANY` question and whether the allocation should be strict, we could add a `strictAllocation` property that fails if the property isn't applicable on the device, or if the target memory node is full. That case could raise an exception.

SimeonEhrig:
> I am also not 100% sure how to handle the case if a policy is not available. As @SimeonEhrig said, the goal is to make writing heterogeneous code to be easy. The question for me is now if this should be a strong memory enforeced policies or memory hints. If strong policies should be enforced, we need to thow exceptions and best have our own exceptions based on std::execption in alpaka to make them catchable. From the point of view of a user, I would like it to see it more as a hint, so it tries to keep the policy in mind but it switches to something working if not available. This would require a method that the user can check before if the policy is supported, as Simeon suggests, e.g. `deviceHaspolicy( property )`, `device.isSupport(property))` or `device.isAllocationSupporting(property)`. I tend to the second case that is only a hint.

I think we should try a hard compile time error, because most of the users will not set the property and the default is used. If the user is setting the policy, I can image two cases:

1. The user writes a dedicated specialization for CPUs with memory location awareness. For example:

```c++
template<typename MemoryProperty = alpaka::memoryProperty::defaultProperty>
void simulation(auto deviceSelector);

simulation<alpaka::memoryProperty::bestLatency>(alpaka::onHost::makeDeviceSelector(api::host, deviceKind::cpu));
simulation(alpaka::onHost::makeDeviceSelector(api::cuda, deviceKind::nvidiaGPU));
```

2. The user sets a fallback like @AlexisLaplanche suggested:

```c++
alpaka::alloc(dev , ... , alpaka::has_policy(dev, alpaka::memoryProperty::bestLatency, alpaka::memoryProperty::defaultProperty))
```

I think option 2. is the most useful solution because if we different allocations with different properties in our code, option 1 would require N template parameter for each allocation. Also this solution is the explicit version of silent ignore unsupported properties. Because it is a optimization, I think we can be explicit.

Also an additional idea. `alpaka::has_policy` could be variadic, means we could implement something like:

```c++
// try to use the memory with best bandwidth
// if the property is not supported, try to use the closets memory -> there is a high chance for the best memory
// if nothing is supported, use the fallback
alpaka::has_property(dev, alpaka::memoryProperty::bestBandwidth, 
  alpaka::memoryProperty::locality, 
  alpaka::memoryProperty::defaultProperty)
```

AlexisLaplanche:
> I think we should try a hard compile time error, because most of the users will not set the property and the default is used. If the user is setting the policy, I can image two cases:
> 
> 1. The user writes a dedicated specialization for CPUs with memory location awareness. For example:
> 
> ```c++
> template<typename MemoryProperty = alpaka::memoryProperty::defaultProperty>
> void simulation(auto deviceSelector);
> 
> simulation<alpaka::memoryProperty::bestLatency>(alpaka::onHost::makeDeviceSelector(api::host, deviceKind::cpu));
> simulation(alpaka::onHost::makeDeviceSelector(api::cuda, deviceKind::nvidiaGPU));
> ```
> 
> 2. The user sets a fallback like @AlexisLaplanche suggested:
> 
> ```c++
> alpaka::alloc(dev , ... , alpaka::has_policy(dev, alpaka::memoryProperty::bestLatency, alpaka::memoryProperty::defaultProperty))
> ```
> 
> I think option 2. is the most useful solution because if we different allocations with different properties in our code, option 1 would require N template parameter for each allocation. Also this solution is the explicit version of silent ignore unsupported properties. Because it is a optimization, I think we can be explicit.
> 
> Also an additional idea. `alpaka::has_policy` could be variadic, means we could implement something like:
> 
> ```c++
> // try to use the memory with best bandwidth
> // if the property is not supported, try to use the closets memory -> there is a high chance for the best memory
> // if nothing is supported, use the fallback
> alpaka::has_property(dev, alpaka::memoryProperty::bestBandwidth, 
>   alpaka::memoryProperty::locality, 
>   alpaka::memoryProperty::defaultProperty)
> ```

I agree with solution 2, it make the implicit fallback more explicit. It is true that in a same part of a code we could have array with different property (I mostly imaging having bestLatency and bestBandwitdh on different array for a same part of code). 

The variadic version is a great idea, has we could also imagine in the futur having memory property for GPUs backends, with this we could have try GPU property, else try CPU property, else fallback to default.

I will implement option 2 and test various possibilities to see if there is other things to improve in the logic

psychocoderHPC:
I agree option 2 looks nice. Take only care we use camelCase for functions so it would be `hasProperty()`.

It should also be in the namespace onHost.

SimeonEhrig:
> I agree option 2 looks nice. Take only care we use camelCase for functions so it would be `hasProperty()`.
> 
> It should also be in the namespace onHost.

We should finish #663 that the computer can solve my naming problems ;-)

PR #675, opened 2026-09-07
Implementation of parallel memcpy

AlexisLaplanche:
The goal of this PR is to implement a parallel memcpy for host devices (CPUs: OMP, TBB, CPUNuma). A single CPU core cannot saturate the full memory bandwidth of a CPU, so a memcpy operation using multiple cores can be sped up.

This has not been done for non-host backends, as on GPUs a single large memcpy generally performs better than multiple small ones.

Parallel memcpy has been implemented by enqueuing a parallel copy into the queue, instead of using the "single task" submit used for a sequential copy.

Several ways of enabling and tuning this behavior have been implemented:
- Set the default mode using the CMake variable ``alpaka_PARALLEL_MEMCPY_HOST_AS_DEFAULT``. It can be set to true (parallel) or false (sequential).
- Use the ``ALPAKA_MEMCPY_MODE``, ``ALPAKA_MEMCPY_NUM_CORES`` and ``ALPAKA_MEMCPY_MIN_SIZE`` environment variables at runtime.
- Or use the per-queue API to set these values for a single queue.

## Summary by CodeRabbit

- **New Features**
  - Added optional parallel memory copying for CPU queues, configurable per queue or through environment variables.
  - Added controls for copy mode, core usage, and minimum transfer size.
  - Added support for parallel copying of multidimensional and pitched data.

- **Documentation**
  - Added configuration, environment-variable, and tutorial documentation for parallel copying.

- **Examples**
  - Added a memory-copy example demonstrating transfers, verification, and throughput reporting.

- **Tests**
  - Added coverage for configuration, transfer directions, multidimensional buffers, and subviews.


psychocoderHPC:
I have not looked deeply into this PR but should a memcopy not look like memset but with a source IView instead of a byte constant? 

https://github.com/alpaka-group/alpaka3/blob/fd65c5a31d62ce6b71b29aa6fd4462e2e0a72b54/include/alpaka/api/generic.hpp#L45-L88

In principle, this should be implementable with `onHost::concurent()`, something like

```
struct Memcpy
{
    constexpr void operator()(
        alpaka::concepts::SimdPtr auto const& dst,
        alpaka::concepts::SimdPtr auto const& src) const
    {
        dst = src.load();
    }
};

alpaka::onHost::concurrent<float>(
    queue,
    alpaka::onHost::getExtents(src),
    Memcpy{},
    dst,
    src);
```

This would use SIMD and will be out of the box support muti dimensional views. but will not use byte copy if not explicitly cast, it is an element-wise copy. Not fully sure if this is faster than `std::memcpy()` but it would also work for GPU and take into acount that the memory access pattern on GPU differs compared to CPU.

Is it possible to compare `onHost::concurrent ' with your implementation?

AlexisLaplanche:
> I have not looked deeply into this PR but should a memcopy not look like memset but with a source IView instead of a byte constant?

It should work, I will make some tests

> This would use SIMD and will be out of the box support muti dimensional views. but will not use byte copy if not explicitly cast, it is an element-wise copy. Not fully sure if this is faster than `std::memcpy()` but it would also work for GPU and take into acount that the memory access pattern on GPU differs compared to CPU.

Not sure if it is really wanted. I don't have a high-performance GPU on hand but it is sometimes preferred to have a single large copy for GPUs instead of multiple small ones, but the configuration for GPU queue could be set to use one core by default and user could choose to use more.

> Is it possible to compare `onHost::concurrent ' with your implementation?

Sure! I will have a look into it.

psychocoderHPC:
> Not sure if it is really wanted. I don't have a high-performance GPU on hand but it is sometimes preferred to have a single large copy for GPUs instead of multiple small ones, but the configuration for GPU queue could be set to use one core by default and user could choose to use more.

If you write the test simply as alpaka benchmark we can also execute it on H100 or similar GPUs.

PR #679, opened 2026-09-13
refactor README

psychocoderHPC:
- Add an eye-catcher to the README file.
- Add a showcase example to the README file.

## Summary by CodeRabbit

* **Documentation**
  * Added a concise C++ example demonstrating device selection, memory allocation, and a dot product computation.
  * Added performance portability documentation featuring BabelStream benchmark context.
  * Included bandwidth comparison charts for NVIDIA GH200 GPU (CUDA) and Grace CPU (OpenMP).


docs/snippets/example/020_device.cpp
```diff
@@ -17,7 +17,7 @@ TEST_CASE("show host devices", "[docs]")
      * host+cpu, cuda+nvidiaGpu, hip+amdGpu, oneApi+intelGpu, oneApi+cpu,
      * oneApi+amdGpu, oneApi+nvidiaGpu
      */
-    auto computeDevSelector = alpaka::onHost::makeDeviceSelector(api::host, deviceKind::cpu);
```
psychocoderHPC, line 20: This should not be part of the commit, I will remove it tomorrow.

README.md
```diff
@@ -32,7 +32,40 @@ Explicit caching of data within a frame via shared memory allows developers to f
...
+auto queue = device.makeQueue();
+
+uint32_t const numElements = 107343742u;
+
+// Allocate the input buffers and a single-element result buffer on the selected device.
+concepts::IBuffer auto resultBuffer = onHost::alloc<double>(device, 1u);
+concepts::IBuffer auto bufferLhs = onHost::alloc<double>(device, numElements);
+concepts::IBuffer auto bufferRhs = onHost::allocLike(device, bufferLhs);
+
+// ... Initialize bufferLhs and bufferRhs ...
```
SimeonEhrig: I think we can use `alpaka::onHost::fill` or `alpaka::onHost::iota`. It is not the benchmark code, therefore I would prefer a code, which you can simply copy, compile and execute.

SimeonEhrig:
`107% of NVIDIA reference throughput on an average` instantly triggers me because, it sounds like this horrible, meaningless (,sometimes wrong) vendor benchmarks for the marketing. Maybe I'm unique. I'm not against it in general but maybe other people has the same feeling/opinion.

@fwyzard What are you thinking about? It's about the text on the benchmark images.

fwyzard:
> `107% of NVIDIA reference throughput on an average` instantly triggers me because, it sounds like this horrible, meaningless (,sometimes wrong) vendor benchmarks for the marketing.

Well, it kind of is ;-)

I would probably write "7% faster" than "107%", and qualify it as "on the BabelStream benchmark" to make it clear it's not an overall claim.

But I don't mind the marketing approach, especially because it's clear it looks like a sales pitch.

psychocoderHPC:
@SimeonEhrig @fwyzard It is a marketing plot, it should catch the reader. The comparison is also fair, because the vendor is running an auto-tune before each benchmark size to find the best setting, alpaka is not doing it, but even then we can compare with the BabelStream results.
On GPU we are slightly slower; therefore, I kept 107% and 99%, it would looks strange to say we are 1% slower -> that's not how marketing works^^

README.md
psychocoderHPC: I updated the example

SimeonEhrig:
Okay. Than I'm fine with it.
I know, it should catch but in my case it did the opposite. At least, I know I'm unique ;-)

SimeonEhrig (approved):

PR #685, opened 2026-09-15
CI: update the patch level of the CMake versions to match pre-installed CMake version

SimeonEhrig:
The alpaka-group-container provides pre-installed CMake versions. If we match the patch level, we can use the preinstalled versions instead pull a CMake version each time.

## Summary by CodeRabbit

* **Chores**
  * Updated CMake versions used by emulated SIMD and sanitizer CI jobs.
  * Refreshed several tested CMake patch versions to newer releases.


psychocoderHPC (approved):

psychocoderHPC:
I will merge it with my admin power because cmake.org is still down and blocks our development.

PR #690, opened 2026-09-16
[API break] make member variables m_kernelFn and m_args of KernelBundle private

SimeonEhrig:
- add getters to access member variables

## Summary by CodeRabbit

* **New Features**
  * Added read-only accessors for retrieving a kernel bundle’s kernel function and arguments.

* **Bug Fixes**
  * Resolved a spurious compiler warning affecting dynamic shared-memory handling with specific HIP and amd-clang versions.

* **Refactor**
  * Updated dynamic shared-memory handling to use the new accessors without changing runtime behavior.


psychocoderHPC (approved):

include/alpaka/onHost/trait.hpp
```diff
@@ -93,18 +93,39 @@ namespace alpaka::onHost
...
+                     * Using [[maybe_unused]] does not solve the problem. The warning only appears if the user
+                     * configures the shared memory size with the variable dynSharedMemBytes.
+                     *
+                     * @todo: remove me, if HIP 7.0 and 7.1 is not supported anymore
+                     */
+#if ALPAKA_LANG_HIP >= ALPAKA_VERSION_NUMBER(7, 0, 0) && ALPAKA_LANG_HIP <= ALPAKA_VERSION_NUMBER(7, 1, 0)
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wunused-result"
+#endif
+                    return kernelBundle.getKernelFn().dynSharedMemBytes;
```
psychocoderHPC, line 125: ONly a note I would say

```
auto const& foo = kernelBundle.getKernelFn();
return foo.dynSharedMemBytes;
```

should work as a workaround too. No strong need to try if it works but I would like to mention it.

psychocoderHPC (approved):

include/alpaka/onHost/trait.hpp
SimeonEhrig: I test it locally. Let's see.

SimeonEhrig: No, does not solve the problem. The compiler simply follows the chain and detects that the intermediate variable does not change the usage. Also the return value of the function is stored in the variable. In the end, the problem is, that the compiler does not count the usage of the variable as cuda launch parameter as usage.

https://github.com/alpaka-group/alpaka3/blob/35ba304fd2215117cd1179830d2cbe468ced3285/include/alpaka/api/unifiedCudaHip/Queue.hpp#L369-L375

PR #693, opened 2026-09-21
[API break] make IdxRange::{m_begin, m_end, m_stride} private

SimeonEhrig:
- It looks like the members should be private because of the naming and the undefined behavior of changing the members after creating the object.
- Didn’t named the getters `begin()`, `end()` and `stride()` because they are already used the the stl requires iterateable objects.

## Summary by CodeRabbit

- **Refactor**
  - Standardized access to index-range information across supported container and range operations.
  - Exposed read-only access to range boundaries and strides for improved interface consistency.
  - Existing iteration, striding, and sub-range behavior remains unchanged.


include/alpaka/mem/FlatIdxContainer.hpp
```diff
@@ -201,26 +201,26 @@ namespace alpaka::onAcc
 
             if constexpr(std::is_same_v<T_IdxMapperFn, layout::Strided>)
             {
-                auto groupOffset = threadIdx * m_idxRange.m_stride;
+                auto groupOffset = threadIdx * m_idxRange.getStrideVec();
```
psychocoderHPC: In alpaka we agreed that we use the plural to mark that it will return a vector, so the name should be `getStrides()`

We discussed it in few dev meetings where I proposed using `Md` as a postfix.

include/alpaka/mem/IdxRange.hpp
```diff
@@ -125,11 +125,32 @@ namespace alpaka
             return stream.str();
         }
 
+        /** Return the begin position of the range as vector coordinate.
+         * Allows not to iterate of the IdxRange. Use begin() instead of. **/
+        [[nodiscard]] constexpr auto getBeginVec() const -> T_Begin
```
psychocoderHPC:
```suggestion
        [[nodiscard]] constexpr auto getBegins() const -> T_Begin
```

The name sounds strange; maybe `getBeginCursor()`? Cursor is a common name for a multidimensional index defining a position in an array. We used it in the past in PIConGPU and [flux](https://github.com/tcbrindle/flux#the-flux-difference) is using it too.
A cursor is not necessary; only an index, it could be any information that is valid together with an object to mark a point in space/range.

psychocoderHPC (changes requested):
We need to decide for naming, the current one is possible where I tent more to `cursor` becaus ethe alpaka way, adding an `s` sound/looked bad.

@SimeonEhrig What is your favorite?

include/alpaka/mem/FlatIdxContainer.hpp
SimeonEhrig: I'm fine with the Md postfix because:
- `getBegins()` sounds to familiar to `begin()`. I would expect, that I get an iterator. Thats the reason, why I used the `Vec` postifx.
- We use already `Md`, therefore I would not introduce the new term `Cursor`. To honest, for me `Cursor` sounds also like more a mighty object, like an iterator not a "stupid" vector.

@psychocoderHPC Are you fine with `getBeginMd()`, `getEndMd()` and `getStrideMd()`?

psychocoderHPC: I am for using `getBeginMd()`, `getEndMd()` and `getStrideMd()` to keep it similar for all getters.

SimeonEhrig: done

psychocoderHPC (approved):

PR #694, opened 2026-09-21
[API break] make ThreadSpace::{m_threadIdx, m_threadCount} private

SimeonEhrig:
# Breaking changes

- make `ThreadSpace::m_threadIdx` and `ThreadSpace::m_threadCount` private
- renamed `ThreadSpace::idx()` to `ThreadSpace::getThreadIdx()`
- renamed `ThreadSpace::size()` to `ThreadSpace::getThreadCount()`
- renamed `WorkerGroup::idx()` to `WorkerGroup::getThreadIdx()`
- renamed `WorkerGroup::size()` to `WorkerGroup::getThreadCount()`

## Summary by CodeRabbit

- **Refactor**
  - Standardized contiguous mapping to use supported thread-space accessors for thread counts.
  - Clarified the return types of thread-space index and count accessors, improving consistency for integrations.
  - Existing thread indexing and contiguous mapping behavior remains unchanged.
  - No user-facing features or capabilities were added or removed.


include/alpaka/mem/ThreadSpace.hpp
```diff
@@ -121,27 +132,27 @@ namespace alpaka
     } // namespace internal
 
     template<std::size_t I, typename T_ThreadIdx, typename T_ThreadCount>
-    constexpr auto get(alpaka::ThreadSpace<T_ThreadIdx, T_ThreadCount> const& v) requires(I == 0u)
+    constexpr auto get(alpaka::ThreadSpace<T_ThreadIdx, T_ThreadCount> const& v) -> T_ThreadIdx requires(I == 0u)
```
SimeonEhrig, line 125: I added the return type because it make the code better readable, there is a small chance that I did a mistake and does an accidental cast and it speeds up the compile time.

include/alpaka/mem/ThreadSpace.hpp
```diff
@@ -94,10 +94,21 @@ namespace alpaka
             return ThreadSpace<ALPAKA_TYPEOF(threadIndex), ALPAKA_TYPEOF(numThreads)>{threadIndex, numThreads};
         }
 
-        T_ThreadIdx m_threadIdx;
-        T_ThreadCount m_threadCount;
+        [[nodiscard]] constexpr auto getThreadIdx() const -> T_ThreadIdx
```
psychocoderHPC, line 87: In line 44 and following, we already have getter methods `size()` and `idx()`; there is no need to introduce new.
I was also not aware of it, but I saw it during the review.

The reason why the corresponding method for `getThreadCount()` is named `size()` is that this would increase compatibility with the standard library call `std::size(foo)`

psychocoderHPC (changes requested):
IMO we should reuse the methods `size()` and `idx()`

include/alpaka/mem/ThreadSpace.hpp
psychocoderHPC: I thought more about the naming, since `size()` is returning a `alpaka::Vec` object, it is not compatible with `std::size()` so we should keep your naming `getThreadCount()` and `getThreadIdx()` and remove the other two methods and use the new interface functions.

SimeonEhrig: I removed the functions and renamed some functions of WorkerGroup, see PR description.

psychocoderHPC (approved):

